PR #17, opened 2023-08-15
Fix catalog-to-spark data type mapping

seankao-az:
### Description
Fix data type mapping from catalog to spark. This is the first of a series of PRs to support more data types, such as varchar and decimal.

Notable differences in catalog types and Spark types:
|CatalogDataType|SparkDataType|
|-|-|
|bigint|long|
|int|integer|
|smallint|short|
|tinyint|byte|

Now these and struct type can be mapped correctly (as well as all other types)

### Issues Resolved

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingIndex.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingIndex.scala
@@ -75,19 +75,15 @@ class FlintSparkSkippingIndex(
   }
 
   private def getSchema: String = {
-    val indexFieldTypes =
-      indexedColumns.flatMap(_.outputSchema()).map { case (colName, colType) =>
-        // Data type INT from catalog is not recognized by Spark DataType.fromJson()
-        val columnType = if (colType == "int") "integer" else colType
-        val sparkType = DataType.fromJson("\"" + columnType + "\"")
-        StructField(colName, sparkType, nullable = false)
-      }
-
     val allFieldTypes =
-      indexFieldTypes :+ StructField(FILE_PATH_COLUMN, StringType, nullable = false)
-
+      indexedColumns.flatMap(_.outputSchema()).toMap + (FILE_PATH_COLUMN -> "string")
+    val catalogDDL =
+      allFieldTypes
+        .map { case (colName, colType) => s"$colName $colType not null" }
+        .mkString(",")
+    val allFieldSparkTypes = StructType.fromDDL(catalogDDL)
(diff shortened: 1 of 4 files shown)


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/FlintSparkSkippingIndex.scala
```diff
@@ -75,19 +75,15 @@ class FlintSparkSkippingIndex(
   }
 
   private def getSchema: String = {
-    val indexFieldTypes =
-      indexedColumns.flatMap(_.outputSchema()).map { case (colName, colType) =>
-        // Data type INT from catalog is not recognized by Spark DataType.fromJson()
```
dai-chen, line 80: Do you know if the issue in this comment gone or now? Because I don't see we change `Data.fromJson()` in this PR?

seankao-az: the catalog types can be parsed with fromDDL, without us making an exception for INT, TINYINT, SMALLINT, BIGINT, and so on. Using fromDDL also helps parse struct correctly

dai-chen: Got it. Do you know what's the difference between Spark SQL type (StructType) and catalog type? So we're actually dealing with 4 data types: source (Parquet type), catalog, Spark SQL and sink (OpenSearch type) ?

seankao-az: There's a 1:1 mapping between catalog types and Spark data types. For example
* tinyint <=> ByteType
* double <=> DoubleType
* varchar(20) <=> VarcharType (with property length: 20).

Parquet only supports primitive types, namely, BOOLEAN, INT32, INT64, INT96, FLOAT, DOUBLE, and BYTE_ARRAY. Logical types are built upon these primitive types, defined in parquet metadata.

In fact, we don't have to deal with parquet types at all, as that's already handled by Spark when creating the table (CREATE TABLE ... USING PARQUET). Parquet has the ability to contain all data types supported by Spark, and thus, in order for us to support parquet file, we need to support all Spark data types.

Most of the Spark types can be mapped directly to OpenSearch types, and are already supported by Flint.
The missing ones are:
* decimal(precision, scale)
* varchar(length)
* char(length)
* binary
* interval year to month
* interval day to second
* array<type>
* map<type,type>

dai-chen (approved):

PR #21, opened 2023-08-23
Support Spark data types VarcharType and CharType

seankao-az:
### Description
Support VarcharType and CharType. Update documentation for Spark data type support.

Length property for both types is omitted when mapping to Flint data type *keyword*. They're both treated as Spark StringType from Flint's perspective.

### DIFF
--- a/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
+++ b/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
@@ -129,6 +129,7 @@ object FlintDataType {
         } else {
           JObject("type" -> JString("keyword"))
         }
+      case _: VarcharType | _: CharType => JObject("type" -> JString("keyword"))
 
       // Numbers
       case LongType => JObject("type" -> JString("long"))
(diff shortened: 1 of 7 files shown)


dai-chen:
Any IT we need to add in low-level `FlintDataSourceV2ITSuite`?

seankao-az:
I think not. From the perspective of FlintTable (OpenSearch index) datasource, varchar and char types are transparent. Because Spark convert them into StringType before writing them to OpenSearch index, OpenSearch index doesn't know varchar and char exist. To OpenSearch, they're treated as regular strings.

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
```diff
@@ -129,6 +129,7 @@ object FlintDataType {
         } else {
           JObject("type" -> JString("keyword"))
         }
+      case _: VarcharType | _: CharType => JObject("type" -> JString("keyword"))
```
penghuo, line 132: it is possible that mapping varchar | char to text field if osType =="text", right

seankao-az: done

dai-chen (approved):
Thanks for the changes!

Issue #23, opened 2023-08-29
[FEATURE] Add covering index support


dai-chen:
**Is your feature request related to a problem?**

Support covering index in Flint Spark integration:

1. Flint API support
2. SQL statement support

Idea: https://github.com/opensearch-project/sql/discussions/1465

PR #36, opened 2023-09-18
Add describe and show covering index SQL support

dai-chen:
### Description

1. Add `DESC INDEX` statement support
2. Add `SHOW INDEX` index statement support based on a new `getAllIndexMetadata()` API in Flint client 

#### Example

```
spark-sql> SHOW INDEX ON stream.lineitem_tiny;

spark-sql> CREATE INDEX orderkey_and_quantity
         > ON stream.lineitem_tiny (l_orderkey, l_quantity)
         > WITH (auto_refresh = true);

spark-sql> SHOW INDEX ON stream.lineitem_tiny;
index_name
orderkey_and_quantity

spark-sql> CREATE INDEX discount_and_quantity
         > ON stream.lineitem_tiny (l_discount, l_quantity)
         > WITH (auto_refresh = true);

spark-sql> SHOW INDEX ON stream.lineitem_tiny;
index_name
discount_and_quantity
orderkey_and_quantity

spark-sql> DESC INDEX discount_and_quantity ON stream.lineitem_tiny;
indexed_col_name	data_type	index_type
l_discount	         float	            indexed
l_quantity	         float	            indexed
```

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/23


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/covering/FlintSparkCoveringIndexAstBuilder.scala
```diff
@@ -55,6 +58,41 @@ trait FlintSparkCoveringIndexAstBuilder extends FlintSparkSqlExtensionsVisitor[C
...
+      AttributeReference("data_type", StringType, nullable = false)(),
+      AttributeReference("index_type", StringType, nullable = false)())
+
+    FlintSparkSqlCommand(outputSchema) { flint =>
+      val indexName = getFlintIndexName(flint, ctx.indexName, ctx.tableName)
+      flint
+        .describeIndex(indexName)
+        .map { case index: FlintSparkCoveringIndex =>
+          index.indexedColumns.map { case (colName, colType) =>
+            Row(colName, colType, "indexed")
```
seankao-az, line 89: What does index_type do?

flint-core/src/main/scala/org/opensearch/flint/core/FlintClient.java
```diff
@@ -33,6 +32,14 @@ public interface FlintClient {
    */
   boolean exists(String indexName);
 
+  /**
+   * Retrieve all metadata for Flint index whose name matches the given pattern.
+   *
+   * @param indexNamePattern index name pattern
+   * @return all matched index metadata
+   */
+  List<FlintMetadata> getAllIndexMetadata(String indexNamePattern);
```
YANG-DB, line 41: why is the 'All' needed - the parameter and return value already indicate this is a pattern and may return a list of indices

YANG-DB (approved):

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/covering/FlintSparkCoveringIndexAstBuilder.scala
dai-chen: For now we only support INDEXED column. Actually there is another type INCLUDED columns which we may support later. Ref: https://www.postgresql.org/docs/current/indexes-index-only-scans.html

seankao-az (approved):

flint-core/src/main/scala/org/opensearch/flint/core/FlintClient.java
dai-chen: Because there is another API `FlintMetadata getIndexMetadata(String indexName)` below with same name and argument list. I'm thinking `get[AllI]ndexMetadata` maybe better than `getIndexMetadata[s]`?

I was also thinking of providing single getIndexMetadata API for both purpose. But it seems confusing when we pass index name and expect single metadata returned. Thanks

YANG-DB: maybe `getIndicesMetadata` - anyhow the return signature also helps distinguish between them IMO ...

dai-chen: I can rename it to `getIndexMetadata[s]` if you have strong opinion? IMO, it's intuitive to have different method name and also required by compiler...

dai-chen: Will merge this for now and rename if we come up with better name. Thanks!

Issue #26, opened 2023-08-30
[FEATURE] Support more options in create index statement


dai-chen:
**Is your feature request related to a problem?**

Currently Flint streaming job uses Foreach Batch sink which creates temporary checkpoint folder by default. The checkpoint will be lost after restart.

**What solution would you like?**

Add another option in create statement to allow user configure where they want to store the checkpoint data. Without the location provided, we can assume user doesn't concern about this and stick with temporary checkpoint as before.

For example:

```
CREATE SKIPPING INDEX ON alb_logs
(
  client_ip VALUE_SET
)
WITH (
  auto_refresh = true,
  checkpoint_location = 's3://...'
)
```

**What alternatives have you considered?**

Use some default location of current database. For example, because we map data source concept to SparkSQL catalog, we may use the root folder of a S3 data source. But this needs user awareness and write permission.

**Do you have any additional context?**

N/A

dai-chen:
Similarly, we want to support other options such as refresh interval, index setting etc.

```
CREATE SKIPPING/INDEX/MV ...
WITH (
  auto_refresh = [true | false],
  auto_start = [true | false],
  refresh_interval = '1s',
  checkpoint_location = 's3://...',
  index_settings = '{ "key": value ... }'
);
```

dai-chen:
`auto_start` is low priority and will be tracked separately.

Issue #24, opened 2023-08-30
Refactoring Flint metadata and build index API


dai-chen:
Code needs refactoring after materialized view added finally:

- [x] Finalize `FlintMetadata` and refactor serde code:
    - [x] common code moved into `flint-core` and other left in `flint-spark-integration`
    - [x] update Flint user reference doc accordingly
- [x] Refactor `FlintSpark.createIndex` which always uses foreach sink
    - [x] MV build DataFrame doesn't use foreach sink
    - [ ] check if covering index can still foreach sink or not

PR #39, opened 2023-09-20
Support refresh interval and checkpoint location option

dai-chen:
### Description

1. Added `refresh_interval` and `checkpoint_location` option and store it in Flint metadata
2. Fixed `STRING` token and `deleteIndex()` stop job first bug
3. Extract Spark parsing to `SparkSqlAstBuilder` and mix-in to Flint parser

#### TODO

1. Didn't find easy way to verify index option in streaming job. Will add UT in next PR for index setting option.
2. https://github.com/opensearch-project/opensearch-spark/issues/24: build API refactor will come first in next PR for MV.
3. ~Will update user manual with new index options in next PR.~

#### Example

```
spark-sql> CREATE INDEX orderkey_and_quantity
         > ON stream.lineitem_tiny (l_orderkey, l_quantity)
         > WITH (
         >   auto_refresh = true,
         >   refresh_interval = '10 seconds',
         >   checkpoint_location = 's3a://test/'
         > );

GET flint_stream_lineitem_tiny_orderkey_and_quantity_index/_mapping
{
  "flint_stream_lineitem_tiny_orderkey_and_quantity_index": {
    "mappings": {
      "_meta": {
        "name": "orderkey_and_quantity",
        "options": {
          "auto_refresh": "true",
          "refresh_interval": "10 seconds",
          "checkpoint_location": "s3a://test/"
        },
        "source": "stream.lineitem_tiny",
        "kind": "covering",
        ...
    }
  }
}

spark-sql> 23/09/21 16:06:30 WARN ProcessingTimeExecutor: Current batch is falling behind. 
The trigger interval is 10000 milliseconds, but spent 72423 milliseconds

$ aws s3 ls test
     commits/
     offsets/
     sources/
```

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/26


integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
```diff
@@ -102,6 +103,27 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
...
+  }
+
+  test("create skipping index with index options successfully") {
+    flint
+      .skippingIndex()
+      .onTable(testTable)
+      .addValueSet("address")
+      .options(FlintSparkIndexOptions(Map(
+        "auto_refresh" -> "true",
+        "refresh_interval" -> "1 Minute",
```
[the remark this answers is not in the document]
dai-chen: Addressed in https://github.com/opensearch-project/opensearch-spark/pull/39/commits/0212b93de2c7f23b9d22767d644c03d7a63846af. Thanks!

docs/index.md
```diff
@@ -190,6 +190,34 @@ DESCRIBE INDEX elb_and_requestUri ON alb_logs
 DROP INDEX elb_and_requestUri ON alb_logs
 ```
 
+#### Create Index Options
+
+User can provide the following options in `WITH` clause of create statement:
+
++ `auto_refresh`: triggers Incremental Refresh immediately after index create complete if true. Otherwise, user has to trigger Full Refresh by `REFRESH` statement manually.
++ `refresh_interval`: a string as the time interval for incremental refresh, e.g. 1 minute, 10 seconds. This is only applicable when auto refresh enabled. Please check `org.apache.spark.unsafe.types.CalendarInterval` for valid duration identifiers.
++ `checkpoint_location`: a string as the location path for incremental refresh job checkpoint. The location has to be a path in an HDFS compatible file system and only applicable when auto refresh enabled.
```
vamsi-amazon, line 199: What is the default value if nothing is provided.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -128,11 +130,22 @@ class FlintSpark(val spark: SparkSession) {
...
-          }
-          .start()
-        Some(job.id.toString)
+
+        index.options
+          .checkpointLocation()
+          .foreach(location => job.option("checkpointLocation", location))
+        index.options
+          .refreshInterval()
+          .foreach(interval => job.trigger(Trigger.ProcessingTime(interval)))
```
vamsi-amazon, line 139: Why are we looping here? does refreshInterval give single interval?

dai-chen: It is actually equivalent to `if (option.isDefined) then do with option.get`. I wrote it in this way earlier but my IDE recommended me to use foreach instead.

docs/index.md
dai-chen: Update doc with default behavior: https://github.com/opensearch-project/opensearch-spark/pull/39/commits/87417fc10bc3e0fa58f9a3af30f36e435d64c664. Thanks!

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexSqlITSuite.scala
```diff
@@ -59,6 +55,26 @@ class FlintSparkSkippingIndexSqlITSuite extends FlintSparkSuite {
     indexData.count() shouldBe 2
   }
 
+  test("create skipping index with streaming job options") {
+    withTempDir { checkpointDir =>
```
vamsi-amazon, line 59: QQ for my learning: withTempDir is it some function provided default or is it defined any where.

dai-chen: yes, it's provide by Spark base class. I found it in Spark test code. :)

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/SparkSqlAstBuilder.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  override def visitPropertyKey(key: PropertyKeyContext): String = {
+    if (key.STRING() != null) {
+      string(key.STRING())
+    } else {
+      key.getText
+    }
+  }
+
+  override def visitPropertyValue(value: PropertyValueContext): String = {
```
vamsi-amazon, line 44: Is every option treated as string.
Few QQs:
* If user passes some random strings in auto refresh. When will the failure occur?
* [Not related to this PR] Any thoughts on communicating syntax and other errors to user

docs/index.md
vamsi-amazon: So, if we change application and nothing is provided here, we might recalculate everything?

vamsi-amazon:
LGTM

docs/index.md
dai-chen: You mean EMR-S application right? As long as the code we submit along with checkpoint location are the same, the job will restart from last checkpoint. I have a task to double check this once this PR and `IF NOT EXISTS` support in https://github.com/opensearch-project/opensearch-spark/pull/42 merged.

vamsi-amazon (approved):

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/SparkSqlAstBuilder.scala
dai-chen: 1. Yes, it's a map of string passed to `FlintSparkIndexOptions` which will interpret it.
2. For the error case, currently exception will be thrown when we build DataFrame for streaming job in `FlintSpark`. Let me think about if Spark error is clear enough or can we validate beforehand. Thanks for the good point!

Issue #27, opened 2023-08-30
[FEATURE] Support IF NOT EXISTS in create index statement


dai-chen:
**Is your feature request related to a problem?**

Currently Flint index already exist error happens if Spark application job with CREATE statement restarts. This is inconvenient for user to restart streaming job.

**What solution would you like?**

For example, support `CREATE SKIPPING INDEX IF NOT EXISTS ...` so user can simply restart the application.

**What alternatives have you considered?**

Require user to submit REFRESH statement to restart streaming job

**Do you have any additional context?**

N/A

PR #42, opened 2023-09-25
Support IF NOT EXISTS in create statement

dai-chen:
### Description

Add `IF NOT EXISTS` in create index statement. This is useful along with `checkpoint_location` option introduced in https://github.com/opensearch-project/opensearch-spark/pull/39. We just submit the exact same job (create with if not exists statement) to Spark to restart index refreshing, instead of submitting new job with `REFRESH` statement.

Doc: https://github.com/dai-chen/opensearch-spark/blob/support-if-not-exists/docs/index.md

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/27


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -74,14 +74,18 @@ class FlintSpark(val spark: SparkSession) {
    *
    * @param index
    *   Flint index to create
+   * @param ignoreIfExists
+   *   Ignore existing index
    */
-  def createIndex(index: FlintSparkIndex): Unit = {
+  def createIndex(index: FlintSparkIndex, ignoreIfExists: Boolean = false): Unit = {
     val indexName = index.name()
     if (flintClient.exists(indexName)) {
```
vamsi-amazon, line 82: As per PR description, if we pass `IF NOT EXISTS` we perform refresh...where is that part? I could only see throw Exception.

vamsi-amazon:
In general, what is preferred REFRESH INDEX statement or CREATE INDEX with IF NOT EXISTS.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
dai-chen: It's here in SQL command layer: https://github.com/opensearch-project/opensearch-spark/blob/main/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/skipping/FlintSparkSkippingIndexAstBuilder.scala#L53

By design, `FlintSpark` is on lower level which provides fine grained API (`createIndex` and `refreshIndex` separately) for higher level to combine them as wish.

dai-chen:
> In general, what is preferred REFRESH INDEX statement or CREATE INDEX with IF NOT EXISTS.

I think run `REFRESH` statement directly is more clear. `CREATE INDEX with IF NOT EXISTS` is for use case like EMR job which we want to just restart with same job info without looking into it. Anyway, these options are only in SQL. In either case, only `FlintSpark.refreshIndex` API will be triggered underlying. Thanks!

vamsi-amazon (approved):

PR #56, opened 2023-10-03
Add  ppl-spark-extension optimzer support & mvn publish

### LINKED ISSUE #55: [FEATURE]Add flint's optimization rules to PPL FlintPPLSparkExtensions
Adding Flint's Query Optimization rules for the FlintPPLSparkExtensions so that queries may benefit the acceleration capabilities  

Adding the optimizer to the PPL spark extension class:

```
    extensions.injectOptimizerRule { spark =>
      new FlintSparkOptimizer(spark)
    }
```
Added FlintPPLSparkExtensions jar to the maven publish job 

---

YANG-DB:
### Description

add optimizer support for flint based acceleration queries - also using the PPL
add ppl-spark-extension jar mvn publish

### Issues Resolved
#55 

### DIFF
--- a/integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintPPLSuite.scala
+++ b/integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintPPLSuite.scala
@@ -3,7 +3,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package org.opensearch.flint.spark
+package org.opensearch.flint.spark.ppl
+
+import org.opensearch.flint.spark.{FlintPPLSparkExtensions, FlintSparkExtensions}
 
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.catalyst.expressions.CodegenObjectFactoryMode
@@ -13,7 +15,7 @@ import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.test.SharedSparkSession
 
 trait FlintPPLSuite extends SharedSparkSession {
-  override protected def sparkConf = {
+  override protected def sparkConf: SparkConf = {
     val conf = new SparkConf()
       .set("spark.ui.enabled", "false")
       .set(SQLConf.CODEGEN_FALLBACK.key, "false")
@@ -22,17 +24,12 @@ trait FlintPPLSuite extends SharedSparkSession {
       // LocalRelation will exercise the optimization rules better by disabling it as
       // this rule may potentially block testing of other optimization rules such as
       // ConstantPropagation etc.
-      .set("spark.sql.extensions", classOf[FlintPPLSparkExtensions].getName)
+      .set(SQLConf.OPTIMIZER_EXCLUDED_RULES.key, ConvertToLocalRelation.ruleName)
+      .set(
+        "spark.sql.extensions",
+        List(classOf[FlintPPLSparkExtensions].getName, classOf[FlintSparkExtensions].getName)
+          .mkString(", "))
+      .set(OPTIMIZER_RULE_ENABLED.key, "false")
     conf
   }
-
-  private def withFlintOptimizerDisabled(block: => Unit): Unit = {
-    spark.conf.set(OPTIMIZER_RULE_ENABLED.key, "false")
-    try {
-      block
-    } finally {
-      spark.conf.set(OPTIMIZER_RULE_ENABLED.key, "true")
-    }
-  }
-
 }
--- a/integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTimeWindowITSuite.scala
+++ b/integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTimeWindowITSuite.scala
@@ -7,13 +7,10 @@ package org.opensearch.flint.spark.ppl
 
 import java.sql.Timestamp
 
-import org.opensearch.flint.spark.FlintPPLSuite
-
 import org.apache.spark.sql.{QueryTest, Row}
 import org.apache.spark.sql.catalyst.analysis.{UnresolvedAttribute, UnresolvedFunction, UnresolvedRelation, UnresolvedStar}
 import org.apache.spark.sql.catalyst.expressions.{Alias, Ascending, Divide, Floor, GenericRowWithSchema, Literal, Multiply, SortOrder, TimeWindow}
 import org.apache.spark.sql.catalyst.plans.logical._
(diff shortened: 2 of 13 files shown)


ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/FlintPPLSuite.scala
```diff
@@ -24,4 +25,14 @@ trait FlintPPLSuite extends SharedSparkSession {
       .set("spark.sql.extensions", classOf[FlintPPLSparkExtensions].getName)
     conf
   }
+
+  private def withFlintOptimizerDisabled(block: => Unit): Unit = {
```
vamsi-amazon, line 29: Is this for running any code without optimizer rule enabled? Where is this used?

integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTimeWindowITSuite.scala
```diff
@@ -394,7 +397,7 @@ class FlintSparkPPLTimeWindowITSuite
     // Define the expected logical plan
     val star = Seq(UnresolvedStar(None))
     val ageField = UnresolvedAttribute("age")
-    val table = UnresolvedRelation(Seq("default", "flint_ppl_test"))
+    val table = UnresolvedRelation(Seq("spark_catalog", "default", "flint_ppl_sales_test"))
```
vamsi-amazon, line 400: datasource in PPL is catalog in spark. There is a PR in this repo to handle that.
 QQ: Does PPL work with other catalog names if passed as a parameter?
 https://github.com/opensearch-project/opensearch-spark/pull/45
 Can you verify that once?

ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/FlintPPLSuite.scala
YANG-DB: @dai-chen my understanding is that anyhow we need to call the SQL for the create tables part - this would only require the flint extension without any need for the ppl 
The PPL query use case does need to utilize the acceleration capabilities of the Flint and this is why it depends on Flint...
The tests are disabled from the optimizer since this is not a testing purpose for PPL, the acceleration is tested in the flint-integration

YANG-DB: since the UX is explicitly requesting the specific query language in the API (PPL/SQL) this allows the API to use different spark extensions depending on the query language

integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTimeWindowITSuite.scala
YANG-DB: @vamsi-amazon the unresolved relation is agnostic to the FQN of the table including the datasource & schema. This is being resolved in the analysis step and totally depended by the user given url

ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/FlintPPLSuite.scala
dai-chen: @YANG-DB I meant whether PPL query can be accelerated or not depends on if Flint extension available in Spark at runtime, right? What I don't understand is why Flint extension is required by PPL. I thought PPL and Flint should be 2 individual extension and can work together if you start Spark with both. Please correct if I'm wrong.

YANG-DB: IMO there are 3 use cases:
1. only using Flint SQL 
2. only using Flint PPL (without accelerations)
3. using both ppl / sql with accelerations

Option 1 only needs the flint-spark jar 
Option 2 only needs the flint-ppl-spark jar 
Option 3 needs both

In order to reduce complexity with the jars list to the spark job - I think that PPL can be depended on flint so that it would bundle both option 2 & 3. Another option is to add another jar that combines both ...

YANG-DB: In order to explicitly add this explicitly in the spark extension :
````
class FlintPPLSparkExtensions extends (SparkSessionExtensions => Unit) {

  override def apply(extensions: SparkSessionExtensions): Unit = {
    extensions.injectParser { (spark, parser) =>
      new FlintSparkPPLParser(parser)
    }

    extensions.injectOptimizerRule { spark =>
      new FlintSparkOptimizer(spark)
    }
  }
}
````
This requires PPL to be dependent on flintSpark ...

Comment (dai-chen) [reply ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/FlintPPLSuite.scala:29]: As discussed, please double check whether:

1. Spark can support multiple extensions
2. Extension can work together, specifically every extension has the chance to parser and optimize a query

We need to think about the impact if this was different from what we thought. Otherwise, we're good to avoid this dependency. Thanks!

Comment (YANG-DB) [reply ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/FlintPPLSuite.scala]: @dai-chen  agree - thanks for u'r comments 👍

**Review by vamsi-amazon — APPROVED** [review]

File: integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintPPLSuite.scala [hunk 3]
```diff
@@ -21,7 +24,12 @@ trait FlintPPLSuite extends SharedSparkSession {
       // LocalRelation will exercise the optimization rules better by disabling it as
       // this rule may potentially block testing of other optimization rules such as
       // ConstantPropagation etc.
-      .set("spark.sql.extensions", classOf[FlintPPLSparkExtensions].getName)
+      .set(SQLConf.OPTIMIZER_EXCLUDED_RULES.key, ConvertToLocalRelation.ruleName)
+      .set(
+        "spark.sql.extensions",
+        List(classOf[FlintPPLSparkExtensions].getName, classOf[FlintSparkExtensions].getName)
```
dai-chen, line 30: `FlintSparkExtensions` is accessible to ppl module?

dai-chen (approved):

integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintPPLSuite.scala
YANG-DB: This is used only in the `integ-test` project which is depended on both flint & ppl plugins

PR #61, opened 2023-10-04
Replace static extensions with params

YANG-DB:
### Description
Currently both [FlintJob](https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala) and [SQLJob](https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/opensearch/sql/SQLJob.scala) are explicitly hard coded using the `FlintSparkExtension` without allowing the job submitter to configure any alternative

### Issues Resolved
Completely remove `extension-list` parameter which will be given to the spark-config as part of the job submission and not be present in the Job input params


spark-sql-application/README.md
```diff
@@ -52,7 +54,7 @@ aws emr-serverless start-job-run \
     --application-id <application-id> \
     --execution-role-arn <execution-role>  \
     --job-driver '{"sparkSubmit": {"entryPoint": "<flint-job-s3-path>", \
-      "entryPointArguments":["'<sql-query>'", "<result-index>", "<data-source-name>"], \
+      "entryPointArguments":["'<spark extensions>'","'<sql-query>'", "<result-index>", "<data-source-name>"], \
```
dai-chen, line 57: `<data-source-name>` seems moved to `spark.flint.datasource.name`. Can we pass this by Spark conf as well?

vamsi-amazon: How do we decide if it should be entry point arguments or SparkSubmitParameters. There is one more extension config in SparkSubmitParameters. Which one should we use?

YANG-DB: IMO the config based features such as:
 - spark extensions
 - data-source-name
 should be part of the in the entry-point.
Other parameters which are more data-driven:
 - query 
 - index names
 are more often changing and should arrive via the submit args - WDYT ?

penghuo: spark.sql.extensions is a Spark conf. we could remove set in FlintJob and SQLJob and depend on plugin conf it.

YANG-DB: I've removed all the `spark extensions` arguments both in FlintJob and SQLJob

vamsi-amazon (approved):

penghuo (approved):

PR #77, opened 2023-10-18
add USING statement for covering / MV indices for using existing inde…

### LINKED ISSUE #72: [FEATURE]Support existing Index usage in Flint
As a flint user, I'd like to use existing indices / index-templates for using as the index targets of the flint accelerated tables

Use existing index name to create the acceleration process - this will actually not create an index but use the given name as the target of the acceleration ETL store.

The next SQL syntax suggested:
```
CREATE (SKIPPING/COVERING/MV) INDEX
ON alb_logs USING ss4o_logs-elb-prod
WITH (
  auto_refresh = true,
  refresh_interval = '1 minute',
  checkpoint_location = 's3://test/'
)
```

Would initiate the acceleration ETL sync process without actually creating a new index in OpenSearch .

It may validate the following:
 - match source table (catalog) columns exist in the given index_mapping
 - match the index has an alias with the correct flint naming convention 
 - match the index_template (mapping) has the correct covering - matadata fields exist (covering / MV columns)

[Using existing SS4O schema definition](https://github.com/opensearch-project/opensearch-catalog/tree/main/docs/schema/observability)

YANG-DB:
As a flint user, I'd like to use existing indices / index-templates for using as the index targets of the flint accelerated tables

Use existing index name to create the acceleration process - this will actually not create an index but use the given name as the target of the acceleration ETL store.

The next SQL syntax suggested:
```
CREATE (SKIPPING/COVERING/MV) INDEX
ON alb_logs USING ss4o_logs-elb-prod
WITH (
  auto_refresh = true,
  refresh_interval = '1 minute',
  checkpoint_location = 's3://test/'
)
```

Would initiate the acceleration ETL sync process without actually creating a new index in OpenSearch, it will use the `ss4o_logs-elb-prod` index (index template) as the data store for the acceleration content.

It may validate the following:
 - match source table (catalog) columns exist in the given index_mapping
 - match the index has an alias with the correct flint naming convention 
 - match the index_template (mapping) has the correct covering - matadata fields exist (covering / MV columns)

### Issues Resolved
#72

### DIFF
--- a/flint-core/src/main/scala/org/opensearch/flint/core/FlintClient.java
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/FlintClient.java
@@ -27,6 +27,13 @@ public interface FlintClient {
   void createIndex(String indexName, FlintMetadata metadata);
 
   /**
+   * Create an alias name for the given index
+   * @param indexName
+   * @param aliasName
+   */
+  void alias(String indexName, String aliasName, FlintMetadata metadata);
+
+  /**
    * Does Flint index with the given name exist
    *
    * @param indexName index name
--- a/flint-core/src/main/scala/org/opensearch/flint/core/metadata/FlintMetadata.scala
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/metadata/FlintMetadata.scala
@@ -18,6 +18,8 @@ import org.opensearch.flint.core.metadata.FlintJsonHelper._
 case class FlintMetadata(
     /** Flint spec version */
     version: FlintVersion,
+    /** Flint index target name */
+    targetName: Option[String],
     /** Flint index name */
     name: String,
     /** Flint index kind */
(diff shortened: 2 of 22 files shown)


dai-chen:
Could you explain the semantics of this new syntax? So user already has index with Flint metadata, this actually not CREATE but just alias their index name with expected Flint name. Why this has to be done in Flint extension?

YANG-DB:
this is to be done in flint since we do need to add the metadata section to an existing opensearch index
it will take an existing index and add the correct flint name according to the convention and the _meta columns that are accelerated

dai-chen:
I see. If so, could you do it as how we support `index_settings` option? I think you just need to add one more `alias` option in `FlintSparkIndexOption` and use it in `FlintClient` and `FlintMetadata`. No need to change so many places.

YANG-DB:
remove this PR since the functionality will be moved to materialized view

PR #158, opened 2023-11-15
Improve Flint index monitor error handling

dai-chen:
### Description

1. Avoid throwing exception which will cause all subsequent task executions cancelled
2. Follow REPL heartbeat updater to use Spark daemon thread pool

### Issues Resolved

N/A

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
@@ -5,7 +5,7 @@
 
 package org.opensearch.flint.spark
 
-import java.util.concurrent.{Executors, ScheduledExecutorService, ScheduledFuture, TimeUnit}
+import java.util.concurrent.{ScheduledExecutorService, ScheduledFuture, TimeUnit}
 
 import scala.collection.concurrent.{Map, TrieMap}
 import scala.sys.addShutdownHook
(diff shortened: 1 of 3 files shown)


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
```diff
@@ -100,7 +100,8 @@ object FlintSparkIndexMonitor extends Logging {
    * Thread-safe ExecutorService globally shared by all FlintSpark instance and will be shutdown
    * in Spark application upon exit. Non-final variable for test convenience.
    */
-  var executor: ScheduledExecutorService = Executors.newScheduledThreadPool(1)
+  var executor: ScheduledExecutorService =
+    newDaemonThreadPoolScheduledExecutor("flint-index-heartbeat", 1)
```
penghuo, line 104: what is issue with scheduled thread pool? do we have test cover assert daemon thread?

dai-chen: The only difference is the thread created by executor will be daemon which won't block JVM exit (in case the thread/executor doesn't shutdown normally). Made this change because I saw it's used in FlintREPL heartbeat updater and thought it maybe safer? @kaituo 

Checked Executor interface but didn't find API expose the internal thread in Executor. Here is the Spark util method we call:

```
  def newDaemonThreadPoolScheduledExecutor(threadNamePrefix: String, numThreads: Int)
      : ScheduledExecutorService = {
    val threadFactory = new ThreadFactoryBuilder()
      .setDaemon(true)
      .setNameFormat(s"$threadNamePrefix-%d")
      .build()
    val executor = new ScheduledThreadPoolExecutor(numThreads, threadFactory)
    // By default, a cancelled task is not automatically removed from the work queue until its delay
    // elapses. We have to enable it manually.
    executor.setRemoveOnCancelPolicy(true)
    executor
  }
```

penghuo (approved):

Issue #157, opened 2023-11-15
[FEATURE] Change default idle timeout to 10mins


penghuo:

kaituo:
fixed in the above PR

PR #160, opened 2023-11-16
Enhance Logging for Job Cleanup and Reduce REPL Inactivity Timeout

kaituo:
### Description

- Added detailed logging to improve visibility during streaming job cleanup.
- Decreased REPL job inactivity timeout from 30 to 10 minutes..

Tested manually to ensure new logs are correctly displayed during streaming job cleanup.

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/157

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
@@ -7,10 +7,6 @@
 package org.apache.spark.sql
 
 import java.util.Locale
-import java.util.concurrent.ThreadPoolExecutor
-
-import scala.concurrent.{ExecutionContext, Future, TimeoutException}
-import scala.concurrent.duration.{Duration, MINUTES}
 
 import org.opensearch.client.{RequestOptions, RestHighLevelClient}
 import org.opensearch.cluster.metadata.MappingMetadata
@@ -23,9 +19,7 @@ import play.api.libs.json._
 import org.apache.spark.SparkConf
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.parser.ParseException
-import org.apache.spark.sql.flint.config.FlintSparkConf
 import org.apache.spark.sql.types.{StructField, _}
-import org.apache.spark.util.ThreadUtils
 
 /**
  * Spark SQL Application entrypoint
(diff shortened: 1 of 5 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -83,15 +84,41 @@ object FlintJob extends Logging with FlintJobExecutor {
...
       // Stop SparkSession if streaming job succeeds
       if (!exceptionThrown && wait.equalsIgnoreCase("streaming")) {
         // wait if any child thread to finish before the main thread terminates
         spark.streams.awaitAnyTermination()
       } else {
         spark.stop()
       }
-
+    } catch {
+      case e: Exception => logError("fail to close spark session", e)
```
dai-chen, line 120: Thanks for the fix! 2 quick questions:

1. For streaming job, we don't need to call `spark.stop()`?
2. Just wondering do we need to write the error message to result index?

kaituo: 1. I think we need to call spark.stop for streaming job as well.
2. yes, we can write the error message when failing to close spark session.

dai-chen: As discussed, the error message belongs to streaming job instead of DDL statement. Will cover this in Flint index monitor. Please ignore the 2nd item.

kaituo: thanks

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -0,0 +1,111 @@
...
+      cleanUpResources(exceptionThrown, threadPool, dataToWrite, resultIndex, osClient)
+    }
+  }
+
+  def cleanUpResources(
+      exceptionThrown: Boolean,
+      threadPool: ThreadPoolExecutor,
+      dataToWrite: Option[DataFrame],
+      resultIndex: String,
+      osClient: OSClient): Unit = {
```
dai-chen, line 79: this `osClient` seems not closed explicitly?

dai-chen (approved):
Thanks for the changes!

Just minor comment that could you add some info logging when we shutting down thread pool and stopping SparkContext? thinking maybe helpful if we know the shutdown code actually runs. Thanks!

kaituo:
> Thanks for the changes!
> 
> Just minor comment that could you add some info logging when we shutting down thread pool and stopping SparkContext? thinking maybe helpful if we know the shutdown code actually runs. Thanks!

added more logging. Also reused JobOperator in FlintREPL as well.

PR #175, opened 2023-11-22
Fix shutdown bug due to non-daemon thread in driver

kaituo:
### Description
Resolve an issue where a non-daemon thread, potentially created by a bug in dependencies, prevents the driver from shutting down properly. The fix ensures the JVM exits gracefully, avoiding resource leaks and preventing hanging EMR-s jobs.

Tests:
- Reproduced the bug to confirm the issue.
- Applied the fix and verified that the driver now shuts down as expected.

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/util/ThreadPoolFactory.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/util/ThreadPoolFactory.scala
@@ -5,10 +5,66 @@
 
 package org.apache.spark.sql.util
 
-import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.{ScheduledExecutorService, TimeUnit}
 
-trait ThreadPoolFactory {
+import scala.collection.JavaConverters._
+
+import org.apache.spark.internal.Logging
+
+trait ThreadPoolFactory extends Logging {
   def newDaemonThreadPoolScheduledExecutor(
       threadNamePrefix: String,
       numThreads: Int): ScheduledExecutorService
+
+  def shutdownThreadPool(executor: ScheduledExecutorService): Unit = {
+    logInfo(s"terminate executor ${executor}")
+    executor.shutdown() // Disable new tasks from being submitted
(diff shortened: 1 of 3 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/util/ThreadPoolFactory.scala
```diff
@@ -5,10 +5,66 @@
...
       threadNamePrefix: String,
       numThreads: Int): ScheduledExecutorService
+
+  def shutdownThreadPool(executor: ScheduledExecutorService): Unit = {
+    logInfo(s"terminate executor ${executor}")
+    executor.shutdown() // Disable new tasks from being submitted
+
+    try {
+      // Wait a while for existing tasks to terminate
+      if (!executor.awaitTermination(60, TimeUnit.SECONDS)) {
```
penghuo, line 25: wait for 60s heartbeat update?

spark-sql-application/src/main/scala/org/apache/spark/sql/util/ThreadPoolFactory.scala
```diff
@@ -5,10 +5,66 @@
...
+        logWarning("Executor did not terminate in the specified time.")
+        val tasksNotExecuted = executor.shutdownNow() // Cancel currently executing tasks
```
penghuo, line 27: comparing to threadpool.shutdown what is solved by executor.shutdownNow?

kaituo: My code has more logs and retry and handling interruption.

spark-sql-application/src/main/scala/org/apache/spark/sql/util/ThreadPoolFactory.scala
kaituo: it can be any running or dangling thread tasks.

penghuo (approved):

PR #179, opened 2023-11-30
Reduce Session Document Retrieval in FintREPL to Enhance Latency Metrics

kaituo:
### Description
This PR reduces the frequency of 'getSessionDoc' calls in two places of FintREPL, addressing the correlation between request count and query latency metrics.

1. **Heartbeat Update Optimization**:
   - Prior to updating the heartbeat, the sequence number and primary term are now obtained for effective concurrency control.
   - This PR removes the get session doc call and directly updates the last update time and state.

2. **Session Document Retrieval before Statement Processing**:
   - Previously, in scenarios where a query takes 10 minutes, the 'getSessionDoc' call is limited to once per 10 minutes. However, in idle states with no running queries, the call frequency is run every 100 milliseconds. 
   -  This PR reduced the frequency of 'getSessionDoc' calls by ensuring we make the call at least 1 minute after the previous call.

**Testing**:
- Verified consistent 1-minute intervals for heartbeat updates.
- Confirmed the 'getSessionDoc' call executes every 1 minute prior to picking up the next statement.

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
@@ -6,8 +6,6 @@
 package org.apache.spark.sql
 
 import java.net.ConnectException
-import java.time.Instant
-import java.util.Map
 import java.util.concurrent.{ScheduledExecutorService, ScheduledFuture}
 
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future, TimeoutException}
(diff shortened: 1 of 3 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
```diff
@@ -481,18 +486,27 @@ object FlintREPL extends Logging with FlintJobExecutor {
...
+      val currentTime = currentTimeProvider.currentEpochMillis()
+
+      // Only call canPickNextStatement if EARLY_TERMIANTION_CHECK_FREQUENCY milliseconds have passed
+      if (currentTime - lastCanPickCheckTime > EARLY_TERMIANTION_CHECK_FREQUENCY) {
+        canPickNextStatementResult =
+          canPickNextStatement(sessionId, jobId, osClient, sessionIndex)
+        lastCanPickCheckTime = currentTime
+      }
+
+      if (!canPickNextStatementResult) {
```
penghuo, line 509: Check flintReader.hasNext achieve same result?
```
      if (!flintReader.hasNext) {
        canProceed = false
      } else if (!canPickNextStatement(sessionId, jobId, osClient, sessionIndex)) {
        canPickNextStatementResult = false
        canProceed = false 
     }
```

penghuo (approved):

PR #215, opened 2024-01-09
Percent-encode invalid flint index characters

### LINKED ISSUE #192: [FEATURE] Improve special character handling for generating Flint index name

Spark table restricts only alphabet, digit and underscore in table name. However, table doesn't from Spark/Hive catalog can have special characters. In this case, the Flint index name generated will be rejected by OpenSearch restricts as below:

> Index names can’t contain spaces, commas, or the following characters: :, ", *, +, /, \, |, ?, #, >, or <

Ref: https://opensearch.org/docs/latest/api-reference/index-apis/create-index/#index-naming-restrictions

Some extra encoding is required here to solve the problem:

1. **Base64** can represent any string into a sequence of characters that don't include spaces or the restricted characters. However,  Base64 might still contain characters like / and + which might not be suitable in this case.
2. **Percent(URI)-encoding** converts the character to its corresponding byte value in ASCII and then represent that value as a pair of hexadecimal preceded by a percent sign. Ref: https://en.wikipedia.org/wiki/Percent-encoding

Because percent sign is valid in OpenSearch index name, percent-encoding maybe the preferred solution. Also it preserves the readability for user who wants to access the index directly.

Here is quick test with OpenSearch 2.9:

```
# Example: loggroup/test
# => loggroup%2Ftest   : encode slash
# => loggroup%2ftest   : lowercase f (FlintOSClient is doing this already)
# => loggroup%252ftest : encode % (automatic conversion by REST client)
PUT loggroup%252ftest

GET _cat/indices
# green open loggroup%2ftest    A8F3EVwqSdCseKOLuD6ULQ 5 2       0 0     3kb    1kb
```

In Flint index metadata, there is `source` field which stores the source table name. This can be used to double check in case of any naming conflicts.

**What alternatives have you considered?**

Define some custom encoding rule to solve this specific problem.

seankao-az:
### Description
Handle special characters when generating Flint index name by percent-encoding the invalid OpenSearch index  name characters.

### Issues Resolved
* #192 


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndex.scala
```diff
@@ -79,6 +79,30 @@ object FlintSparkIndex {
...
+
+  /**
+   * Percent-encode invalid OpenSearch index name characters.
+   *
+   * @param indexName
+   *   raw Flint index name
+   * @return
+   *   percent-encoded index name
+   */
+  def percentEncode(indexName: String): String = {
```
dai-chen, line 96: I'm just thinking is it a little better or any problem to do this here in `FlintOpenSearchClient`: https://github.com/opensearch-project/opensearch-spark/blob/main/flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java#L285 (this is previously another restriction by OpenSearch that requires index name all lowercased)

Because this is actually a restriction by OpenSearch. If we have other Flint index store, ex. `FlintS3Client`, we may have different index name restriction?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndex.scala
```diff
@@ -79,6 +79,30 @@ object FlintSparkIndex {
...
+    indexName
+      .flatMap(ch =>
+        if (INVALID_INDEX_NAME_CHARS.contains(ch)) {
+          s"%${ch.toInt.toHexString}"
+        } else {
+          ch.toString
+        })
+  }
```
noCharger, line 104: Since string concatenation in a loop can be costly due to the immutability of strings in Scala, I would prefer using a mutable StringBuilder.

```suggestion
def percentEncode(indexName: String): String = {
  val encoded = new StringBuilder(indexName.length)
  for (ch <- indexName) {
    if (INVALID_INDEX_NAME_CHARS.contains(ch)) {
      encoded.append("%" + ch.toInt.toHexString.toUpperCase)
    } else {
      encoded.append(ch)
    }
  }
  encoded.toString()
}
```

seankao-az: ack! will make the change. Thanks

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
```diff
@@ -287,4 +295,32 @@ private String toLowercase(String indexName) {
...
+   * Percent-encode invalid OpenSearch index name characters.
+   */
+  private String percentEncode(String indexName, Set<Character> charsToEncode) {
+    Objects.requireNonNull(indexName);
+    Objects.requireNonNull(charsToEncode);
+
+    StringBuilder builder = new StringBuilder(indexName.length());
+    for (char ch : indexName.toCharArray()) {
+      if (charsToEncode.contains(ch)) {
+        builder.append("%").append(String.format("%02X", (int) ch));
```
seankao-az, line 309: prefer this over `String.format("%%%02X", (int) ch)` for readability. suggestions are welcome

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
```diff
@@ -287,4 +295,32 @@ private String toLowercase(String indexName) {
 
     return indexName.toLowerCase(Locale.ROOT);
   }
+
+  /*
+   * Percent-encode invalid OpenSearch index name characters.
+   */
+  private String percentEncode(String indexName, Set<Character> charsToEncode) {
```
dai-chen, line 302: np: the second arg is not necessary because field `INVALID_INDEX_NAME_CHARS` is accessible?

dai-chen (approved):
Thanks for the changes!

noCharger (approved):
Lgtm

dai-chen:
FYI, the `main` branch version is already bumped to 0.2. You can add backport label as needed.

seankao-az:
@dai-chen I couldn't merge nor add label 😞

dai-chen:
> @dai-chen I couldn't merge nor add label 😞

No worries. Let me do it. :)

dai-chen:
@seankao-az Auto backport failed. Could you check if any conflict and publish backport PR manually?

PR #219, opened 2024-01-11
[POC]Flint metrics AOP

noCharger:
### Description
_Describe what this change achieves._

### Issues Resolved


flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
```diff
@@ -127,6 +129,7 @@ public void createIndex(String indexName, FlintMetadata metadata) {
     createIndex(indexName, metadata.getContent(), metadata.indexSettings());
   }
 
+  @PublishMetrics(metricNamePrefix=MetricConstants.OS_WRITE_METRIC_PREFIX)
```
dai-chen, line 132: Actually metrics published at this level is per Flint index operation (or per Flint metadata log operation below), right? It's called OS read or write just because currently each only has single OS client call? 

I'm just thinking should we or can we AOP `RestHighLevelClient` instead of changing everywhere we use it?

flint-core/src/main/java/org/opensearch/flint/core/metrics/aop/PublishMetrics.java
```diff
@@ -0,0 +1,12 @@
+package org.opensearch.flint.core.metrics.aop;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.TYPE})
+public @interface PublishMetrics {
```
dai-chen, line 10: Could you add javadoc for all new interfaces and API introduced in this PR?

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchMetadataLog.java
```diff
@@ -113,6 +117,7 @@ public void purge() {
     }
   }
 
+  @PublishMetrics(metricNamePrefix=MetricConstants.OS_WRITE_METRIC_PREFIX)
```
dai-chen, line 120: Does AspectJ also supports AOP on private method like this?

noCharger: yes

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
dai-chen: IMHO, it's better to AOP the client underlying `FlintClient`. Just feel the current implementation is like AOP all callers of FlintClient API to publish metric for OS client underlying (with the assumption above that 1-1 API call mapping from FlintClient API to OS client API).

dai-chen:
Is this PoC PR still needed?

noCharger:
> Is this PoC PR still needed?

Nope, close it

PR #224, opened 2024-01-12
Quote table name with backticks to escape special characters in spark.read.table()

seankao-az:
### Description
Quote table name with backticks to escape special characters in spark.read.table()
Follow up on
* #215 

which allows for creating flint index on table name with special characters, but build() on these index fails if table name not escaped by backticks (``` ` ```)

### Issues Resolved


dai-chen (commented):
Could you check if this also needs change: https://github.com/opensearch-project/opensearch-spark/blob/main/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala#L375C28-L375C28

seankao-az:
Missed that one.. added the change.
Not quite sure how to test this one in UT, though. In IT we cannot create table with special characters due to spark table identifier constraint.

dai-chen:
> Missed that one.. added the change. Not quite sure how to test this one in UT, though. In IT we cannot create table with special characters due to spark table identifier constraint.

No problem. Should be fine if all existing IT can pass. Thanks!

dai-chen (approved):
Thanks for the fix!

Issue #100, opened 2023-10-24
[FEATURE] Support incremental refresh in refresh index statement


dai-chen:
**Is your feature request related to a problem?**

Currently, all refresh index statement only support manual refresh and always pass `RefreshMode.FULL` to underlying `refreshIndex()` API. Re-trigger incremental refresh is also useful when we can support stop incremental refreshing.

**What solution would you like?**

1. Remove RefreshMode param in `refreshIndex()` API [TBD]
2. `refreshIndex()` API decides RefreshMode accordingly to `auto_refresh` option in index metadata

Issue #195, opened 2023-12-13
[FEATURE] Support on-demand triggered incremental refresh


dai-chen:
**Is your feature request related to a problem?**

When the Flint index is created with `auto_refresh=true`, it triggers a long-running Spark streaming job that continuously refreshes the index data. While this setup is ideal for continuous data ingestion scenarios, it becomes inefficient for use cases where data ingestion is infrequent, leading to unnecessary resource utilization, operational cost and user expenses.

**What solution would you like?**

Leveraging the existing `AvailableNow` trigger in Spark structured streaming, the streaming job can autonomously conclude after processing all currently available new data. Later, users can trigger the job manually to fetch the most recent data when needed.

To implement this approach effectively, the following modifications may be necessary:

1. Ensure the Flint index state transits to 'ACTIVE' to prevent potential misinterpretation as a refresh failure
2. Provide an API, either 'RECOVER INDEX JOB,' or a more user-friendly alternative like https://github.com/opensearch-project/opensearch-spark/issues/100, to facilitate user-initiated job restarts

**What alternatives have you considered?**

Stick with current default trigger and user controls by CANCEL/RECOVER statement in https://github.com/opensearch-project/opensearch-spark/issues/190.

PR #228, opened 2024-01-18
Refactor Flint index refresh mode

dai-chen:
### Description

This PR primarily removes the `refreshMode` in the `refreshIndex()` API, which is internally used by the SQL layer. This change serves as a prerequisite for follow-up changes that introduce Manual-Incremental refresh mode for https://github.com/opensearch-project/opensearch-spark/issues/195.

### Before the change

Currently, Flint SQL layer always pass the following mode to refresh index API layer:

- `INCREMENTAL` mode if `CREATE INDEX ... WITH (auto_refresh=true)`
- `FULL` mode if `REFRESH INDEX` statement

#### The meaning of refresh mode

The revision clarifies that the **refresh mode is inherently a characteristic of the Flint index upon creation**, rather than something that can be specified later through either the `refreshIndex` API or the `REFRESH` statement. In other words, the `refreshIndex` API will autonomously determine the mode by reading the Flint index options.

#### Refresh statement behavior

The refresh statement behavior remains unchanged:

- For manual refresh index, a batch refresh job is triggered as before.
- For auto refresh index, the refresh statement will exit as before due to the index state already being in the `refreshing` mode (a new IT has been added to verify this).

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/100

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
@@ -135,10 +135,11 @@ class FlintSpark(val spark: SparkSession) extends Logging {
    * @return
    *   refreshing job ID (empty if batch job for now)
    */
-  def refreshIndex(indexName: String, mode: RefreshMode): Option[String] = {
-    logInfo(s"Refreshing Flint index $indexName with mode $mode")
+  def refreshIndex(indexName: String): Option[String] = {
+    logInfo(s"Refreshing Flint index $indexName")
     val index = describeIndex(indexName)
       .getOrElse(throw new IllegalStateException(s"Index $indexName doesn't exist"))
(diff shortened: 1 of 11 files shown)


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -343,17 +341,13 @@ class FlintSpark(val spark: SparkSession) extends Logging {
     }
 
     val jobId = mode match {
-      case FULL if isIncrementalRefreshing(indexName) =>
-        throw new IllegalStateException(
-          s"Index $indexName is incremental refreshing and cannot be manual refreshed")
```
dai-chen, line 348: Because refresh mode is decided by index options, this case is impossible now.

noCharger (approved):
LGTM

dai-chen:
Merged from `main` and resolved conflicts.

Issue #206, opened 2023-12-22
[FEATURE] Add bloom filter skipping index type


dai-chen:
**Is your feature request related to a problem?**

- https://github.com/opensearch-project/opensearch-spark/issues/197
- https://github.com/opensearch-project/opensearch-spark/issues/202

**What solution would you like?**

## User Experience

Here is the example. See more details in comment below.

<pre>
spark-sql> <b>CREATE SKIPPING INDEX ON stream.test (name BLOOM_FILTER);</b>
spark-sql> <b>REFRESH SKIPPING INDEX ON stream.test;</b>

spark-sql> EXPLAIN SELECT input_file_name() FROM stream.test WHERE name = 'Hello';
== Physical Plan ==
*(1) Project [input_file_name() AS input_file_name()#42]
+- *(1) Filter (isnotnull(name#0) AND (name#0 = hello))
   +- FileScan csv stream.value_set_test[name#0] Batched: false,
  DataFilters: [isnotnull(name#0), (name#0 = hello)], Format: CSV,
  Location: <b>FlintSparkSkippingFileIndex(1 paths)[s3://test],
  PartitionFilters: [], PushedFilters: [IsNotNull(name), EqualTo(name,hello)],</b>
  ReadSchema: struct<name:string>

spark-sql> SELECT input_file_name() FROM stream.test WHERE name = 'hello';
s3://test/part-00000-560ff9f5-6180-4d82-9d34-9d913522f397-c000.csv
Time taken: 10.871 seconds, Fetched 1 row(s)

spark-sql> SELECT input_file_name() FROM stream.test WHERE name = 'test';
<b>Time taken: 0.872 seconds</b>
</pre>

Skipping index data in OpenSearch:

```
POST flint_myglue_stream_value_set_test_skipping_index/_search
        ...
        "_source": {
          "file_path": "s3://test/part-00000-560ff9f5-6180-4d82-9d34-9d913522f397-c000.csv",
          "name": "AAAAAQAAAAUAAABzAAA...="
        }
```

## Proposed Solution

Design decision from problem space to solution space:

![Screenshot 2023-12-22 at 9 09 42 AM](https://github.com/opensearch-project/opensearch-spark/assets/46505291/7c03da19-d6ce-4284-8bc9-1c5ae6740b26)

1. Restrict value set size as proposed above
[...]

PR #242, opened 2024-02-02
Implement BloomFilter skipping index building logic

dai-chen:
### Description

This is the first PR for BloomFilter skipping index support. This PR is focus on bloom filter building side and introduced core classes as below. Please read https://github.com/opensearch-project/opensearch-spark/issues/206 for big picture, including final user experience, design decision, proof of concept and benchmark.

#### PR Planned

1. https://github.com/opensearch-project/opensearch-spark/pull/242 *[Current]*
3. Implement query rewrite logic without pushdown
4. Implement pushdown optimization by OpenSearch painless script
5. Implement AdaptiveBloomFilter algorithm
6. Support bloom filter type in Flint SQL

#### Documentation

Updated user manual: https://github.com/dai-chen/opensearch-spark/blob/add-bloom-filter-building-logic/docs/index.md#feature-highlights

#### Class Diagram

Screenshot 2024-02-06 at 9 42 34 AM

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/206

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #248, opened 2024-02-08
Implement BloomFilter query rewrite (without pushdown optimization)

dai-chen:
### Description

Implemented BloomFilter skipping index query rewrite by introducing the new `BloomFilterMightContain` expression. This internal expression serves to represent BloomFilter queries, aligning with the approach taken in a previous PR with the addition of `BloomFilterAgg`. In the absence of pushdown optimization in the Flint data source, this PR includes updates to the integration tests to validate both code generation and evaluation execution.

#### PR Planned

- [x] https://github.com/opensearch-project/opensearch-spark/pull/242
- [ ] https://github.com/opensearch-project/opensearch-spark/pull/248 *[Current]*
- [ ] Implement pushdown optimization by OpenSearch painless script
- [ ] Implement AdaptiveBloomFilter algorithm
- [ ] Support bloom filter type in Flint SQL

#### Documentation

https://github.com/dai-chen/opensearch-spark/blob/implement-bloom-filter-query-rewrite-no-pushdown/docs/index.md#feature-highlights

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/206


flint-core/src/main/java/org/opensearch/flint/core/field/bloomfilter/classic/ClassicBloomFilter.java
```diff
@@ -132,16 +132,23 @@ public void writeTo(OutputStream out) throws IOException {
    * @param in input stream
    * @return bloom filter
    */
-  public static BloomFilter readFrom(InputStream in) throws IOException {
-    DataInputStream dis = new DataInputStream(in);
+  public static BloomFilter readFrom(InputStream in) {
```
dai-chen, line 135: need to try-catch as Spark codegen doesn't allow checked exception

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/bloomfilter/BloomFilterMightContain.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  override protected def withNewChildrenInternal(
+      newBloomFilterExpression: Expression,
+      newValueExpression: Expression): BloomFilterMightContain =
+    copy(bloomFilterExpression = newBloomFilterExpression, valueExpression = newValueExpression)
+
+  override def eval(input: InternalRow): Any = {
+    val value = valueExpression.eval(input)
+    if (value == null) {
+      null
```
penghuo, line 66: why eval result is null? Should bloomFilter.test(null) return false?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/bloomfilter/BloomFilterMightContain.scala
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  /**
+   * Generate expression code for Spark codegen execution. Sample result code (assume right value
+   * expression is not null):
+   * ```
+   *   boolean filter_isNull_0 = true;
+   *   boolean filter_value_0 = false;
+   *   if (!(false)) {
```
penghuo, line 80: if (!(false)) ? always true?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/bloomfilter/BloomFilterMightContain.scala
dai-chen: Following Spark SQL [NULL semantics](https://spark.apache.org/docs/3.0.0-preview/sql-ref-null-semantics.html#builtin-aggregate-expressions-), NULL is ignored in [BloomFilterAgg]( https://github.com/opensearch-project/opensearch-spark/blob/main/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/bloomfilter/BloomFilterAgg.scala#L58). So NULL is returned for `bloom_filter_might_contain(clientip, NULL)`. Reference: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/BloomFilterMightContain.scala#L100

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/bloomfilter/BloomFilterMightContain.scala
dai-chen: It was assuming right value is not null. Updated in https://github.com/opensearch-project/opensearch-spark/pull/248/commits/ce21393446eaca2013b0668af873f13981d8f8d0. Thanks!

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/bloomfilter/BloomFilterMightContain.scala
dai-chen: As I understand, what's discussed here will happen only if `WHERE clientip = NULL`. We're concerned it's rewritten to `bloom_filter_might_contain(clientip, NULL)` which skips source file by mistake.

I did some test and found out that `col = NULL` will be optimized by Spark directly because it always returns empty result:

```
spark-sql> EXPLAIN SELECT `@timestamp`, request FROM ds_tables.http_logs WHERE clientip = null;
== Physical Plan ==
LocalTableScan <empty>, [@timestamp#5, request#7]
```

penghuo (approved):

PR #253, opened 2024-02-14
Move query from entry point to SparkConf

noCharger:
### Description
Move the query parameter from the entry point's main() arguments to SparkConf for batch and streaming jobs. There is no effect on interactive jobs other than removing the dummy query from the arg, which always reads from the request index.

Breaking change - together with https://github.com/opensearch-project/sql/pull/2519

### Issues Resolved
https://github.com/opensearch-project/sql/issues/2376

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
@@ -19,6 +19,7 @@ import play.api.libs.json._
 import org.apache.spark.SparkConf
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.parser.ParseException
+import org.apache.spark.sql.flint.config.FlintSparkConf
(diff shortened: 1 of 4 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -34,15 +35,16 @@ import org.apache.spark.sql.types.{StructField, _}
...
+      throw new IllegalArgumentException("Usage: FlintJob <resultIndex>")
     }
 
-    val Array(query, resultIndex) = args
+    val Array(resultIndex) = args
 
     val conf = createSparkConf()
     val wait = conf.get("spark.flint.job.type", "continue")
     val dataSource = conf.get("spark.flint.datasource.name", "")
+    val query = conf.get(FlintSparkConf.QUERY.key, "")
```
noCharger, line 47: // TODO: throws exception if conf key not found

PR #262, opened 2024-02-29
Use alternative ARN as credentials when configured

Swiddis:
### Description
Optionally, a user can specify an ARN which will be used for credentials in REPL, instead of the default provider.

### Issues Resolved
N/A


flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
```diff
@@ -77,6 +77,13 @@ object FlintSparkConf {
       .doc("AWS customAWSCredentialsProvider")
       .createWithDefault(FlintOptions.DEFAULT_CUSTOM_AWS_CREDENTIALS_PROVIDER)
 
+  val REPL_FLINT_ASSUME_SESSION_ROLE_ARN =
+    FlintConfig("spark.repl.flint.ASSUME_ROLE_CREDENTIALS_ROLE_ARN")
+      .datasourceOption()
+      .doc("The role to use for writing state information to the session index. " +
```
dai-chen, line 83: Could you update user manual as well? Btw, this doesn't seem to only apply to REPL. It's used by low level Flint client? Should we follow `spark.datasource.flint` prefix?

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
```diff
@@ -133,4 +133,8 @@ public String getPassword() {
   public int getSocketTimeoutMillis() {
     return Integer.parseInt(options.getOrDefault(SOCKET_TIMEOUT_MILLIS, String.valueOf(DEFAULT_SOCKET_TIMEOUT_MILLIS)));
   }
+
+  public String getSessionIndexARN() {
+    return options.getOrDefault("spark.repl.flint.ASSUME_ROLE_CREDENTIALS_ROLE_ARN", "");
```
dai-chen, line 138: np: use the constant?

Swiddis:
Dropping this PR for adding a POC for OSS first, will revisit after the POC is done.

PR #251, opened 2024-02-09
Implement adaptive BloomFilter algorithm

dai-chen:
### Description

This pull request (PR) introduces an enhanced BloomFilter implementation that extends the capabilities of the previously added classic BloomFilter. This new implementation builds the BloomFilter adaptively, determining optimal parameters for construction without relying on prior knowledge of cardinality.

Documentation: user manual will be updated in the next PR of adding SQL support.

- Adaptive BloomFilter parameters
    - num_candidates: by default adaptive algorithm uses 10 candidates
    - fpp: false positive probability (passed to underlying BloomFilter algorithm)
- Classic BloomFilter parameters
    - num_items: expected maximum number of distinct items (NDV)
    - fpp: same as above  

#### PR Planned

- [x] https://github.com/opensearch-project/opensearch-spark/pull/242
- [x] https://github.com/opensearch-project/opensearch-spark/pull/248
- [x] https://github.com/opensearch-project/opensearch-spark/pull/271
- [ ] https://github.com/opensearch-project/opensearch-spark/pull/251 *[Current]*
- [ ] Support bloom filter type in Flint SQL

#### New Classes

1. Added new abstraction `BloomFilterFactory` that creates or deserialize BloomFilter.
2. Added new `AdaptiveBloomFilter` that builds BloomFilter adaptively.

Screenshot 2024-03-11 at 9 36 06 AM

#### Adaptive Algorithm

The Adaptive BloomFilter algorithm dynamically adjusts to varying cardinalities by initially creating 10 BloomFilters, each with a doubled number of expected number of items (NDV, number of distinct values) as candidates. Upon inserting a unique element, the `cardinality` counter increments. Because the BloomFilter's put item result can determine if it's the first time the item has been seen, this is achieved using the put item result of the largest BloomFilter candidate, which is the most accurate.

At last, **the algorithm selects the candidate with an NDV just greater than the current cardinality as the best candidate**. To reduce the overhead introduced by more candidates, candidates with NDV smaller than the best are ignored during the put items or merge operations. In the scenario that the cardinality exceeds the largest candidate's NDV, the algorithm designates the largest candidate as the best, even if its false positive rate (FPP) decreases due to overflow.

In summary, this adaptive approach ensures the use of a BloomFilter with the right size, even in the absence of prior knowledge of cardinality, ensuring optimal performance and accuracy in diverse scenarios.

Screenshot 2024-03-11 at 12 59 17 PM

#### Distributed BloomFilter Aggregation

[...]

PR #271, opened 2024-03-05
Implement BloomFilter query pushdown optimization

dai-chen:
### Description

This PR introduces BloomFilter query pushdown optimization in OpenSearch by using a Painless script. Without dependency on OpenSearch SQL or another new mapper plugin, the following BloomFilter deserialization and membership check code are encoded in a Painless script https://github.com/dai-chen/opensearch-spark/blob/implement-bloom-filter-query-pushdown/flint-spark-integration/src/main/resources/bloom_filter_query.script.

```
  public static BloomFilter readFrom(InputStream in) {
      DataInputStream dis = new DataInputStream(in);
      int version = dis.readInt();
      int numHashFunctions = dis.readInt();
      BitArray bits = BitArray.readFrom(dis);
      ...
  }

 public boolean mightContainLong(long item) {
    int h1 = Murmur3_x86_32.hashLong(item, 0);
    int h2 = Murmur3_x86_32.hashLong(item, h1);

    long bitSize = bits.bitSize();
    for (int i = 1; i <= numHashFunctions; i++) {
      int combinedHash = h1 + (i * h2);
      // Flip all the bits if it's negative (guaranteed positive number)
      if (combinedHash < 0) {
        combinedHash = ~combinedHash;
      }
      if (!bits.get(combinedHash % bitSize)) {
        return false;
      }
    }
    return true;
  }
```

Because Painless language only support basic JDK API, API in `Murmur3`, `DataInputStream` and `ByteArrayInputStream` needs to be inlined accordingly. The Painless script length is 15.4k while OpenSearch default limit is 64k.

#### TODO

If any limitation or performance issue, we can either store the script by https://opensearch.org/docs/latest/api-reference/script-apis/create-stored-script/ or add dependency on SQL or a new plugin.

#### PR Planned

- [x] https://github.com/opensearch-project/opensearch-spark/pull/242
- [x] https://github.com/opensearch-project/opensearch-spark/pull/248
- [ ] https://github.com/opensearch-project/opensearch-spark/pull/271 *[Current]*
- [ ] https://github.com/opensearch-project/opensearch-spark/pull/251
- [ ] Support bloom filter type in Flint SQL

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/206

### DIFF
--- a/flint-spark-integration/src/main/resources/bloom_filter_query.script
+++ b/flint-spark-integration/src/main/resources/bloom_filter_query.script
@@ -1,38 +1,35 @@
 int hashLong(long input, int seed) {
-int low = (int) input;
-int high = (int) (input >>> 32);
-
-int k1 = mixK1(low);
-int h1 = mixH1(seed, k1);
-
-k1 = mixK1(high);
-h1 = mixH1(h1, k1);
-
-return fmix(h1, 8);
+  int low = (int) input;
+  int high = (int) (input >>> 32);
+  int k1 = mixK1(low);
+  int h1 = mixH1(seed, k1);
+  k1 = mixK1(high);
+  h1 = mixH1(h1, k1);
+  return fmix(h1, 8);
 }
 
 int mixK1(int k1) {
-k1 *= 0xcc9e2d51L;
-k1 = Integer.rotateLeft(k1, 15);
-k1 *= 0x1b873593L;
-return k1;
+  k1 *= 0xcc9e2d51L;
+  k1 = Integer.rotateLeft(k1, 15);
+  k1 *= 0x1b873593L;
+  return k1;
 }
 
 int mixH1(int h1, int k1) {
-h1 ^= k1;
-h1 = Integer.rotateLeft(h1, 13);
-h1 = h1 * 5 + (int) 0xe6546b64L;
-return h1;
+  h1 ^= k1;
+  h1 = Integer.rotateLeft(h1, 13);
+  h1 = h1 * 5 + (int) 0xe6546b64L;
+  return h1;
 }
 
 int fmix(int h1, int length) {
-h1 ^= length;
-h1 ^= h1 >>> 16;
-h1 *= 0x85ebca6bL;
-h1 ^= h1 >>> 13;
-h1 *= 0xc2b2ae35L;
-h1 ^= h1 >>> 16;
-return h1;
+  h1 ^= length;
+  h1 ^= h1 >>> 16;
+  h1 *= 0x85ebca6bL;
+  h1 ^= h1 >>> 13;
+  h1 *= 0xc2b2ae35L;
+  h1 ^= h1 >>> 16;
+  return h1;
 }
 
 BytesRef bfBytes = doc[params.fieldName].value;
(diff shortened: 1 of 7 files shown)


flint-spark-integration/src/main/resources/bloom_filter_query.script
```diff
@@ -0,0 +1,109 @@
+int hashLong(long input, int seed) {
```
vamsi-amazon, line 1: We can probably minify this removing spaces and changing variable names. That reduces readability. Not sure if we can add a step in final build before bundling.

vamsi-amazon: Ignore if it is too much optimization and we are way below limits.

dai-chen: Good point. I recall in C there is tool compress/obfuscate the code. Painless is not that popular that I'm not sure if any tool rather than do this manually. Will take another look. Thanks!

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/storage/FlintQueryCompiler.scala
```diff
@@ -112,6 +114,26 @@ case class FlintQueryCompiler(schema: StructType) {
         s"""{"wildcard":{"${compile(p.children()(0))}":{"value":"*${compile(
             p.children()(1),
             false)}"}}}"""
+      case "BLOOM_FILTER_MIGHT_CONTAIN" =>
```
vamsi-amazon, line 117: Now that we have added here..can we remove the BloomFilterMightContain extension created for spark without pushdown?

flint-spark-integration/src/test/scala/org/apache/spark/sql/flint/datatype/FlintDataTypeSuite.scala
```diff
@@ -44,7 +44,8 @@ class FlintDataTypeSuite extends FlintSuite with Matchers {
                           |      "type": "text"
                           |    },
                           |    "binaryField": {
-                          |      "type": "binary"
+                          |      "type": "binary",
+                          |      "doc_values": true
```
vamsi-amazon, line 48: Curious what is this for?

vamsi-amazon (approved):

flint-spark-integration/src/main/resources/bloom_filter_query.script
vamsi-amazon: Since painless is mainly java, I just used this https://www.aspect-ratios.com/minify-java/

penghuo (approved):

flint-spark-integration/src/main/resources/bloom_filter_query.script
dai-chen: I'm thinking of something at runtime when loading or sending to OpenSearch. Otherwise like you pointed out, it will result in code hard to maintain? Will make a note and address it if any issue. Thanks!

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/storage/FlintQueryCompiler.scala
dai-chen: yes, for OpenSearch, `BloomFilterMightContain` will always be pushed down. But if we run it with other index store such as S3 in future, BloomFilterMightContain is still required.

PR #274, opened 2024-03-10
Move query from entry point to SparkConf 

noCharger:
### Description
Move the query parameter from the entry point's main() arguments to SparkConf for batch and streaming jobs. There is no effect on interactive jobs other than removing the dummy query from the arg, which always reads from the request index.

Breaking change - together with https://github.com/opensearch-project/sql/pull/2519

### Issues Resolved
https://github.com/opensearch-project/sql/issues/2376

### DIFF
--- a/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
+++ b/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
@@ -150,6 +150,10 @@ object FlintSparkConf {
     FlintConfig(s"spark.flint.datasource.name")
       .doc("data source name")
       .createOptional()
+  val QUERY =
+    FlintConfig("spark.flint.job.query")
+      .doc("Flint query for batch and streaming job")
+      .createOptional()
   val JOB_TYPE =
     FlintConfig(s"spark.flint.job.type")
       .doc("Flint job type. Including interactive and streaming")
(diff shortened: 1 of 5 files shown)


flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
```diff
@@ -150,6 +150,10 @@ object FlintSparkConf {
     FlintConfig(s"spark.flint.datasource.name")
       .doc("data source name")
       .createOptional()
+  val QUERY =
+    FlintConfig("spark.flint.job.query")
+      .doc("Flint query for batch and streaming job")
+      .createOptional()
```
dai-chen, line 156: This is optional?

dai-chen (approved):
Thanks for the changes!

QQ regarding the dependency between SQL plugin and Spark. Is the changes on plugin side already done?

noCharger:
> Thanks for the changes!
> 
> QQ regarding the dependency between SQL plugin and Spark. Is the changes on plugin side already done?

Yes, it's pending for review https://github.com/opensearch-project/sql/pull/2519

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -37,19 +38,24 @@ import org.apache.spark.sql.types.{StructField, _}
  */
 object FlintJob extends Logging with FlintJobExecutor {
   def main(args: Array[String]): Unit = {
+    CustomLogging.logInfo("Spark Job is Launching...")
     // Validate command line arguments
-    if (args.length != 2) {
-      throw new IllegalArgumentException("Usage: FlintJob <query> <resultIndex>")
+    if (args.length != 1) {
```
penghuo, line 43: it is a breaking change? it required plugin upgrade?

penghuo: could we support plugin in 2.11 version?

noCharger: @penghuo  Yes it's a breaking change. We can keep the signature for backword compatiblity.

penghuo (approved):

Issue #265, opened 2024-03-01
[FEATURE] Show Flint indexes support


seankao-az:
**Is your feature request related to a problem?**
1. Getting a list of all Flint indexes takes too long.
Currently, user needs to make request to show all databases, show all tables, and then get all the indexes from the databases and tables.
2. Cannot get Flint index names right away.
Users need to be aware of the Flint index naming rule and construct the name themselves, using datasource, database, table, index type, and index name.

**What solution would you like?**
provide a query for showing all the flint indexes with their info, including the internal index state.
`SHOW FLINT [INDEX|INDEXES] IN catalog[.database]`

```
SHOW FLINT INDEXES IN glue_1

flint_index_name | kind | database | table | index_name | auto refresh  | status
--------------------------------------------------------------------------------
flint_glue_1_default_http_logs_cv_index | covering | default | http_logs | cv | true | active
flint_glue_1_other_http_logs_skipping_index | skipping | other| http_logs | - | true | refreshing
flint_glue_1_other_http_count_view | mv | other| - | http_count_view | false | active
```

**What alternatives have you considered?**
1. `SHOW FLINT [INDEX|INDEXES]`
Compared to the proposed solution, the intention for this statement is more unclear. It's not clear whether this statement should fetch all indexes in all catalogs or in current catalog only

2. Debated on whether to add FLINT as a keyword.
`SHOW INDEXES` or `SHOW ALL INDEXES` can be confused with the already existing `SHOW INDEXES ON <table>` statement for showing covering indexes, thus adding `FLINT` to make the distinction. In addition, this statement does not return the index name provided by user, but the Flint index name stored in OpenSearch, so that lines up as well.

**Do you have any additional context?**

dai-chen:
This can be closed now?

Issue #121, opened 2023-11-01
[FEATURE] Flint metadata log improvement


dai-chen:
**Is your feature request related to a problem?**

PR https://github.com/opensearch-project/opensearch-spark/pull/110 introduced Flint metadata log initial implementation. In particular, `latestId` field in Flint metadata points to the latest metadata log entry which is a single OpenSearch doc per Flint index in another OpenSearch index. See more details in PR above.

The problem in this simple implementation:

1. Only latest metadata log entry is available without history
2. Index will be always in `REFRESHING` state if auto refreshed
3. Mix metrics in metadata log, e.g. for auto refresh index, `lastUpdateTime` represents heartbeat timestamp

**What solution would you like?**

1. [Low Priority] Append metadata log entry rather than update in-place (this requires dedicated OS index or S3 folder)
2. Figure out how to transit to `REFRESHING` at beginning of each micro batch and transit back to `ACTIVE` once complete (exactly same as manual refresh)
3. Separate metrics such as heartbeat elsewhere and let `lastUpdateTime` only represents created/updated time of metadata log entry

PR #276, opened 2024-03-11
Implement show flint index statement

seankao-az:
### Description
This PR implements the `SHOW FLINT INDEX IN catalog[.database]` statement, with the following output columns:
- flint_index_name: the full OpenSearch index name 
- kind: type of the index (skipping / covering / mv)
- database: database name for the index
- table: table name for skipping and covering index
- index_name: user defined name for covering index and materialized view
- auto_refresh: auto refresh option of the index (true / false)
- status: status of the index. It's fetched from Flint metadata log.

Example:
```
spark-sql> SHOW FLINT INDEXES IN mys3;
flint_mys3_default_http_count_view	mv	default	NULL	http_count_view	false	active
flint_mys3_default_http_logs_skipping_index	skipping	default	http_logs	NULL	false	active
flint_mys3_default_http_logs_status_clientip_and_request_index	covering	default	http_logs	status_clientip_and_request	false	active
```

### Issues Resolved
* #265 


flint-core/src/main/scala/org/opensearch/flint/core/FlintClient.java
```diff
@@ -71,6 +73,15 @@ <T> OptimisticTransaction<T> startTransaction(String indexName, String dataSourc
...
 
+  /**
+   * Retrieve metadata latest log entry for a Flint index.
+   *
+   * @param indexName index name
+   * @param dataSourceName TODO: read from elsewhere in the future
+   * @return index metadata latest log entry
+   */
+  Optional<FlintMetadataLogEntry> getIndexMetadataLatestLogEntry(String indexName, String dataSourceName);
+
```
dai-chen, line 84: I'm thinking is it better to add `FlintMetadataLogEntry` to `FlintMetadata` instead?

1. There is 1-to-1 mapping relation between these 2 model classes (Flint metadata's `latestId` is pointing to the log entry)
2. No need to add new special API for log entry retrieval in `FlintClient` and `FlintSpark`
3. Both maybe merged in future: https://github.com/opensearch-project/opensearch-spark/issues/121

seankao-az: Tried the approach as discussed offline of adding `FlintMetadataLogEntry` to the `FlintMetadata` response in `getAllIndexMetadata`.

One problem I encounter is that `FlintSpark.describeIndexes` does not return the `FlintMetadata` constructed in that function directly. It will create an instance of `FlintSparkIndex` using the metadata. And the log entry field won't be kept around in the new index instance.
https://github.com/opensearch-project/opensearch-spark/blob/aec62ad44ffb50a9057afee661a2c9a8cbc6e298/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala#L174-L184
And later in the AST builder, when I fetch metadata() from the index, the log entry is already lost.

I think piggybacking the log entry in `getAllIndexMetadata` response wouldn't really work. We have to for real make the log entry a part of FlintMetadata and FlintSparkIndex

seankao-az: I'm thinking the AST builder indeed need an API from FlintSpark just for getting the metadata log entry. Is there another way around?

seankao-az: Added metadata log entry to both FlintMetadata and FlintSparkIndex.
Now flint client when getting index metadata, will attempt to get its latest metadata log entry as well.

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkCoveringIndexITSuite.scala
```diff
@@ -63,6 +66,7 @@ class FlintSparkCoveringIndexITSuite extends FlintSparkSuite {
          |       "auto_refresh": "false",
          |       "incremental_refresh": "false"
          |     },
+         |     "latestId": "$testLatestId",
```
dai-chen, line 69: do you know why `latestId` was missing in these IT?

docs/index.md
```diff
@@ -278,6 +278,26 @@ DROP MATERIALIZED VIEW alb_logs_metrics
...
+  - auto_refresh: auto refresh option of the index (true / false)
+  - status: status of the index
+
+```sql
+SHOW FLINT [INDEX|INDEXES] IN catalog[.database]
+```
+
+Example:
+```sql
+SHOW FLINT INDEXES IN spark_catalog.default
```
dai-chen, line 298: We have a long pending task for adding doctest, so almost all example now doesn't have sample result... Could you add some for this?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/index/FlintSparkIndexAstBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+            // Skipping index doesn't have a user defined name
+            case _: FlintSparkSkippingIndex => null
+          }
+
+          val status = index.latestLogEntry match {
+            case Some(entry) => entry.state.toString
+            case None => "unavailable"
+          }
+
+          Row(
```
dai-chen, line 72: I forgot the proposed format in design. Should we also return catalog name?

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
```diff
@@ -78,6 +78,8 @@ public class FlintOptions implements Serializable {
   public static final int DEFAULT_SOCKET_TIMEOUT_MILLIS = 60000;
 
   public static final int DEFAULT_INACTIVITY_LIMIT_MILLIS = 10 * 60 * 1000;
+  
+  public static final String DATA_SOURCE_NAME = "spark.flint.datasource.name";
```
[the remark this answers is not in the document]
seankao-az: I think this is only for datasource options `spark.datasource.flint.<something>`
But here it's `spark.flint.datasource.name`, not for DataSource V2 API.
DATA_SOURCE_NAME is not a datasourceOption
https://github.com/opensearch-project/opensearch-spark/blob/7b4614ae0701d9e8319bc0fe4faa20c509e8ed3a/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala#L144-L152

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/index/FlintSparkIndexAstBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+            // Table name must be qualified when metadata created
+            case _ => parts.drop(2).mkString(".")
+          }
+          val indexName = index match {
+            case covering: FlintSparkCoveringIndex => covering.indexName
+            // MV name must be qualified when metadata created
+            case _: FlintSparkMaterializedView => parts.drop(2).mkString(".")
+            // Skipping index doesn't have a user defined name
+            case _: FlintSparkSkippingIndex => null
+          }
```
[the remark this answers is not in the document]
seankao-az: thanks! that's so much better

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/index/FlintSparkIndexAstBuilder.scala
seankao-az: I don't think so. catalog name is always specified in query.
`show flint index in catalog`
`show flint index in catalog.database`

docs/index.md
seankao-az: added sample result for this.. not entirely sure about the syntax so I'm using https://github.com/opensearch-project/sql/blob/main/docs/user/dql/basics.rst as reference

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkCoveringIndexITSuite.scala
seankao-az: It's not missing per se. It's added because of this PR.
`metadata().getContent` attempts to add latestId if the field exists: https://github.com/opensearch-project/opensearch-spark/pull/276/files#diff-b5f1e45dca03d9b66976fcd74aec975d01f2fd500c2110fb82f58078fc554799R66-R68

And in this PR when adding a log entry to FlintMetadata, latestId is added as well:
https://github.com/opensearch-project/opensearch-spark/pull/276/files#diff-b5f1e45dca03d9b66976fcd74aec975d01f2fd500c2110fb82f58078fc554799R243-R247
In this test suite `getIndexMetadata` successfully adds log entry to FlintMetadata, hence also adding `latestId` to it

docs/index.md
dai-chen: it looks good. thanks!

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkCoveringIndexITSuite.scala
dai-chen: I see. I just debugged to confirm. `latestId` is persisted in Flint index mapping but `FlintMetadata` doesn't set it when deserialization because nowhere use it.

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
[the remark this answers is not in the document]
dai-chen: Ah, I was confused by the naming:

1. Flint data source related config has prefix `spark.datasource.flint.xxx`
2. Flint DataSource name config is `spark.flint.datasource.name`

If so, maybe we use `FlintSparkConf.DATA_SOURCE_NAME.key` directly?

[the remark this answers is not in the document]
seankao-az: It's in a different package that cannot import from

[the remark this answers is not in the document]
dai-chen: nvm. I forgot that

dai-chen (approved):
Thanks for the changes!

PR #288, opened 2024-03-19
Move analyze skipping index rules to config

rupal-bq:
### Description
- Address comments from https://github.com/opensearch-project/opensearch-spark/pull/284

### Issues Resolved
- https://github.com/opensearch-project/opensearch-spark/issues/221 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #284, opened 2024-03-14
Implement analyze skipping index statement

rupal-bq:
### Description
Add `ANALYZE SKIPPING INDEX` statement. This returns recommendation for skipping index based on following rules.
- All top-level columns are selected
- PARTITION algorithm is recommended for partition columns
- MIN_MAX algorithm is recommended for numerical data types columns
- VALUE_SET algorithm is recommended for boolean data type columns
- BLOOM_FILTER algorithm is recommended for all other supported columns
- Unsupported data type columns are skipped.

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/221

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #221, opened 2024-01-11
[FEATURE] Automate skipping index column and algorithm selection


dai-chen:
**Is your feature request related to a problem?**

Currently, users are tasked with the responsibility of determining not only which columns to apply skipping algorithms to but also the specific algorithms to use for each selected column. This demands a nuanced understanding of both:

- the skipping algorithms: their difference and scenario and;
- the attributes of the user's dataset: table size, column type, cardinality etc

**What solution would you like?**

### Automate Index Column Selection

- Choose all top-level columns
- Choose all top-level and nested columns
- Choose first X columns
- ...

### Automate Index Algorithm Selection

From low level to high level solution:

1. Offer an analysis tool for users to gain insights into their dataset.
2. Provide recommendations based on the analysis result and the available skipping index options.
3. Automate this process and make it transparent to user by Flint table or on-demand index build.

Note that for advanced algorithm such as BloomFilter, this also determines algorithm parameters automatically.

**What alternatives have you considered?**

As per [《Big Metadata: When Metadata is Big Data》](https://vldb.org/pvldb/vol14/p3083-edara.pdf), create all applicable skipping data structure on a column.

dai-chen:
Analysis tool quick test:

ANALYZE is very slow and doesn't provide stats per source file:

```
spark-sql> analyze table ds_tables.http_logs compute statistics for all columns;
Time taken: 854.506 seconds

spark-sql> desc extended ds_tables.http_logs status;
col_name	status
data_type	int
comment	NULL
min	200
max	500
num_nulls	0
distinct_count	8
avg_col_len	4
max_col_len	4
histogram	NULL
Time taken: 1.319 seconds, Fetched 10 row(s)

spark-sql> desc extended ds_tables.http_logs clientip;
col_name	clientip
data_type	string
comment	NULL
min	NULL
max	NULL
num_nulls	0
distinct_count	1079058
avg_col_len	11
max_col_len	12
histogram	NULL
Time taken: 1.258 seconds, Fetched 10 row(s)
```

`approx_count_distinct` function is a little faster but there is no table sampling function by number/percentage of file. `tablesample` only samples by rows.

```
SELECT input_file_name(), approx_count_distinct(clientip) FROM ds_tables.http_logs GROUP BY input_file_name();
[...]

Issue #298, opened 2024-04-01
[RFC] Accelerate Spark SQL queries by covering index


dai-chen:
**Is your feature request related to a problem?**

Currently, covering indexes are ingested and exposed directly in OpenSearch. While users can utilize this data for visualization, alerting, reporting, and manage its lifecycle independently, Spark SQL queries on the source table are not accelerated by the covering index data.

**What solution would you like?**

We propose improving the performance of Spark SQL queries by leveraging covering indexes. Specifically, this involves rewriting query with the covering index data from the same source table.

#### Use Cases

After visualizing the data to gain initial insights, users typically seek more in-depth analysis to uncover underlying trends, patterns, and potential root causes driving the observed phenomena. This phase often involves examining various data points, and conducting comprehensive root cause analysis. The proposed feature aims to expedite this in-depth analysis process by utilizing the pre-existing covering index data within OpenSearch, thereby eliminating the necessity for redundant scans of the source data.

#### Open questions regarding this proposal:

1. **Use Case**: Is it acceptable for user to query the OpenSearch index directly as proposed in Alternative section below? Is it more natural because visualization is built on the index and DSL query too. Or Spark SQL query experience is required?
2. **Consistency**:
    a. Currently, we only guarantee eventual consistency, and the freshness of OpenSearch index data is determined by the latency of index refresh.
    b. Users may face difficulty viewing the latest data if query rewrite is enabled always
4. **Data Management**: Differently from the current mental model, the index will be exposed to both external customers and internal use. For example, if users delete the OpenSearch index data fully or partially, it may impact the correctness of the rewritten query.
5. **Security**: If query rewrite happens in Spark and any access control enabled, could this query rewrite pose a security risk?
6. **Performance**: It is uncertain whether there will be an improvement in latency if query rewriting occurs in Spark. Comparing *scan S3 directly* vs. *scan OpenSearch index in Spark and transmit data to OpenSearch back*, it depends on the covering index size, query result size and predicate pushdown to OpenSearch.
7. **Dev Efforts**: One of the challenge is how to rewrite query using partial covering index. Proof of concept is necessary to evaluate the feasibility.

**What alternatives have you considered?**

Alternatively, users can:

1. Query the index data directly via OpenSearch DSL or SQL; or
2. Query the OpenSearch table directly via SparkSQL (PoC item 1)

**Do you have any additional context?**

Most of the discussion here applies to query rewrite for materialized views. For clarity, this will be discussed separately.

[...]

PR #300, opened 2024-04-01
Add skipping index recommendations for specific columns

rupal-bq:
### Description
- Address comments from https://github.com/opensearch-project/opensearch-spark/pull/284 and https://github.com/opensearch-project/opensearch-spark/pull/288
- Add support for `ANALYZE SKIPPING INDEX ON TABLE datasource.database.table(column1, column2, ...)`

### Issues Resolved
- #221


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -369,13 +369,13 @@ class FlintSpark(val spark: SparkSession) extends Logging {
...
-   * @param tableName
-   *   table name
+   * @param inputs
+   *   inputs for recommendation strategy. This can table name, columns or functions.
    * @return
    *   skipping index recommendation dataframe
    */
-  def analyzeSkippingIndex(tableName: String): Seq[Row] = {
-    new DataTypeSkippingStrategy().analyzeSkippingIndexColumns(tableName, spark)
+  def analyzeSkippingIndex(inputs: Map[String, List[String]]): Seq[Row] = {
```
dai-chen, line 377: Could you use some abstraction here instead of generic Map? `DataFrame` or any existing Table abstraction we can use here?

Because you may ignore query/function as input for now. The reason is in Limitation no.1 in https://github.com/opensearch-project/opensearch-spark/issues/298#issuecomment-2030554530. I'm thinking can we add generic query analyze API for all Flint index, ex. `ANALYZE FLINT INDEX FOR query`.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/recommendations/RecommendationRules.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package org.opensearch.flint.spark.skipping.recommendations
+
+import com.typesafe.config.{Config, ConfigFactory}
+
+/**
+ * Recommendation rules for skipping index column and algorithm selection.
+ */
+class RecommendationRules {
```
dai-chen, line 13: either add some fields or make this Scala Object?

rupal-bq: changed to Scala Object

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
rupal-bq: raised revision with `DataFrame`. bdw should grammar be `ANALYZE FLINT INDEX` or `ANALYZE SKIPPING INDEX`? Can same static rules apply to any type of index?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -369,13 +370,16 @@ class FlintSpark(val spark: SparkSession) extends Logging {
...
-   * @param tableName
-   *   table name
+   * @param data
+   *   data for recommendation strategy.
    * @return
    *   skipping index recommendation dataframe
    */
-  def analyzeSkippingIndex(tableName: String): Seq[Row] = {
-    new DataTypeSkippingStrategy().analyzeSkippingIndexColumns(tableName, spark)
+  def analyzeSkippingIndex(schema: StructType, data: Seq[Row]): Seq[Row] = {
```
dai-chen, line 378: `SparkSession` is available in this class. I think the input of this API can be simply tableName and columnNames? This is convenient for users who rely on Flint API instead of SQL layer.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/sql/skipping/FlintSparkSkippingIndexAstBuilder.scala
```diff
@@ -133,8 +133,21 @@ trait FlintSparkSkippingIndexAstBuilder extends FlintSparkSqlExtensionsVisitor[A
...
       AttributeReference("skipping_type", StringType, nullable = false)())
 
+    var data = Seq.empty[Row]
+    if (ctx.indexColumns != null) {
+      ctx.indexColumns.multipartIdentifierProperty().forEach { indexColCtx =>
+        data = data :+ Row(ctx.tableName().getText, indexColCtx.multipartIdentifier().getText)
+      }
+    } else {
+      data = data :+ Row(ctx.tableName().getText, null.asInstanceOf[String])
+    }
```
dai-chen, line 143: use Scala stream `map()` instead of `forEach()`?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/recommendations/DataTypeSkippingStrategy.scala
```diff
@@ -7,32 +7,67 @@ package org.opensearch.flint.spark.skipping.recommendations
...
+          .map(_.getString(0))
+          .toList
+          .filter(_ != null)
 
-    val partitionFields = table.partitioning().flatMap { transform =>
+        if (columns.isEmpty) {
+          columns = table.schema().fields.map(field => field.name).toList
+        }
+        columns.foreach(column => {
+          val field = findField(table.schema(), column).get
```
dai-chen, line 50: Could you refactor this method and make it more readable? I think only line 50 - 62 is the core logic.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/skipping/recommendations/RecommendationRules.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package org.opensearch.flint.spark.skipping.recommendations
+
+import com.typesafe.config.{Config, ConfigFactory}
+
+/**
+ * Recommendation rules for skipping index column and algorithm selection.
+ */
+object RecommendationRules {
```
dai-chen, line 13: Just some thought: making this Rule abstraction may be more useful than static util methods?
Probably we can think about how to extend this for recommendation on WHERE clause?

rupal-bq: Sure that can be useful if have separate implementation for data type and function based rules, but I was thinking to have all static rules at one place. e.g https://github.com/rupal-bq/opensearch_spark/blob/query-recommendations/flint-spark-integration/src/main/resources/skipping_index_recommendation.conf#L50

Do you see any problem with this approach for recommendation on WHERE clause?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
dai-chen: Just some thoughts. I was thinking of something like `ANALYZE FLINT INDEX FOR query`. For example:

1. `ANALYZE FLINT INDEX FOR SELECT * FROM test`: recommend covering index
2. `ANALYZE FLINT INDEX FOR SELECT ... WHERE clientip = ...`: recommend skipping and covering
3. `ANALYZE FLINT INDEX FOR SELECT ... GROUP BY ...`: recommend MV

dai-chen:
Closing this PR due to prolonged inactivity. Please rebase if you wish to reopen it.

PR #301, opened 2024-04-02
Adding support to run integ tests on iceberg tables

asuresh8:
### Description
This change adds the capability to run integration tests on iceberg tables. 

### Testing
```
sbt "integtest/test" 

[info] Run completed in 34 minutes, 32 seconds.
[info] Total number of tests run: 302
[info] Suites: completed 29, aborted 0
[info] Tests: succeeded 302, failed 0, canceled 0, ignored 1, pending 0
[info] All tests passed.
```

### DIFF
--- a/integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
+++ b/integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
@@ -38,7 +38,7 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
 
   override def beforeEach(): Unit = {
     super.beforeEach()
-    createPartitionedMultiRowTable(testTable)
+    createPartitionedMultiRowAddressTable(testTable)
   }
 
   override def afterEach(): Unit = {
--- a/integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
+++ b/integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
@@ -32,7 +32,7 @@ import org.apache.spark.sql.internal.SQLConf
 class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
 
   /** Test table and index name */
-  private val testTable = "spark_catalog.default.test"
+  private val testTable = "spark_catalog.default.skipping_test"
   private val testIndex = getSkippingIndexName(testTable)
   private val testLatestId = Base64.getEncoder.encodeToString(testIndex.getBytes)
 
@@ -42,11 +42,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   override def afterEach(): Unit = {
-    super.afterEach()
-
     // Delete all test indices
     deleteTestIndex(testIndex)
     sql(s"DROP TABLE $testTable")
+    super.afterEach()
   }
 
   test("create skipping index with metadata successfully") {
@@ -63,7 +62,7 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
     index shouldBe defined
     index.get.metadata().getContent should matchJson(s"""{
         |   "_meta": {
-        |     "name": "flint_spark_catalog_default_test_skipping_index",
+        |     "name": "flint_spark_catalog_default_skipping_test_skipping_index",
         |     "version": "${current()}",
         |     "kind": "skipping",
         |     "indexedColumns": [
@@ -101,7 +100,7 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
         |        "columnName": "name",
         |        "columnType": "string"
         |     }],
-        |     "source": "spark_catalog.default.test",
+        |     "source": "spark_catalog.default.skipping_test",
         |     "options": {
         |       "auto_refresh": "false",
         |       "incremental_refresh": "false"
@@ -323,6 +322,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   test("should not rewrite original query if no skipping index") {
+    assume(
+      System.getProperty("TABLE_TYPE") != "iceberg",
+      """Test disabled for iceberg because Iceberg tables have a built-in
+        skipping index which rewrites queries""")
     val query =
       s"""
          | SELECT name
@@ -338,6 +341,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   test("should not rewrite original query if skipping index is logically deleted") {
+    assume(
+      System.getProperty("TABLE_TYPE") != "iceberg",
+      """Test disabled for Iceberg because Iceberg tables have a built-in
+        skipping index which rewrites queries""")
     flint
       .skippingIndex()
       .onTable(testTable)
@@ -360,6 +367,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   test("can build partition skipping index and rewrite applicable query") {
+    assume(
+      System.getProperty("TABLE_TYPE") != "iceberg",
+      """Test disabled for Iceberg because Iceberg tables have a built-in
+        skipping index which rewrites queries""")
     flint
       .skippingIndex()
       .onTable(testTable)
@@ -385,6 +396,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   test("can build value set skipping index and rewrite applicable query") {
+    assume(
+      System.getProperty("TABLE_TYPE") != "iceberg",
+      """Test disabled for Iceberg because Iceberg tables have a built-in
+        skipping index which rewrites queries""")
     flint
       .skippingIndex()
       .onTable(testTable)
@@ -414,6 +429,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   test("can build min max skipping index and rewrite applicable query") {
+    assume(
+      System.getProperty("TABLE_TYPE") != "iceberg",
+      """Test disabled for Iceberg because Iceberg tables have a built-in
+        skipping index which rewrites queries""")
     flint
       .skippingIndex()
       .onTable(testTable)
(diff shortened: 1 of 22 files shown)


dai-chen (commented):
A high level question: Should we add a new Iceberg IT suite separately without impact on existing ITs? Because:

1. Iceberg is optional at runtime and should not add it in Flint base suite as prerequisite?
2. Existing IT may cover something unsupported by Iceberg, such as features from Spark built-in `FileStreamSource`

asuresh8:
Agree on both your points! That's why I refactored the integration tests to support additional tables types but by default they just run as the CSV type they were running as before. I think once we have skipping index logic specific to Iceberg then we can visit actually running this test continuously and how to do that.

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSuite.scala
```diff
@@ -5,41 +5,74 @@
...
+  lazy protected val tableType: String = Option(System.getProperty("TABLE_TYPE")).getOrElse("CSV")
+  lazy protected val tableOptions: String =
+    TableOptions.opts.getOrElse(tableType.toLowerCase(), "")
 
-  override def beforeAll(): Unit = {
-    super.beforeAll()
+  override protected def sparkConf: SparkConf = {
+    val conf = super.sparkConf
+      .set("spark.sql.catalog.spark_catalog.type", "hadoop")
+      .set("spark.sql.catalog.spark_catalog.warehouse", s"spark-warehouse/${suiteName}")
```
dai-chen, line 54: what are these new config for?

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
```diff
@@ -322,6 +321,10 @@ class FlintSparkSkippingIndexITSuite extends FlintSparkSuite {
   }
 
   test("should not rewrite original query if no skipping index") {
+    assume(
+      System.getProperty("TABLE_TYPE") != "iceberg",
+      """Test disabled for iceberg because Iceberg tables have a built-in
+        skipping index which rewrites queries""")
```
dai-chen, line 327: Any timeline when we will add this support?

integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTimeWindowITSuite.scala
```diff
@@ -274,12 +232,16 @@ class FlintSparkPPLTimeWindowITSuite
...
-    implicit val timestampOrdering: Ordering[Timestamp] = new Ordering[Timestamp] {
-      def compare(x: Timestamp, y: Timestamp): Int = x.compareTo(y)
+
+    // Define ordering for rows that first compares by the timestamp and then by the productId
+    implicit val rowOrdering: Ordering[Row] = new Ordering[Row] {
+      def compare(x: Row, y: Row): Int = {
+        val dateCompare = x.getAs[Timestamp](2).compareTo(y.getAs[Timestamp](2))
+        if (dateCompare != 0) dateCompare
+        else x.getAs[String](1).compareTo(y.getAs[String](1))
+      }
```
dai-chen, line 242: what make this change required?

dai-chen (commented):
I'm a little concern about the approach that we run all the IT against different source file format. In the existing tests, we create CSV, JSON, Parquet table which is fine because all use the same `FileStreamSource`. For Iceberg, do you have a full list of streaming options it support? I don't find it in the documentation.

Meanwhile I checked the other acceleration framework Hyperspace which supports File, Delta, Iceberg. But I don't see they mix or share the test code for different source.

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
dai-chen: Also in future if any test cannot work for all the file formats, we have to add this assume?

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSuite.scala
asuresh8: At least for iceberg and hive the catalog type has to be set as hadoop otherwise it assumes hive metastore but for unit tests we don't want to use hive metastore.

integ-test/src/test/scala/org/opensearch/flint/spark/FlintSparkSkippingIndexITSuite.scala
asuresh8: It will be in the next few months. I was told to prioritize Hive, so that's coming next and then will add Iceberg. 

We can probably do something more extensible than this assume. We can either make a helper method or add tags, but I think is fine to start with.

integ-test/src/test/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTimeWindowITSuite.scala
asuresh8: The ordering was indeterministic. For Iceberg tables those tests failed because records with the exact same timestamps but different other columns were flipped in order.

asuresh8:
> I'm a little concern about the approach that we run all the IT against different source file format. In the existing tests, we create CSV, JSON, Parquet table which is fine because all use the same `FileStreamSource`. For Iceberg, do you have a full list of streaming options it support? I don't find it in the documentation.
> 
> Meanwhile I checked the other acceleration framework Hyperspace which supports File, Delta, Iceberg. But I don't see they mix or share the test code for different source.

Iceberg supports micro batch streaming read. I'm not sure what you mean about file types. Iceberg uses parquet. 

I don't see why we can't share test code between sources. Like if you create a skipping index for a hive table or iceberg table and test that the skipping index was created correctly or test that the query was accelerated correctly, the only different in those two tests is what kind of table you put in the SQL query.

dai-chen:
Actually most test are not related to different source format. I'm thinking can we make Iceberg test capability a mix-in (similar like this? https://github.com/dai-chen/opensearch-spark/blob/improve-index-validation/integ-test/src/test/scala/org/apache/spark/sql/SparkHiveSupportSuite.scala) and only enable it for `FlintSparkSkippingIndexSqlITSuite`, `FlintSparkCoveringIndexSqlITSuite`, `FlintSparkMaterializedSqlITSuite`. I think these are the main IT focus on index create and refresh on different data source.

Later we only enable it on more test as needed or move it to Flint test base suite. Does this make sense?

asuresh8:
> Actually most test are not related to different source format. I'm thinking can we make Iceberg test capability a mix-in (similar like this? https://github.com/dai-chen/opensearch-spark/blob/improve-index-validation/integ-test/src/test/scala/org/apache/spark/sql/SparkHiveSupportSuite.scala) and only enable it for `FlintSparkSkippingIndexSqlITSuite`, `FlintSparkCoveringIndexSqlITSuite`, `FlintSparkMaterializedSqlITSuite`. I think these are the main IT focus on index create and refresh on different data source.
> 
> Later we can enable it on more tests as needed or move it to Flint test base suite. Does this make sense?

Yep, I have updated the PR. Thanks for the guidance!

dai-chen (approved):
Thanks for the changes!

integ-test/src/test/scala/org/opensearch/flint/spark/iceberg/FlintSparkIcebergSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  // You can also override tableOptions if Iceberg requires different options
+  override lazy protected val tableOptions: String = ""
+
+  // Override the sparkConf method to include Iceberg-specific configurations
+  override protected def sparkConf: SparkConf = {
+    val conf = super.sparkConf
+      // Set Iceberg-specific Spark configurations
+      .set("spark.sql.catalog.spark_catalog", "org.apache.iceberg.spark.SparkSessionCatalog")
+      .set("spark.sql.catalog.spark_catalog.type", "hadoop")
```
vamsi-amazon, line 30: This is hadoop because integ tests doesn't have any local hive metastore. For emr-serverless environment, this would be hive right and it works well.

asuresh8: Yes that is correct

vamsi-amazon:
@asuresh8 
Given that our application now supports Iceberg tables, wouldn't it be more appropriate to include the Iceberg runtime dependency in the main package itself, rather than limiting it to the runtime scope of the integration tests.

The consumers could specify the catalog class when they want to use iceberg but expecting them to include iceberg every job submission. is this a common practice?

asuresh8:
Agreed. That was the original approach, but Chen felt like this was better as there is no Iceberg specific code in Flint.

vamsi-amazon:
> Agreed. That was the original approach, but Chen felt like this was better as there is no Iceberg specific code in Flint.

https://github.com/microsoft/hyperspace/pull/358

I saw the hyperspace pull request for iceberg support. It seems like expecting the packages were added as provided dependencies. So it feels like the burden is on the spark job executor to provide these dependencies. Not sure of the rationale behind this.

vamsi-amazon:
> > I'm a little concern about the approach that we run all the IT against different source file format. In the existing tests, we create CSV, JSON, Parquet table which is fine because all use the same `FileStreamSource`. For Iceberg, do you have a full list of streaming options it support? I don't find it in the documentation.
> > Meanwhile I checked the other acceleration framework Hyperspace which supports File, Delta, Iceberg. But I don't see they mix or share the test code for different source.
> 
> Iceberg supports micro batch streaming read. I'm not sure what you mean about file types. Iceberg uses parquet.
> 
> I don't see why we can't share test code between sources. Like if you create a skipping index for a hive table or iceberg table and test that the skipping index was created correctly or test that the query was accelerated correctly, the only different in those two tests is what kind of table you put in the SQL query.

@asuresh8 
I couldn't comprehend @dai-chen's concern here: but I am more worried about sql plugin changes where we have changed the default catalog in all queries. Would this impact on adding auto refresh indices(leverages streaming jobs) on top of non-iceberg tables with different file format. Have we added integ tests verifying non-iceberg tables with different file formats ?

asuresh8:
Changing to the default catalog to iceberg catalog is not a breaking change because it defaults to default catalog for non iceberg tables which is the exact same behavior that Flint does today. One of the iterations of this PR I had default catalog set to iceberg catalog for all tests as well and they all passed.

PR #303, opened 2024-04-04
Introduce aws sigv4a request signer

noCharger:
### Description
* Introduce `AWSRequestSigV4ASigningApacheInterceptor` to sign metadata access with sigV4a
* Refactor `ResourceBasedAWSRequestSigningApacheInterceptor` interface

Current signer:
```
24/04/04 20:53:48 INFO AWSRequestSigningApacheInterceptor: unsigned request: GET /query_results2?master_timeout=30s&ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open%2Cclosed&allow_no_indices=false HTTP/1.1 [Content-Length: 0, Host: xxx, Connection: Keep-Alive, User-Agent: Apache-HttpAsyncClient/4.1.5 (Java/17.0.9)]
24/04/04 20:53:48 INFO AWSRequestSigningApacheInterceptor: signed request: GET /query_results2?master_timeout=30s&ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open%2Cclosed&allow_no_indices=false HTTP/1.1 [Authorization: AWS4-HMAC-SHA256 Credential=xxx, SignedHeaders=host;user-agent;x-amz-date;x-amz-security-token, Signature=xxx, Connection: Keep-Alive, Host: xxx, User-Agent: Apache-HttpAsyncClient/4.1.5 (Java/17.0.9), X-Amz-Date: xxx, X-Amz-Security-Token: xxx]
```

New sigV4a signer
```
24/04/04 20:02:46 INFO AWSRequestSigV4ASigningApacheInterceptor: before: HEAD /flint_ql_sessions?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open%2Cclosed&allow_no_indices=false HTTP/1.1 [Host: xxx, Connection: Keep-Alive, User-Agent: Apache-HttpAsyncClient/4.1.5 (Java/17.0.9)]
24/04/04 20:02:46 INFO AWSRequestSigV4ASigningApacheInterceptor: after: HEAD /flint_ql_sessions?ignore_throttled=false&ignore_unavailable=false&expand_wildcards=open%2Cclosed&allow_no_indices=false HTTP/1.1 [Authorization: AWS4-ECDSA-P256-SHA256 Credential=xxx, SignedHeaders=connection;host;x-amz-date;x-amz-region-set;x-amz-security-token, Signature=xxx, Connection: Keep-Alive, Host: xxx, User-Agent: Apache-HttpAsyncClient/4.1.5 (Java/17.0.9), X-Amz-Date: 20240404T200246Z, X-Amz-Region-Set: eu-west-1, X-Amz-Security-Token: xxx]
```

### DIFF
--- a/flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigV4ASigningApacheInterceptor.java
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigV4ASigningApacheInterceptor.java
@@ -1,0 +1,228 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.flint.core.auth;
+
+import com.amazonaws.auth.AWSSessionCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.glue.model.InvalidStateException;
+import org.apache.http.Header;
+import org.apache.http.HttpEntity;
+import org.apache.http.HttpEntityEnclosingRequest;
+import org.apache.http.HttpHost;
+import org.apache.http.HttpRequest;
+import org.apache.http.HttpRequestInterceptor;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.entity.BasicHttpEntity;
+import org.apache.http.message.BasicHeader;
+import org.apache.http.protocol.HttpContext;
+import software.amazon.awssdk.auth.credentials.AwsSessionCredentials;
+import software.amazon.awssdk.auth.signer.AwsSignerExecutionAttribute;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.signer.Signer;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.regions.Region;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import static org.apache.http.protocol.HttpCoreContext.HTTP_TARGET_HOST;
+import static org.opensearch.flint.core.auth.AWSRequestSigningApacheInterceptor.nvpToMapParams;
+import static org.opensearch.flint.core.auth.AWSRequestSigningApacheInterceptor.skipHeader;
+
+/**
+ * Interceptor for signing AWS requests according to Signature Version 4A.
+ * This interceptor processes HTTP requests, signs them with AWS credentials,
+ * and updates the request headers to include the signature.
+ */
+public class AWSRequestSigV4ASigningApacheInterceptor implements HttpRequestInterceptor {
+    private static final String HTTPS_PROTOCOL = "https";
+    private static final int HTTPS_PORT = 443;
+
+    private final String service;
+    private final String region;
+    private final Signer signer;
+    private final AWSCredentialsProvider awsCredentialsProvider;
+
+    /**
+     * Constructs an interceptor for AWS request signing with metadata access.
+     *
+     * @param service                The AWS service name.
+     * @param region                 The AWS region for signing.
+     * @param signer                 The signer implementation.
+     * @param awsCredentialsProvider The credentials provider for metadata access.
+     */
+    public AWSRequestSigV4ASigningApacheInterceptor(String service, String region, Signer signer, AWSCredentialsProvider awsCredentialsProvider) {
+        this.service = service;
+        this.region = region;
+        this.signer = signer;
+        this.awsCredentialsProvider = awsCredentialsProvider;
+    }
+
+    /**
+     * Processes and signs an HTTP request, updating its headers with the signature.
+     *
+     * @param request the HTTP request to process and sign.
+     * @param context the HTTP context associated with the request.
+     * @throws IOException if an I/O error occurs during request processing.
+     */
+    @Override
+    public void process(HttpRequest request, HttpContext context) throws IOException {
+        SdkHttpFullRequest requestToSign = buildSdkHttpRequest(request, context);
+        SdkHttpFullRequest signedRequest = signRequest(requestToSign);
+        updateRequestHeaders(request, signedRequest.headers());
+        updateRequestEntity(request, signedRequest);
+    }
+
+    /**
+     * Builds an {@link SdkHttpFullRequest} from the Apache {@link HttpRequest}.
+     *
+     * @param request the HTTP request to process and sign.
+     * @param context the HTTP context associated with the request.
+     * @return an SDK HTTP request ready to be signed.
+     * @throws IOException if an error occurs while building the request.
+     */
+    private SdkHttpFullRequest buildSdkHttpRequest(HttpRequest request, HttpContext context) throws IOException {
+        URIBuilder uriBuilder = parseUri(request);
+        SdkHttpFullRequest.Builder builder = SdkHttpFullRequest.builder()
+                .method(SdkHttpMethod.fromValue(request.getRequestLine().getMethod()))
+                .protocol(HTTPS_PROTOCOL)
+                .port(HTTPS_PORT)
+                .headers(headerArrayToMap(request.getAllHeaders()))
+                .rawQueryParameters(nvpToMapParams(uriBuilder.getQueryParams()));
+
+        HttpHost host = (HttpHost) context.getAttribute(HTTP_TARGET_HOST);
+        if (host == null) {
+            throw new InvalidStateException("Host must not be null");
+        }
+        builder.host(host.getHostName());
+        try {
+            builder.encodedPath(uriBuilder.build().getRawPath());
+        } catch (URISyntaxException e) {
+            throw new IOException("Invalid URI", e);
+        }
+        setRequestEntity(request, builder);
+        return builder.build();
+    }
+
+    /**
+     * Sets the request entity for the {@link SdkHttpFullRequest.Builder} if the original request contains an entity.
+     * This is used for requests that have a body, such as POST or PUT requests.
+     *
+     * @param request the original HTTP request.
+     * @param builder the SDK HTTP request builder.
+     */
+    private void setRequestEntity(HttpRequest request, SdkHttpFullRequest.Builder builder) {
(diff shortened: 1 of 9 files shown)


shikharj05 (commented):
Thanks for the changes! I have left some comments.

flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigV4ASigningApacheInterceptor.java
```diff
@@ -0,0 +1,157 @@
...
+    public void process(HttpRequest request, HttpContext context) throws IOException {
+        SdkHttpFullRequest requestToSign = buildSdkHttpRequest(request, context);
+        SdkHttpFullRequest signedRequest = signRequest(requestToSign);
+        updateRequestHeaders(request, signedRequest.headers());
+    }
+
+    private SdkHttpFullRequest buildSdkHttpRequest(HttpRequest request, HttpContext context) throws IOException {
+        URIBuilder uriBuilder = parseUri(request);
+        SdkHttpFullRequest.Builder builder = SdkHttpFullRequest.builder()
+                .method(SdkHttpMethod.fromValue(request.getRequestLine().getMethod()))
```
shikharj05, line 82: this is missing payload to be copied from the original request?

flint-core/src/main/scala/org/opensearch/flint/core/auth/ResourceBasedAWSRequestSigningApacheInterceptor.java
```diff
@@ -19,33 +25,36 @@ public class ResourceBasedAWSRequestSigningApacheInterceptor implements HttpRequ
...
+                                                           final String region,
                                                            final AWSCredentialsProvider primaryCredentialsProvider,
                                                            final AWSCredentialsProvider metadataAccessCredentialsProvider,
                                                            final String metadataAccessIdentifier) {
-        this(service,
-                new AWSRequestSigningApacheInterceptor(service, signer, primaryCredentialsProvider),
-                new AWSRequestSigningApacheInterceptor(service, signer, metadataAccessCredentialsProvider),
-                metadataAccessIdentifier);
+        this.service = service == null ? "unknown" : service;
+        AWS4Signer signer = new AWS4Signer();
```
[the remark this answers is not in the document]
noCharger: > shouldn't this be [DefaultAwsCrtV4aSigner](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/authcrt/signer/internal/DefaultAwsCrtV4aSigner.html)?

this is the original signer for `AWSRequestSigningApacheInterceptor `, the v4a signer is used for `AWSRequestSigV4ASigningApacheInterceptor `

flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigV4ASigningApacheInterceptor.java
noCharger: > this is missing payload to be copied from the original request?

SdkHttpFullRequest does not contain a payload in the creation path; could you perhaps provide an example if I missed anything?

flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigV4ASigningApacheInterceptor.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Builds an {@link SdkHttpFullRequest} from the Apache {@link HttpRequest}.
+     *
+     * @param request the HTTP request to process and sign.
+     * @param context the HTTP context associated with the request.
+     * @return an SDK HTTP request ready to be signed.
+     * @throws IOException if an error occurs while building the request.
+     */
+    private SdkHttpFullRequest buildSdkHttpRequest(HttpRequest request, HttpContext context) throws IOException {
```
vamsi-amazon, line 94: Are we creating this new SdkHttpFullRequest fake request because the signer works only on SdkHttpFullRequest?
Don't we have any existing aws libraries which exposes this interceptor?

flint-core/src/main/scala/org/opensearch/flint/core/auth/ResourceBasedAWSRequestSigningApacheInterceptor.java
```diff
@@ -19,33 +27,45 @@ public class ResourceBasedAWSRequestSigningApacheInterceptor implements HttpRequ
...
+        }
+        this.service = service;
+        this.metadataAccessIdentifier = metadataAccessIdentifier;
+        AWS4Signer signer = new AWS4Signer();
+        signer.setServiceName(service);
+        signer.setRegionName(region);
+        this.primaryInterceptor = new AWSRequestSigningApacheInterceptor(service, signer, primaryCredentialsProvider);
+        this.metadataAccessInterceptor = primaryCredentialsProvider.equals(metadataAccessCredentialsProvider)
+                ? this.primaryInterceptor
+                : new AWSRequestSigV4ASigningApacheInterceptor(service, region, AwsCrtV4aSigner.builder().build(), metadataAccessCredentialsProvider);
```
vamsi-amazon, line 61: Why is there a builder for AwsCrtV4asigner? Waht are the parameters available? Are we good with defaults. I see 
plain constructor for `AWS4Signer signer = new AWS4Signer();`

Also when should we use AwsCrtV4asigner insread of AWS4Signer. what are the downsides of making AwsCrtV4asigner default.

vamsi-amazon (approved):

flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigV4ASigningApacheInterceptor.java
noCharger: > Are we creating this new SdkHttpFullRequest fake request because the signer works only on SdkHttpFullRequest? Don't we have any existing aws libraries which exposes this interceptor?

`HttpRequestInterceptor` from the Apache standard library offers the interface `process(HttpRequest request, HttpContext context)` for intercepting requests. Meanwhile, signers have their own signing interface, which may not compatible with standard Apache HTTP requests. If anything is built upon the Apache standard library, we can smoothly transition.

flint-core/src/main/scala/org/opensearch/flint/core/auth/ResourceBasedAWSRequestSigningApacheInterceptor.java
noCharger: > Why is there a builder for AwsCrtV4asigner? Waht are the parameters available? Are we good with defaults. I see plain constructor for `AWS4Signer signer = new AWS4Signer();`
> 
> Also when should we use AwsCrtV4asigner insread of AWS4Signer. what are the downsides of making AwsCrtV4asigner default.

AwsCrtV4aSigner is an interface that employs the builder pattern to instantiate. However, the AWS4Signer class has its own getter and setter. 
```
@SdkPublicApi
@Immutable
@ThreadSafe
public interface AwsCrtV4aSigner extends Signer, Presigner {

    /**
     * Create a default Aws4aSigner.
     */
    static AwsCrtV4aSigner create() {
        return DefaultAwsCrtV4aSigner.create();
    }

    static Builder builder() {
        return DefaultAwsCrtV4aSigner.builder();
    }

    interface Builder {
        /**
         * The region scope that this signer will default to if not provided explicitly when the signer is invoked.
         *
         * @param defaultRegionScope The default region scope.
         * @return This builder for method chaining.
         */
        Builder defaultRegionScope(RegionScope defaultRegionScope);

        AwsCrtV4aSigner build();
    }
}
```

The main difference is that AWS4Signer utilizes AWS4-HMAC-SHA256, whereas AWSCrtV4asigner uses AWS4-ECDSA-P256-SHA256:

- Algorithm: The primary difference is the algorithm used for signing requests: HMAC with SHA-256 in the former, and ECDSA with P-256 and SHA-256 in the latter.
- Security: Both are considered secure, but ECDSA might offer better performance with equivalent levels of security due to the efficiency of elliptic curve cryptography.
- Implementation: AWS4Signer is from aws-java-sdk V1, whereas AwsCrtV4asigner is from V2. They are incompatible with one another. 

For the long run, we could use AwsCrtV4asigner as the default. However, in this PR, I would like to confine the blast radius to only metadata access using a specified credential provider.

flint-core/src/main/scala/org/opensearch/flint/core/auth/AWSRequestSigningApacheInterceptor.java
```diff
@@ -154,10 +154,11 @@ private static Map<String, String> headerArrayToMap(final Header[] headers) {
...
    * @return true if the given header should be excluded when signing
    */
-  private static boolean skipHeader(final Header header) {
+  static boolean skipHeader(final Header header) {
     return ("content-length".equalsIgnoreCase(header.getName())
-        && "0".equals(header.getValue())) // Strip Content-Length: 0
-        || "host".equalsIgnoreCase(header.getName()); // Host comes from endpoint
+            && "0".equals(header.getValue())) // Strip Content-Length: 0
+            || "host".equalsIgnoreCase(header.getName()) // Host comes from endpoint
+            || "connection".equalsIgnoreCase(header.getName()); // Skip setting Connection manually
```
shikharj05, line 161: nit: better to check if header is against a set of headers

shikharj05 (approved):

flint-core/src/main/scala/org/opensearch/flint/core/auth/ResourceBasedAWSRequestSigningApacheInterceptor.java
penghuo: long term plan is replace signer with signer4a?, and deprecated AWSRequestSigningApacheInterceptor? In case the major difference of singer vs singer 4a is With AWS Signature Version 4A, the signature does not include Region-specific information and is calculated using the AWS4-ECDSA-P256-SHA256 algorithm.
I think we should use one Interceptor instead of two. 

if it is correct, could u create a issue to track future maintain works.

noCharger: > long term plan is replace signer with signer4a?, and deprecated AWSRequestSigningApacheInterceptor? In case the major difference of singer vs singer 4a is With AWS Signature Version 4A, the signature does not include Region-specific information and is calculated using the AWS4-ECDSA-P256-SHA256 algorithm. I think we should use one Interceptor instead of two.
> 
> if it is correct, could u create a issue to track future maintain works.

https://github.com/opensearch-project/opensearch-spark/issues/321

penghuo (approved):

Issue #304, opened 2024-04-04
[FEATURE] Introduce bytes-based batch size config to optimize Flint write throughput 


dai-chen:
**Is your feature request related to a problem?**

The current fixed batch size configuration (`spark.datasource.flint.write.batch_size`) poses limitations, particularly when dealing with documents of varying sizes. This can lead to inefficient resource utilization and potential memory issues. For example, smaller documents may result in underutilization of the batch, while larger documents may cause memory overhead.

**What solution would you like?**

Introduce a new configuration option (ex. `spark.datasource.flint.write.batch_bytes`) to enable bytes-based batch sizing. This adaptive approach would dynamically adjust the batch size based on the overall size of the OpenSearch bulk request body, ensuring optimal resource utilization and minimizing memory overhead.

This proposal is similar to `maxFilesPerTrigger` and `maxBytesPerTrigger` in Spark built-in file source. We're introducing this enhancement specifically for the FlintWrite sink to provide similar adaptability and efficiency in writing data to the OpenSearch cluster.

**What alternatives have you considered?**

Alternative solutions could involve manually adjusting batch sizes based on the expected document sizes, however this approach is:

1. less efficient and prone to errors, especially in dynamic environments where document sizes may vary significantly.
2. requiring changes on the batch size config which is a Spark conf, instead of being configurable per SQL statement

**Do you have any additional context?**

Currently, FlintWrite is instantiated per partition, necessitating the need to optimize throughput globally. A minimal requirement is to avoid over-pressuring the OpenSearch cluster.

penghuo:
## set refresh_policy to false
Flint spark.datasource.flint.write.refresh_policy default value is wait_for. For _bulk, Flint could use false, which return immediately without wait for refresh finished.

PR #329, opened 2024-05-02
add batch_bytes configuration for Flint

penghuo:
### Change the default value
- `spark.datasource.flint.write.batch_size`: 1000 --> Integer.MAX_VALUE.
- `spark.datasource.flint.write.refresh_policy`: wait_for --> false.

#### Add new settings
- `spark.datasource.flint.write.batch_bytes`: The approximately amount of data in bytes written to Flint in a single batch request. The actual data write to OpenSearch may more than it. Default value is 1mb. The writing process checks after each document whether the total number of documents (docCount) has reached batch_size or the buffer size has surpassed batch_bytes. If either condition is met, the current batch is flushed and the document count resets to zero.

### Test Result
With EMR-S 3 Executors (4v CPU 16 GB memory), write to single node OpenSearch cluster, throughput is 40MB/s.

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/304


flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintWriter.java
```diff
@@ -19,4 +19,9 @@ public abstract class FlintWriter extends Writer {
    * { "title": "Prisoners", "year": 2013 }
    */
   public static final String ACTION_CREATE = "create";
+
+  /**
+   * @return current data written into buffer in bytes.
+   */
+  public abstract long getBufferSize();
```
dai-chen, line 26: I found we extend abstract class `Writer` which has an internal buffer array. But actually we never use it as well as any API in Writer?

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchWriter.java
```diff
@@ -27,19 +29,20 @@ public class OpenSearchWriter extends FlintWriter {
...
+  private final ByteArrayOutputStream baos;
 
   private IRestHighLevelClient client;
 
   public OpenSearchWriter(IRestHighLevelClient client, String indexName, String refreshPolicy) {
     this.client = client;
     this.indexName = indexName;
-    this.sb = new StringBuilder();
     this.refreshPolicy = refreshPolicy;
+    this.baos = new ByteArrayOutputStream();
```
dai-chen, line 40: I think we should specify initial size to `batch_bytes` at least? Otherwise, this will be 32 bytes initially and keep growing?

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintWriter.java
penghuo: it is temporary buffer used to hold writes of strings and single characters,  it is not a BufferedWriter.

dai-chen (approved):
Thanks for the changes!

Issue #353, opened 2024-05-21
[BUG] Create Skipping Index does not support multiple part identifier


penghuo:
**What is the bug?**
Create Skipping Index statement does not support multiple part identifier

**How can one reproduce the bug?**
* submit create skipping index
```
CREATE SKIPPING INDEX ON glue.default.vpc_flow ( accountid BLOOM_FILTER, region VALUE_SET, severity_id VALUE_SET, src_endpoint.ip BLOOM_FILTER, dst_endpoint.ip BLOOM_FILTER, src_endpoint.svc_name VALUE_SET, dst_endpoint.svc_name VALUE_SET, request_processing_time MIN_MAX, traffic.bytes MIN_MAX ) WITH ( auto_refresh = true, refresh_interval = '15 Minutes', checkpoint_location = 's3://bucket/', watermark_delay = '1 Minute' )
```

```
[...]

PR #213, opened 2024-01-04
Support struct field as indexed column

dai-chen:
### Description

1. To preserve the nested structure of subfields without flattening them, store the root `StructType` and retrieve fields by examining this structure.
2. When query rewrite, match struct column name with index expression.

#### TODO

To support any index expression in general, we need to analyze the given index expression string when create index or query rewrite. PoC is done here but requires more changes: https://github.com/opensearch-project/opensearch-spark/pull/232. Will delay this until it's required.

#### Testing

I'm unable to check skipping index data in IT due to issue https://github.com/opensearch-project/opensearch-spark/issues/233. Checked this manually by debugging IT:

```
curl "localhost:56262/flint_spark_catalog_default_nested_field_table_skipping_index/_search?pretty"
{
  "took" : 2,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 2,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "flint_spark_catalog_default_nested_field_table_skipping_index",
        "_id" : "a66a1b7b17d545a9248d8a236b2d1c01a358b4a1",
        "_score" : 1.0,
        "_source" : {
          "file_path" : "file:///.../nested_field_table/part-00000-0f065a88-7d52-4e68-a8e3-d39e389221be-c000.json",
          "MinMax_struct_col.field2_0" : 123,
          "MinMax_struct_col.field2_1" : 456,
          "struct_col.field1.subfield" : [
            "value1",
            "value2"
          ]
        }
      },
      {
        "_index" : "flint_spark_catalog_default_nested_field_table_skipping_index",
        "_id" : "1ae12dc8aa4215668c0f4da81f5312ef21987568",
        "_score" : 1.0,
        "_source" : {
          "file_path" : "file:///.../nested_field_table/part-00000-f86ea4eb-19d7-4d60-bf76-61167537a2ad-c000.json",
          "MinMax_struct_col.field2_0" : 789,
          "MinMax_struct_col.field2_1" : 789,
          "struct_col.field1.subfield" : [
            "value3"
          ]
        }
      }
    ]
  }
}
```

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/194

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
[...]

PR #366, opened 2024-06-03
Support nested indexed field in Flint skipping index SQL statement

dai-chen:
### Description

PR https://github.com/opensearch-project/opensearch-spark/pull/213 introduced the support for using nested field as indexed column in Flint skipping index. However, users continue to face issues due to the limitation that the indexed column identifier only supports a single part.

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/353


noCharger (approved):
Thanks for the change! @dai-chen should this patch to 0.4?

dai-chen:
> Thanks for the change! @dai-chen should this patch to 0.4?

Make sense. Because this PR introduces grammar changes, I think we should back port. Thanks!

PR #373, opened 2024-06-10
[Refactor] Introduce flint-commons for models and interfaces

noCharger:
### Description
* [Refactor] Introduce flint-data for models and interfaces
  *  Full details https://github.com/opensearch-project/opensearch-spark/issues/376
  *  This PR will only move FlintCommand and FlintInstance
  *  Other interfaces will be updated continously: For example, SessionManager and CommandLifecycleManager will go next, then FlintMetadataLogEntry, FlintMetadataLogService, etc
* A new jar will be assembled `flint-data-assembly-0.5.0-SNAPSHOT.jar`


ykmr1224 (changes requested):

flint-data/src/main/scala/org/opensearch/flint/data/ContextualData.scala
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.flint.data
+
+trait ContextualData {
```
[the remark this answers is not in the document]
dai-chen: +1 Could you share more context in PR description or Scaladoc for this new module and main interfaces? For now, it seems not very clear why `FlintCommand` + `FlintInstance` = `fint-data`.

flint-data/src/main/scala/org/opensearch/flint/data/ContextualData.scala
```diff
@@ -0,0 +1,18 @@
...
+  var context: Map[String, Any] = Map.empty
```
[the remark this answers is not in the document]
noCharger: The map assignment is required during the flintInstance initialization.

flint-data/src/main/scala/org/opensearch/flint/data/FlintInstance.scala
```diff
@@ -3,34 +3,40 @@
...
+  def isComplete: Boolean = state == "complete"
+  def isFailed: Boolean = state == "failed"
+  def isWaiting: Boolean = state == "waiting"
+
   override def toString: String = {
     val excludedJobIdsStr = excludedJobIds.mkString("[", ", ", "]")
     val errorStr = error.getOrElse("None")
     s"FlintInstance(applicationId=$applicationId, jobId=$jobId, sessionId=$sessionId, state=$state, " +
-      s"lastUpdateTime=$lastUpdateTime, jobStartTime=$jobStartTime, excludedJobIds=$excludedJobIdsStr, error=$errorStr)"
+      s"lastUpdateTime=$lastUpdateTime, jobStartTime=$jobStartTime, excludedJobIds=$excludedJobIdsStr, error=$errorStr, context=$context)"
```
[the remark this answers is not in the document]
noCharger: removed context from toString

flint-data/src/main/scala/org/opensearch/flint/data/ContextualData.scala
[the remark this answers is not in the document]
noCharger: Good point, so the long term goal is to refactor current entry point. Will create an RFC later.

flint-data/src/main/scala/org/opensearch/flint/data/ContextualData.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Adds a key-value pair to the context map.
+   *
+   * @param key
+   *   The key under which the value is stored.
+   * @param value
+   *   The data value to store.
+   */
```
ykmr1224, line 23: Let's try not to add redundant comment. The best state is the code explaining by itself without comment. Reference: https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#comments-rules

flint-data/src/main/scala/org/opensearch/flint/data/ContextualData.scala
```diff
@@ -0,0 +1,39 @@
...
+  def addContext(key: String, value: Any): Unit = {
```
vamsi-amazon, line 24: context is map right...this method is adding a key value pair to the context..`addContext` is giving a wrong impression of adding entire map. Am I getting this correct?

flint-data/src/main/scala/org/opensearch/flint/data/FlintCommand.scala
```diff
@@ -3,52 +3,61 @@
...
 
-package org.opensearch.flint.app
+package org.opensearch.flint.data
 
 import org.json4s.{Formats, NoTypeHints}
 import org.json4s.JsonAST.JString
 import org.json4s.native.JsonMethods.parse
 import org.json4s.native.Serialization
 
+object CommandStates {
```
vamsi-amazon, line 13: Is FlintCommand statement? Just trying to understanding why are we using different nomenclature in sql plugin and spark repo.

flint-data/src/main/scala/org/opensearch/flint/data/FlintCommand.scala
```diff
@@ -3,52 +3,61 @@
...
+ * @param queryId
+ *   Unique identifier for the query.
+ * @param submitTime
+ *   Timestamp when the command was submitted.
+ * @param error
+ *   Optional error message if the command fails.
+ * @param commandContext
+ *   Additional context for the command as key-value pairs.
+ */
 class FlintCommand(
```
vamsi-amazon, line 38: Can we change this to FlintStatement? If it is a big change ignore.

dai-chen:
Just read issue https://github.com/opensearch-project/opensearch-spark/issues/376 for context.
IMHO, `flint-data` is confusing and not sure what we plan to put into this module in future.
Any better name? At the moment, the module is more like Flint session model?

flint-data/src/main/scala/org/opensearch/flint/data/ContextualData.scala
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.flint.data
+
+/**
+ * Provides a mutable map to store and retrieve contextual data using key-value pairs.
+ */
+trait ContextualData {
```
vamsi-amazon, line 11: ContextualData looks like a POJO model. ContextualDataStore or ContextDataStore or ContextValueStore would be better trait names.

build.sbt
```diff
@@ -48,7 +48,7 @@ lazy val commonSettings = Seq(
 
 // running `scalafmtAll` includes all subprojects under root
 lazy val root = (project in file("."))
-  .aggregate(flintCore, flintSparkIntegration, pplSparkIntegration, sparkSqlApplication, integtest)
+  .aggregate(flintData, flintCore, flintSparkIntegration, pplSparkIntegration, sparkSqlApplication, integtest)
```
vamsi-amazon, line 51: What is the reason behind naming flintData?

noCharger:
> Just read issue #376 for context. IMHO, `flint-data` is confusing and not sure what we plan to put into this module in future. Any better name? At the moment, the module is more like Flint session model?

@dai-chen The primary goal is to provide a shared library that allows for easier extension. So maybe `flint-common` or `flint-utils`, any suggestions?

vamsi-amazon:
> > Just read issue #376 for context. IMHO, `flint-data` is confusing and not sure what we plan to put into this module in future. Any better name? At the moment, the module is more like Flint session model?
> 
> > What is the rationale behind naming flintData? It seems to suggest to have all the related model classes, but it does have behavior classes.
> 
> @dai-chen @vamsi-amazon The primary goal is to provide a shared library that allows for easier extension. So maybe `flint-common` or `flint-utils`, any suggestions?

flint-commons can be a good name....but we need to define what the responsibility of the module.

noCharger:
> > > Just read issue #376 for context. IMHO, `flint-data` is confusing and not sure what we plan to put into this module in future. Any better name? At the moment, the module is more like Flint session model?
> > 
> > 
> > > What is the rationale behind naming flintData? It seems to suggest to have all the related model classes, but it does have behavior classes.
> > 
> > 
> > @dai-chen @vamsi-amazon The primary goal is to provide a shared library that allows for easier extension. So maybe `flint-common` or `flint-utils`, any suggestions?
> 
> flint-commons can be a good name....but we need to define what the responsibility of the module.

Thanks, update the responsibility in readme

dai-chen (approved):

Issue #371, opened 2024-06-06
[FEATURE] Index metadata service refactor


seankao-az:
**Is your feature request related to a problem?**
Flint core assumes storage for `FlintMetadataLog` to be OpenSearch. This meta issue aims to refactor code components and data model that are coupled with OpenSearch, and allow plugging in custom implementations using different storage.

**What solution would you like?**
* #377 
  * Refactor OpenSearch `IRestHighLevelClient` creation with static factory method
  * Remove `FlintOpenSearchMetadataLog` dependency on `FlintClient` using the static factory
* #378 
* #379
  * Extract metadata log operations from `FlintClient` into `FlintMetadataLogService`
  * `FlintClient.getIndexMetadata` no longer attaches metadata log entry. Instead, `FlintSpark.describeIndex` does this
  * Abstract `FlintSparkIndexMonitor` record heartbeat action
* #389 
  * Move interfaces to flint-commons module
  * Use reflection for `FlintMetadataLogServiceBuilder`
  * Add parameter for building `FlintMetadataLogService` reading from spark properties
* #406
  * Update `FlintMetadataLogEntry` and `DefaultOptimisticTransaction` to be generic to data storage
  * Remove dataSourceName from `DefaultOptimisticTransaction`
* #495 
* #682 

<img width="1322" alt="Screenshot 2024-09-23 at 4 26 08 PM" src="https://github.com/user-attachments/assets/fe6170eb-3ce3-4264-9cf3-0540a3db1cdd">

**What alternatives have you considered?**
N/A

**Do you have any additional context?**
N/A

PR #377, opened 2024-06-10
Refactor static method for OpenSearch client utils

seankao-az:
### Description
1. Refactor OpenSearch `IRestHighLevelClient` creation with static factory method.
2. Remove `FlintOpenSearchMetadataLog` dependency on `FlintClient` using the static factory.

### Issues Resolved
* #371 


noCharger:
In parent issue https://github.com/opensearch-project/opensearch-spark/issues/371 it mentioned
> Flint core assumes storage for FlintMetadataLog to be OpenSearch. This meta issue aims to refactor code components and data model that are coupled with OpenSearch, and allow plugging in custom implementations using different storage.

1. what problem is this PR attempting to address and how does it connect to the parent issue? 
2. any test to cover the new factory class?

seankao-az:
> what problem is this PR attempting to address and how does it connect to the parent issue?

`FlintOpenSearchMetadataLog` creation will be moved to `FlintOpenSearchMetadataLogService`. Using the static method prevents circular dependency. Otherwise `FlintOpenSearchMetadataLogService` needs `FlintClient` to `createClient`, and `FlintClient` needs `FlintOpenSearchMetadataLogService` to `startTransaction` and `getMetadataLog`

seankao-az:
> consider restructuring the code to directly create instances of IRestHighLevelClient without focusing so heavily on the builder configuration

I think I shouldn't have named it Factory, instead it can be something like `OpenSearchClientUtils`. We don't need the flexibility provided by the Factory Pattern. I don't see an issue why having the createClient method (for the most part) configure the RestClientBuilder is a bad thing? Can you elaborate on this

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchClientFactory.java
```diff
@@ -0,0 +1,114 @@
...
+   */
+  public final static String META_LOG_NAME_PREFIX = ".query_execution_request";
+
+  public static IRestHighLevelClient createClient(FlintOptions options) {
+    RestClientBuilder
+        restClientBuilder =
+        RestClient.builder(new HttpHost(options.getHost(), options.getPort(), options.getScheme()));
+
+    // SigV4 support
+    if (options.getAuth().equals(FlintOptions.SIGV4_AUTH)) {
```
dai-chen, line 46: np: later we can abstract the config logic in case this if-else block becomes more and more complex.

dai-chen (approved):
Thanks for the changes!

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchClientFactory.java
seankao-az: I'm adding another commit. Will make these blocks into methods, should make it slightly better

dai-chen (approved):

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchClientUtils.java
```diff
@@ -0,0 +1,131 @@
...
+ * Utility functions to create {@link IRestHighLevelClient}.
+ */
+public class OpenSearchClientUtils {
+
+  private static final String SERVICE_NAME = "es";
+
+  /**
+   * Metadata log index name prefix
+   */
+  public final static String META_LOG_NAME_PREFIX = ".query_execution_request";
```
noCharger, line 38: this can be refactored with default value assigned 

```
  public String getDataSourceName() {
    return options.getOrDefault(DATA_SOURCE_NAME, "");
  }
```

PR #389, opened 2024-06-20
Support custom metadata log service implementation

seankao-az:
### Description
1. Use reflection for `FlintMetadataLogServiceBuilder` to support custom implementation of `FlintMetadataLogService`
2. Read `SparkConf` for such builder
3. Make `FlintMetadataLogService` an abstract class with a constructor using `SparkConf`
4. Move interfaces to flint-commons
    1. Moved files: `FlintMetadataLog`, `FlintMetadataLogEntry`, `FlintMetadataLogService`, `OptimisticTransaction`
    2. Remove unused `failLogEntry` from `FlintMetadataLogEntry` (no access to org.opensearch.index namespace). Can add back in next PR when storage is abstracted away from the common `FlintMetadataLogEntry`

### Issues Resolved
* #371 

### DIFF
--- a/flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.flint.core.metadata.log;
 
+import java.lang.reflect.Constructor;
+import org.apache.spark.SparkConf;
 import org.opensearch.flint.core.FlintOptions;
 import org.opensearch.flint.core.storage.FlintOpenSearchMetadataLogService;
 
@@ -12,7 +14,19 @@ import org.opensearch.flint.core.storage.FlintOpenSearchMetadataLogService;
  * {@link FlintMetadataLogService} builder.
  */
 public class FlintMetadataLogServiceBuilder {
-  public static FlintMetadataLogService build(FlintOptions options) {
-    return new FlintOpenSearchMetadataLogService(options);
+  public static FlintMetadataLogService build(FlintOptions options, SparkConf sparkConf) {
+    String className = options.getCustomFlintMetadataLogServiceClass();
+    if (className.isEmpty()) {
+      return new FlintOpenSearchMetadataLogService(options);
+    }
+
+    // Attempts to instantiate Flint metadata log service with sparkConf using reflection
+    try {
+      Class<?> flintMetadataLogServiceClass = Class.forName(className);
+      Constructor<?> constructor = flintMetadataLogServiceClass.getConstructor(SparkConf.class);
+      return (FlintMetadataLogService) constructor.newInstance(sparkConf);
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to instantiate FlintMetadataLogService: " + className, e);
+    }
   }
 }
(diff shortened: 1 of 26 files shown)


flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
```diff
@@ -5,14 +5,29 @@
...
+  public static FlintMetadataLogService build(FlintOptions options, SparkConf sparkConf) {
+    String className = options.getCustomFlintMetadataLogServiceClass();
+    if (className.isEmpty()) {
+      return new FlintOpenSearchMetadataLogService(options);
+    }
+
+    // Attempts to instantiate Flint metadata log service with sparkConf using reflection
+    try {
+      Class<?> flintMetadataLogServiceClass = Class.forName(className);
+      Constructor<?> constructor = flintMetadataLogServiceClass.getConstructor(SparkConf.class);
```
penghuo, line 27: it required FlintMetadataLogService constructor has SparkConf as parameter. could we add doc on FlintMetadataLogService to clarify on it?

seankao-az: good point. Enforcing the constructor by making it a abstract class instead. what do you think?

penghuo: it is ok for now, add doc should be fine.

penghuo (approved):
thx!

flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
seankao-az: just realized there really is no way to enforce implementation class to have a specific signature for constructor (my attempt doesn't work). Reverting it and add doc instead.

flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
```diff
@@ -5,14 +5,33 @@
...
 import org.opensearch.flint.core.storage.FlintOpenSearchMetadataLogService;
 
 /**
  * {@link FlintMetadataLogService} builder.
+ * <p>
+ * Custom implementations of {@link FlintMetadataLogService} are expected to provide a public
+ * constructor with the signature {@code public MyCustomService(SparkConf sparkConf)} to be
+ * instantiated by this builder.
  */
 public class FlintMetadataLogServiceBuilder {
```
noCharger, line 21: should FlintMetadataLogServiceBuilder moved to common as well?

flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
noCharger: Is this change reverted? it seems still use `SparkConf `

flint-core/src/main/scala/org/opensearch/flint/core/metadata/log/FlintMetadataLogServiceBuilder.java
seankao-az: no this needs access to `FlintOpenSearchMetadataLogService`

PR #406, opened 2024-06-30
Separate metadata log entry data model and persistence

seankao-az:
### Description
* Refactor `FlintMetadataLogEntry` to be generic to data storage
* Remove `dataSourceName` from `DefaultOptimisticTransaction`
* Move persistence logic for OpenSearch away from `FlintMetadataLogEntry` and `DefaultOptimisticTransaction` into `FlintMetadataLogEntryOpenSearchConverter`

#### Before
```
FlintMetadataLogEntry(
    id: String,
    seqNo: Long,
    primaryTerm: Long,
    createTime: Long,
    state: IndexState,
    dataSource: String,
    error: String)
```

#### After
```
FlintMetadataLogEntry(
    id: String,
    createTime: Long,
    state: IndexState,
    entryVersion: Map[String, Any],
    error: String,
    storageContext: Map[String, Any])
```
* `entryVersion` is used for `FlintSpark` `updateIndex`. For OpenSearch, it should specify `seqNo` and `primaryTerm`
* `storageContext` can be used to store additional context required for persistence. For example, `dataSourceName` for OpenSearch

### Issues Resolved
* #371 

### DIFF
--- a/flint-commons/src/main/scala/org/opensearch/flint/common/metadata/log/FlintMetadataLogEntry.scala
+++ b/flint-commons/src/main/scala/org/opensearch/flint/common/metadata/log/FlintMetadataLogEntry.scala
@@ -22,6 +22,8 @@ import org.opensearch.flint.common.metadata.log.FlintMetadataLogEntry.IndexState
  *   entry version fields for consistency control
  * @param error
  *   error details if in error state
+ * @param storageContext
+ *   extra context fields required for storage
  */
 case class FlintMetadataLogEntry(
     id: String,
@@ -31,16 +33,28 @@ case class FlintMetadataLogEntry(
      */
     createTime: Long,
     state: IndexState,
-    entryVersion: Map[String, _],
-    error: String) {
+    entryVersion: Map[String, Any],
+    error: String,
+    storageContext: Map[String, Any]) {
 
   def this(
       id: String,
       createTime: Long,
       state: IndexState,
-      entryVersion: JMap[String, _],
-      error: String) {
-    this(id, createTime, state, entryVersion.asScala.toMap, error)
+      entryVersion: JMap[String, Any],
+      error: String,
+      storageContext: JMap[String, Any]) = {
+    this(id, createTime, state, entryVersion.asScala.toMap, error, storageContext.asScala.toMap)
+  }
+
+  def this(
+      id: String,
+      createTime: Long,
+      state: IndexState,
+      entryVersion: JMap[String, Any],
+      error: String,
+      storageContext: Map[String, Any]) = {
+    this(id, createTime, state, entryVersion.asScala.toMap, error, storageContext)
   }
 }
 
--- a/flint-commons/src/main/scala/org/opensearch/flint/common/metadata/log/FlintMetadataLogEntry.scala
+++ b/flint-commons/src/main/scala/org/opensearch/flint/common/metadata/log/FlintMetadataLogEntry.scala
@@ -5,51 +5,71 @@
 
 package org.opensearch.flint.common.metadata.log
 
+import java.util.{Map => JMap}
+import java.util.Base64
+
+import scala.collection.JavaConverters._
+
 import org.opensearch.flint.common.metadata.log.FlintMetadataLogEntry.IndexState
 import org.opensearch.flint.common.metadata.log.FlintMetadataLogEntry.IndexState.IndexState
 
 /**
- * Flint metadata log entry. This is temporary and will merge field in FlintMetadata here and move
- * implementation specific field, such as seqNo, primaryTerm, dataSource to properties.
+ * Flint metadata log entry. This is temporary and will merge field in FlintMetadata here.
  *
  * @param id
  *   log entry id
- * @param seqNo
- *   OpenSearch sequence number
- * @param primaryTerm
- *   OpenSearch primary term
+ * @param indexName
+ *   Flint index name
+ * @param dataSource
+ *   data source associated //TODO: remove?
+ * @param createTime
+ *   streaming job start time
  * @param state
  *   Flint index state
- * @param dataSource
- *   OpenSearch data source associated //TODO: remove?
+ * @param entryVersion
+ *   entry version fields for consistency control
  * @param error
  *   error details if in error state
  */
 case class FlintMetadataLogEntry(
     id: String,
-    seqNo: Long,
-    primaryTerm: Long,
+    indexName: String,
+    dataSource: String,
     /**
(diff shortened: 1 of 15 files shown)


flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchMetadataLog.java
```diff
@@ -113,26 +121,46 @@ public void purge() {
...
+    return new FlintMetadataLogEntry(
+        "",
+        0L,
+        FlintMetadataLogEntry.IndexState$.MODULE$.EMPTY(),
+        Map.of("seqNo", UNASSIGNED_SEQ_NO, "primaryTerm", UNASSIGNED_PRIMARY_TERM),
+        "",
+        Map.of("dataSourceName", dataSourceName));
+  }
+
+  public FlintMetadataLogEntry failLogEntry(String error) {
```
seankao-az, line 135: This really isn't used anywhere. It's there before this series of refactoring. Should I remove?

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintMetadataLogEntryOpenSearchConverter.java
```diff
@@ -0,0 +1,149 @@
...
+   * Convert a log entry to json string for persisting to OpenSearch.
+   * Expects the following field in storage context:
+   * - dataSourceName: data source name
+   *
+   * @param logEntry
+   *   log entry to convert
+   * @return
+   *   json string representation of the log entry
+   */
+  public static String toJson(FlintMetadataLogEntry logEntry) {
```
dai-chen, line 93: Can we take this opportunity to switch to some Json library to do this instead of `String.format`? I recall I met issue earlier that the latter won't take care of escaping special character.

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintMetadataLogEntryOpenSearchConverter.java
```diff
@@ -0,0 +1,149 @@
...
+            "  \"dataSourceName\": \"%s\",\n" +
+            "  \"jobStartTime\": %d,\n" +
+            "  \"lastUpdateTime\": %d,\n" +
+            "  \"error\": \"%s\"\n" +
+            "}",
+        logEntry.id(),
+        logEntry.state(),
+        applicationId,
+        jobId,
+        logEntry.storageContext().get("dataSourceName").get(),
```
dai-chen, line 115: better to serialize the entire Map context no matter what's in it?

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintMetadataLogEntryOpenSearchConverter.java
```diff
@@ -0,0 +1,149 @@
...
+package org.opensearch.flint.core.storage;
+
+import org.opensearch.flint.common.metadata.log.FlintMetadataLogEntry;
+
+import java.util.Map;
+
+/**
+ * Utility class for handling Flint metadata log entries in OpenSearch storage.
+ */
+public class FlintMetadataLogEntryOpenSearchConverter {
```
dai-chen, line 15: Should we have an interface for this? As I understand, this abstraction is responsible for serializing/deserializing `FlintMetadataLogEntry` for different storage and this class is for OpenSearch?

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchMetadataLog.java
seankao-az: no concerns. I'll remove it then

flint-commons/src/main/scala/org/opensearch/flint/common/metadata/log/FlintMetadataLogEntry.scala
```diff
@@ -5,67 +5,56 @@
...
  * @param state
  *   Flint index state
- * @param dataSource
- *   OpenSearch data source associated //TODO: remove?
+ * @param entryVersion
+ *   entry version fields for consistency control
  * @param error
  *   error details if in error state
+ * @param storageContext
+ *   extra context fields required for storage
```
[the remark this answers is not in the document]
seankao-az: sure. making it `properties`

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintMetadataLogEntryOpenSearchConverter.java
dai-chen: @seankao-az I found my PR for storing error message in index state log requires this. Let me know if you plan to address this or not. Thanks!

seankao-az: yes will switch to Json library

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintMetadataLogEntryOpenSearchConverter.java
seankao-az: Doing this means that we need to change the fields for `flintindexstate` doc and bump its version from 1.0 to 2.0, assuming that we replace the `dataSourceName` field with a new `properties` field. I think this is not necessary.

The flexibility for the `FlintMetadataLogEntry` `properties` field is to accommodate custom property requirement for different storages. For OpenSearch, it's just `dataSourceName`. Don't think the whole Map needs to be serialized.

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintMetadataLogEntryOpenSearchConverter.java
seankao-az: This might not be the best approach, since the serialization and deserialization requirements, as well as the storage client interactions, can vary significantly across different storage systems.

I think the abstraction mainly happens at `FlintMetadataLog` level, where the `add` and `getLatest` should define how each storage system would like to store `FlintMetadataLogEntry`. And this OpenSearchConverter is merely an implementation detail for `FlintOpenSearchMetadataLog`.

Perhaps it makes sense to make this class into a inner class of `FlintOpenSearchMetadataLog`? Or make these into private member / method of it?

dai-chen (approved):
Thanks for the changes!

PR #407, opened 2024-07-02
PPL Lookup

salyh:
### Description
Implement PPL Lookup

### Issues Resolved
https://github.com/opensearch-project/sql/issues/2651

### DIFF
--- a/ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
+++ b/ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
@@ -6,13 +6,18 @@
 package org.opensearch.sql.ppl;
 
 import org.apache.spark.sql.catalyst.analysis.UnresolvedAttribute$;
+import org.apache.spark.sql.catalyst.analysis.UnresolvedFieldName;
 import org.apache.spark.sql.catalyst.analysis.UnresolvedRelation;
 import org.apache.spark.sql.catalyst.analysis.UnresolvedStar$;
+import org.apache.spark.sql.catalyst.expressions.EqualTo;
 import org.apache.spark.sql.catalyst.expressions.Expression;
 import org.apache.spark.sql.catalyst.expressions.NamedExpression;
 import org.apache.spark.sql.catalyst.expressions.Predicate;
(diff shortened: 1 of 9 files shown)


YANG-DB:
@salyh please take a look at this existing [PPL correlation API ](https://github.com/opensearch-project/opensearch-spark/blob/main/docs/PPL-Correlation-command.md)

salyh:
@YANG-DB any comments on the proposal so far?

salyh:
@rupal-bq @anasalkouz any comments on the proposal so far?

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -256,6 +267,93 @@ public LogicalPlan visitDedupe(Dedupe node, CatalystPlanContext context) {
...
+
+        while(!root.getChild().isEmpty()) {
+            root = root.getChild().get(0);
+        }
+
+        org.opensearch.sql.ast.tree.Relation source = (org.opensearch.sql.ast.tree.Relation) root;
+
+        node.getChild().get(0).accept(this, context);
+
+        //TODO: not sure how to implement appendonly
```
LantaoJin, line 282: To implement `appendonly`, the output Project (parent of Join) should be
```
Project [other fields in source, CASE WHEN isnull('source.copyField) THEN coalesce('source.copyField, 'lookup.copyField) AS copyField ELSE 'lookup.copyField END AS copyField]
  +- Join LeftOuter
```
In `CASE WHEN expression1 ELSE expression2`, the expression1 is for appendonly (not performing when the source field already exist), the else is for overwriting existing field.

kt-eliatra: Doesn't it mean that we need to know list of fields available in the "source" table, in order to build project consisting of two lists:
a list of fields which we want to overwrite with values from the lookup table and a list of fields which should remain untouched?
I'm not sure whether it is possible to retrieve such list of available fields.

LantaoJin:
Close this as issue resolved by https://github.com/opensearch-project/opensearch-spark/pull/686

PR #410, opened 2024-07-03
handle MetaException with glue AccessDeniedException

noCharger:
### Description

handleGeneralException should handle MetaException with AccessDeniedException properly

```
24/07/03 21:34:42 INFO FlintREPL: Original exception class name: org.apache.spark.SparkException
24/07/03 21:34:42 INFO FlintREPL: Root cause exception class name: org.apache.hadoop.hive.metastore.api.MetaException
```
Screenshot 2024-07-03 at 15 43 16

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
@@ -13,7 +13,7 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import org.apache.commons.text.StringEscapeUtils.unescapeJava
 import org.opensearch.flint.core.IRestHighLevelClient
-import org.opensearch.flint.core.logging.{CustomLogging, OperationMessage}
+import org.opensearch.flint.core.logging.{CustomLogging, ExceptionMessages, OperationMessage}
 import org.opensearch.flint.core.metrics.MetricConstants
 import org.opensearch.flint.core.metrics.MetricsUtil.incrementCounter
 import play.api.libs.json._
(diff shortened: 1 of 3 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
```diff
@@ -454,34 +454,47 @@ trait FlintJobExecutor {
...
+    }
+  }
+
+  private def handleGeneralException(ex: Exception): String = {
+    val rootCauseClassName = ex.getClass.getName
+    val errMsg = ex.getMessage
+    logDebug(s"Root cause class name: $rootCauseClassName")
+    logDebug(s"Root cause error message: $errMsg")
+    if (rootCauseClassName == "org.apache.hadoop.hive.metastore.api.MetaException" &&
+      errMsg.contains("com.amazonaws.services.glue.model.AccessDeniedException")) {
```
noCharger, line 493: Use class name string here since `hadoop.hive.metastore` is not a dependency for `spark-hive`

vamsi-amazon:
LGTM. but check the build failure.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
dai-chen, line 493: Is it more generic to match error message only? Also if this method is `handleGeneralException`, should we move this to match case outside?

noCharger: > Is it more generic to match error message only? Also if this method is `handleGeneralException`, should we move this to match case outside?

If the exception itself is `com.amazonaws.services.glue.model.AccessDeniedException` it may not included in the error message.

noCharger:
> LGTM. but check the build failure.

Fixed. Failed due to flint-commons refactor, will need a manual backport

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
dai-chen: I think probably in future we should maintain a "white list". For other exceptions, a general message is returned as before. We can discuss if we see more other cases. Thanks!

vamsi-amazon (approved):

Issue #397, opened 2024-06-24
[FEATURE] Support configurable query loop frequency


noCharger:
**Is your feature request related to a problem?**

Currently, the query loop sleeps for a fixed 100ms for each batch execution.
https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala#L395

**What solution would you like?**

Add a new spark configuration to support the customizable query loop frequency.

dblock:
[Catch All Triage - [1](https://github.com/dblock), [2](https://github.com/reta)]

PR #411, opened 2024-07-04
Add config on query loop execution frequency

noCharger:
### Description

With default query loop execution frequency there are 3 min session idle time and 30000 / 100 = 300 additional calls to check queries waiting in the loop.

For example, on sanity IT, there are 5000 / 100 = 50 calls theoretically, and in actuality 47 calls due to the actual query execution time.

```
FlintREPL: read from flint_ql_sessions, sessionId: 10
```

On new IT "query loop should exit with inactivity timeout due to large query loop frequency," there should be no subsequent calls because the loop waits 5 seconds, which is deemed inactivity. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/397

### DIFF
--- a/integ-test/src/integration/scala/org/apache/spark/sql/FlintREPLITSuite.scala
+++ b/integ-test/src/integration/scala/org/apache/spark/sql/FlintREPLITSuite.scala
@@ -18,9 +18,10 @@ import org.opensearch.flint.OpenSearchSuite
 import org.opensearch.flint.core.{FlintClient, FlintOptions}
 import org.opensearch.flint.core.storage.{FlintOpenSearchClient, FlintReader, OpenSearchUpdater}
 import org.opensearch.flint.data.{FlintStatement, InteractiveSession}
-import org.scalatest.prop.TableDrivenPropertyChecks._
+import org.opensearch.search.sort.SortOrder
 
 import org.apache.spark.SparkFunSuite
+import org.apache.spark.sql.FlintREPLConfConstants.DEFAULT_QUERY_LOOP_EXECUTION_FREQUENCY
 import org.apache.spark.sql.flint.config.FlintSparkConf.{DATA_SOURCE_NAME, EXCLUDE_JOB_IDS, HOST_ENDPOINT, HOST_PORT, JOB_TYPE, REFRESH_POLICY, REPL_INACTIVITY_TIMEOUT_MILLIS, REQUEST_INDEX, SESSION_ID}
 import org.apache.spark.sql.util.MockEnvironment
 import org.apache.spark.util.ThreadUtils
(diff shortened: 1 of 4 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
```diff
@@ -134,7 +135,10 @@ object FlintREPL extends Logging with FlintJobExecutor {
         SECONDS)
       val queryWaitTimeoutMillis: Long =
         conf.getLong("spark.flint.job.queryWaitTimeoutMillis", DEFAULT_QUERY_WAIT_TIMEOUT_MILLIS)
-
+      val queryLoopExecutionFrequency: Long =
+        conf.getLong(
+          "spark.flint.job.queryLoopExecutionFrequency",
```
dai-chen, line 140: Update user manual?

integ-test/src/integration/scala/org/apache/spark/sql/FlintREPLITSuite.scala
```diff
@@ -212,220 +217,232 @@ class FlintREPLITSuite extends SparkFunSuite with OpenSearchSuite with JobTest {
...
-
-      // shutdown hook is called after all tests have finished. We cannot verify if session has correctly been set in IT.
     }
   }
 
   test("create table with dummy location should fail with excepted error message") {
     try {
       createSession(jobRunId, "")
-      threadLocalFuture.set(startREPL())
+      threadLocalFuture.set(startREPL(100L))
```
dai-chen, line 445: use `DEFAULT_QUERY_LOOP_EXECUTION_FREQUENCY` instead?

integ-test/src/integration/scala/org/apache/spark/sql/FlintREPLITSuite.scala
```diff
@@ -212,220 +217,232 @@ class FlintREPLITSuite extends SparkFunSuite with OpenSearchSuite with JobTest {
...
-          result.schemas.size == 3,
-          s"expected schema size is 3, but got ${result.schemas.size}")
-        val expectedZerothSchema = "{'column_name':'namespace','data_type':'string'}"
+  // Define the test parameters
+  val testParams = Table(
+    ("testName", "queryLoopExecutionFrequency"),
+    ("Sanity with 100ms frequency", 100),
+    ("Sanity with 1000ms frequency", 1000))
+
+  forAll(testParams) { (testName: String, queryLoopExecutionFrequency: Int) =>
```
dai-chen, line 226: np: maybe do this in a separate test that focus on the loop frequency.

dai-chen (commented):
Please check CI failure on REPL suite.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
noCharger: Currently no manual around REPL config, so I created an issue to track this https://github.com/opensearch-project/opensearch-spark/issues/526

integ-test/src/integration/scala/org/apache/spark/sql/FlintREPLITSuite.scala
noCharger: DEFAULT_QUERY_LOOP_EXECUTION_FREQUENCY is private to FlintREPL

dai-chen: I found some constants are private and some are public in `FlintREPL`. Maybe they can be public and in future move to some REPL config class?

noCharger: Ideally, separate the repl conf from FlintSparkConf, but for now, use the same.

dai-chen (approved):
Thanks for the changes!

PR #473, opened 2024-07-24
Fix SigV4 signature when connecting to OpenSearchServerless

ykmr1224:
### Description
- Fix SigV4 signature when connecting to OpenSearch Serverless (AOSS)
- AOSS requires service name to be "aoss", and additional header for the signature of the body content.

### Issues Resolved
- It was not able to read/write AOSS index.


flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
```diff
@@ -33,16 +33,15 @@ public class FlintOptions implements Serializable {
 
   public static final String SCHEME = "scheme";
 
-  public static final String AUTH = "auth";
+  public static final String INDEX_TYPE = "index.type";
```
penghuo, line 36: SERVICER_NAME?

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
```diff
@@ -33,16 +33,15 @@ public class FlintOptions implements Serializable {
...
+  public static final String INDEX_TYPE_AOS = "aos";
```
penghuo, line 37: what is aos, aoss? add comments,  and update doc.

penghuo: we can use es and aoss directly?

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchClientUtils.java
```diff
@@ -30,13 +34,19 @@
...
 public class OpenSearchClientUtils {
 
-  private static final String SERVICE_NAME = "es";
-
   /**
    * Metadata log index name prefix
    */
   public final static String META_LOG_NAME_PREFIX = ".query_execution_request";
 
+  public static final String AOS_SIGV4_SERVICE_NAME = "es";
```
penghuo, line 42: not necessary? we can let user configu in FlintOptions

penghuo (commented):
1. Add a integration test with serverless. 
2. In local dev, verify `sbt integtest/integration` success.
more reading. https://github.com/opensearch-project/opensearch-spark/blob/main/DEVELOPER_GUIDE.md#integration-test

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
ykmr1224: I felt directly using `es` is confusing. Let me reconsider the option name and directly use `es` and `aoss`.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchClientUtils.java
ykmr1224: will address.

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
dai-chen: maybe use "index store"? Ref: https://github.com/opensearch-project/opensearch-spark/blob/main/docs/index.md#index-store. If more difference between OS and AOSS in future, we should consider subclass instead of adding more conditional check?

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
ykmr1224: fixed

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
ykmr1224: Fixed to directly use `es` and `aoss`.

>If more difference between OS and AOSS in future, we should consider subclass instead of adding more conditional check?
Agree. So far, the difference is limited and I think we don't need subclass.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchClientUtils.java
ykmr1224: Fixed.

ykmr1224:
> 1. Add a integration test with serverless.
> 2. In local dev, verify `sbt integtest/integration` success.
>    more reading. https://github.com/opensearch-project/opensearch-spark/blob/main/DEVELOPER_GUIDE.md#integration-test

1 -> Let me add integration test in the next PR for fixing the refresh policy, since it fails without that change.
2 -> Verified it succeeds.

penghuo (approved):
Thx!

PR #495, opened 2024-07-29
Abstract service for accessing Flint index metadata

seankao-az:
### Description
* add `FlintIndexMetadataService`
  * move `FlintMetadata` and `FlintVersion` to flint-commons
  * remove storage-specific operation from `FlintMetadata`
    * remove OpenSearch json (de)serialization from `FlintMetadata`
    * move to OpenSearch implementation for `FlintIndexMetadataService`
  * remove schema parser in `FlintMetadata` builder to remove dependency to flint-core
    * `FlintSparkIndex` generate schema map from json
* operations
  * move `get(All)IndexMetadata` from `FlintClient` to `FlintIndexMetadataService`
  * remove `updateIndex` from `FlintClient`; add `updateIndexMetadata` to `FlintIndexMetadataService`
  * add empty implementation for OpenSearch `deleteIndexMetadata`, as _meta will get deleted alongside `FlintOpenSearchCilent.deleteIndex`
* add builder for `FlintIndexMetadataService` and options
  * fix `FlintOptions` to read spark properties for custom service class
  * remove `SparkConf` from service builder argument
* misc
  * rename for log entry properties
  * correct typo planTransformer for ppl
  * move OpenSearch index name sanitization to utils class
  * fix OpenSearch index name not sanitized for metadata log bug
  * remove OpenSearchIndexTable's dependency on Flint client/services

### Issues Resolved
* #371 
* #555 


seankao-az:
Scrapping this design. Further creating yet another class for metadata isn't ideal.

flint-core/src/main/scala/org/opensearch/flint/core/table/OpenSearchIndexTable.scala
```diff
@@ -160,8 +161,8 @@ object OpenSearchCluster {
    *   An list of OpenSearchIndexTable instance.
    */
   def apply(indexName: String, options: FlintOptions): Seq[OpenSearchIndexTable] = {
-    val client = FlintClientBuilder.build(options)
-    client
+    val indexMetadataService = FlintIndexMetadataServiceBuilder.build(options)
```
penghuo, line 164: OpenSearchIndexTable should not depend on FlintIndexMetadataServiceBuilder. If the reasons is getAllIndexMetadata is moved away from FlintClient. can we add in FlintClient instead.
```
def getAllIndexMetadata(): IndexMetadata
```

seankao-az: any reason why that's the case? Reason why I move `getAllIndexMetadata` away from FlintClient is because we might use `FlintOpenSearchClient` while the index metadata is stored with some other custom storage. In this case we would like to fetch index metadata with `FlintIndexMetadataService`, not with `FlintClient`

penghuo: FlintIndexMetadataService can be backed by OpenSearch or DDB? but OpenSearchIndexTable is refering to OpenSearch cluster.

seankao-az: got it. This is for regular OpenSearch index, not about Flint indexes (skipping, covering, mv)

seankao-az: I see. So this is dealing with all OpenSearch indexes including regular, non-Flint indexes. It doesn't use the `_meta` for Flint indexes, and it's only using the mappings and settings part of `FlintMetadata`. Then those won't be covered by `FlintIndexMetadataService` then.

However methods in `FlintClient` are designed for Flint indexes, not regular OS index. I think mixing use case for Flint index and regular OS index in `FlintClient` would be confusing. Also, having both `FlintClient.getAllIndexMetadata` and `FlintIndexMetadataService.getAllIndexMetadata` is confusing.

I'll see how to do it.. I don't think OpenSearchIndexTable should use FlintClient or FlintIndexMetadataService. It should either use a private method for that or I move the method into some util class

seankao-az: ended up getting rid of `FlintClient`, `FlintIndexMetadataService`, and `FlintMetadata` from `OpenSearchCluster`
It has its own method for getting `MetaData` for regular non-Flint OpenSearch indices.

Had some trouble with mocking it in `ApplyFlintSparkCoveringIndexSuite` so I made it into java class (rather than scala object) which can be easily mocked.

seankao-az:
CI passed. just added one commit for documentation

flint-core/src/main/scala/org/opensearch/flint/core/table/OpenSearchCluster.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * Retrieve all metadata for OpenSearch table whose name matches the given pattern.
+   *
+   * @param options The options for Flint.
+   * @param indexNamePattern index name pattern
+   * @return list of OpenSearch table metadata
+   */
+  public static List<MetaData> getAllOpenSearchTableMetadata(FlintOptions options, String... indexNamePattern) {
```
penghuo, line 49: why not keep getOpenSearchTableMetadata method in FlintClient?

seankao-az: It's not really an operation for Flint indexes. In the context of FlintClient, it's for operations on Flint index and shouldn't be bounded to specific storage.
>  Flint index client that provides API for metadata and data operations on a Flint index regardless of concrete storage.

Having `flintClient.getAllOpenSearchTableMetadata` will violate that.

Do you see a better way to fit this method into FlintClient?

penghuo (approved):

flint-core/src/main/scala/org/opensearch/flint/core/table/OpenSearchCluster.java
seankao-az: Alternative is to add yet another client OpenSearchClient (non-Flint version) but that's overkill I think

PR #530, opened 2024-08-06
Use refresh policy from config

ykmr1224:
### Description
- Use refresh policy from config when calling OpenSearch APIs
- This will allow Flint to connect to OpenSearch Serverless collection. It originally use WAIT_FOR refresh policy for some requests and it failed against Serverless collection since it only accept NONE refresh policy.
- To use Serverless collection as index store, `spark.datasource.flint.write.refresh_policy=false` config needs to be specified.
- Added integration test for Serverless

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._ 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #554, opened 2024-08-09
Fix refresh policy back to WAIT_FOR other than writing query result

ykmr1224:
### Description
- Fix refresh policy back to WAIT_FOR other than writing query result
- Those are fixed in #530 , but should not be changed since it could break due to eventual consistency. In actual use case, AOSS doesn't need to write to other than result and index(MV/CV/SK)

### Issues Resolved


dai-chen (approved):
Just to confirm: my understanding is we need to make sure refresh policy is set to `WAIT_FOR` in Spark job before we deploy this in Flint 0.5, right? Otherwise, the refresh policy of writing to result index becomes `NONE`?

ykmr1224:
> Just to confirm: my understanding is we need to make sure refresh policy is set to `WAIT_FOR` in Spark job before we deploy this in Flint 0.5, right? Otherwise, the refresh policy of writing to result index becomes `NONE`?

That's right.

penghuo (approved):

noCharger:
QQ: why it takes longer after this change? Previously it was around 40-45 mins and now it's around 1 hour

noCharger (commented):
QQ: why it takes longer after this change? Previously it was around 40-45 mins and now it's around 1 hour

Issue #356, opened 2024-05-23
[FEATURE] Enhance handling of Flint index data deletion to prevent dangling metadata log entry


dai-chen:
**Is your feature request related to a problem?**

Currently, users can directly delete an OpenSearch index without using `DROP` and `VACUUM` index statement. This action can lead to issues where the index cannot be recreated because the metadata log entry persists. This scenario occurs frequently when indices are created with auto-refresh enabled, which can lead to errors.

Although the recent improvements in pre-validation (as seen in [PR#297](https://github.com/opensearch-project/opensearch-spark/pull/297)) have reduced these incidents, it is still possible for an OpenSearch index to be left dangling after a forceful deletion.

**What solution would you like?**

Enhances the handling of direct Flint index data deletions, particularly one that can intelligently manage or clean up metadata logs to avoid leaving dangling indices. This could possibly involve additional checks or cleanup processes when an index is deleted.

**What alternatives have you considered?**

One alternative is to improve user education regarding the management of indices. Users could be encouraged to use Flint SQL for index manipulations and avoid direct interactions with the OpenSearch index. However, this approach relies heavily on user compliance.

**Do you have any additional context?**

We've added such cleanup logic in `recoverIndex` API in https://github.com/opensearch-project/opensearch-spark/pull/241.

PR #241, opened 2024-02-01
Fix recover index bug when Flint data index is deleted accidentally

dai-chen:
### Description

To address the 2 issues related:

1. Quick fixed recover index API to clean up metadata log entry if Flint data index is gone. This prevents index stuck in refreshing state and infinite attempt on recover index API.
2. Added check to prevent FlintJob hang even though no streaming job launched by recover statement.

#### Documentation

Updated user manual: https://github.com/dai-chen/opensearch-spark/blob/fix-recover-index-for-index-data-deleted/docs/index.md#index-job-management

#### TODO

1. Integration test with FlintJob *[Checking with @kaituo if possible to add IT]*
3. https://github.com/opensearch-project/opensearch-spark/issues/244

### Testing

Manual test to double confirm. First of all, replicate the problematic scenario as outlined below:

<pre>
CREATE SKIPPING INDEX ON stream.lineitem_tiny
(l_shipdate VALUE_SET)
WITH ( auto_refresh = true );

# Delete Flint data index
<b>DELETE flint_myglue_stream_lineitem_tiny_skipping_index</b>

# Check index state metadata log
GET .query_execution_request_myglue/_search
      {
        "_index": ".query_execution_request_myglue",
        "_id": "ZmxpbnRfbXlnbHVlX3N0cmVhbV9saW5laXRlbV90aW55X3NraXBwaW5nX2luZGV4",
        "_score": 1,
        "_source": {
          "version": "1.0",
          "latestId": "ZmxpbnRfbXlnbHVlX3N0cmVhbV9saW5laXRlbV90aW55X3NraXBwaW5nX2luZGV4",
          "type": "flintindexstate",
          <b>"state": "refreshing",</b>
          "applicationId": "unknown",
          "jobId": "unknown",
          "dataSourceName": "myglue",
          "jobStartTime": 1706916312007,
          "lastUpdateTime": 1706916452474,
          "error": ""
        }
      }
</pre>

Now verify the enhanced recover index API:

<pre>
spark-sql> RECOVER INDEX JOB flint_myglue_stream_lineitem_tiny_skipping_index;
<b>24/02/02 23:48:31 WARN FlintSpark: Cleaning up metadata log as index data has been deleted</b>

# The metadata log is gone
GET .query_execution_request_myglue/_search
</pre>

### Issues Resolved

- https://github.com/opensearch-project/opensearch-spark/issues/227
- https://github.com/opensearch-project/opensearch-spark/issues/237

[...]

PR #558, opened 2024-08-12
Lazy clean up dangling index metadata log entry

dai-chen:
### Description

This PR introduces a lazy approach to handling potentially corrupted indices. Previously only `recoverIndex` API has such capability introduced in https://github.com/opensearch-project/opensearch-spark/pull/241 and this PR makes this logic generic across all Flint APIs.

A corrupted index is defined as a Flint index **has an index log entry but no corresponding data index**. Upon the next Flint API call, the system will check for such corrupted indices and perform the necessary cleanup, ensuring that users can recreate indices without encountering errors.

#### Typical Scenario

The most common scenario this PR aims to address is:

1. Users delete OS data index for a Flint index directly
2. Index monitor terminates streaming job if auto refresh enabled
3. Users attempt to create index again
    a. **Before changes**: the creation failed and require to remove the log entry manually
    b. **After changes**: the log entry is cleaned up automatically and creation succeeds

#### Uncommon Scenarios and Handling

After step 1 above, user may rarely attempt to:

1. **Create or vacuum index again concurrently**: Skip corruption check if index in `CREATING` or `VACUUMIGN` state to reduce the possibility of race condition, ensuring log entry won't be removed mistakenly during ongoing operation.
2. **Operate before index monitor terminates streaming job**: If index corrupted, skip heartbeat reporting in index monitor which relies on log entry to reduce potential conflict.

> Note: The changes aim to reduce the possibility of race conditions with best efforts. However, due to the lack of transaction support in OpenSearch, any checks performed before acquiring an optimistic lock are still subject to **"dirty reads."** This means there remains a small possibility of inconsistencies during concurrent operations.

### Example

Prepare corrupted Flint index:

<pre>
# Create an auto refresh Flint index
CREATE SKIPPING INDEX ON glue.default.http_logs (
  year PARTITION
)
WITH (
  auto_refresh = true,
  checkpoint_location = 's3://checkpoint-1'
);

# Delete OS index once all data refreshed
DELETE flint_glue_default_http_logs_skipping_index

# Index monitor detects and terminates streaming job
24/09/30 18:27:43 WARN FlintSparkIndexMonitor: Streaming job is active but data is deleted
24/09/30 18:27:43 INFO FlintSparkIndexMonitor: Terminating streaming job and index monitor for
 flint_glue_default_http_logs_skipping_index

# Flint index log entry is dangling in failed state
GET .query_execution_request_glue/_search
          "version": "1.0",
          "latestId": "ZmxpbnRfZ2x1ZV9kZWZhdWx0X2h0dHBfbG9nc19za2lwcGluZ19pbmRleA==",
          "type": "flintindexstate",
          <b>"state": "failed",</b>
          "applicationId": "XXX",
          "jobId": "YYY",
          "dataSourceName": "glue",
          "jobStartTime": 1727720050924,
          "lastUpdateTime": 1727720863284,
          "error": ""
        }
</pre>

Verify cleanup logic works:

<pre>
# Attempt to create the same index again
CREATE SKIPPING INDEX ON glue.default.http_logs (
  year PARTITION
)
WITH (
  auto_refresh = true,
  checkpoint_location = 's3://checkpoint-2'
);

# Index metadata log entry is cleaned up
24/09/30 18:34:35 INFO FlintSpark: Starting index operation 
[Create Flint index flint_glue_default_http_logs_skipping_index] with forceInit=true
<b>24/09/30 18:34:35 WARN FlintSpark: 
 Cleaning up corrupted index:
 - logEntryExists [true]
 - dataIndexExists [false]
 - isCreatingOrVacuuming [false]</b>
24/09/30 18:34:39 INFO FlintSpark: Index operation
[Create Flint index flint_glue_default_http_logs_skipping_index] complete

GET .query_execution_request_glue/_search
          "version": "1.0",
          "latestId": "ZmxpbnRfZ2x1ZV9kZWZhdWx0X2h0dHBfbG9nc19za2lwcGluZ19pbmRleA==",
          "type": "flintindexstate",
          <b>"state": "refreshing",</b>
          "applicationId": "XXX",
          "jobId": "YYY",
          "dataSourceName": "glue",
          "jobStartTime": 1727721279398,
          "lastUpdateTime": 1727723128616,
          "error": ""
        }
</pre>

### Issues Resolved

https://github.com/opensearch-project/opensearch-spark/issues/356

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
@@ -159,18 +159,22 @@ class FlintSparkIndexMonitor(
     override def run(): Unit = {
       logInfo(s"Scheduler trigger index monitor task for $indexName")
       try {
-        if (isStreamingJobActive(indexName)) {
-          if (flintClient.exists(indexName)) {
-            logInfo("Streaming job is still active")
+        val isJobActive = isStreamingJobActive(indexName)
+        val indexExists = flintClient.exists(indexName)
+
+        (isJobActive, indexExists) match {
+          case (true, true) =>
+            logInfo("Streaming job is active and index exists")
             flintMetadataLogService.recordHeartbeat(indexName)
-          } else {
-            logWarning("Streaming job is active but data is deleted")
+
+          case (true, false) =>
+            logWarning("Streaming job is active but index is deleted")
             stopStreamingJobAndMonitor(indexName)
-          }
-        } else {
-          logError("Streaming job is not active. Cancelling monitor task")
-          stopMonitor(indexName)
-          logInfo("Index monitor task is cancelled")
+
+          case (false, _) =>
+            logError("Streaming job is not active. Cancelling monitor task")
+            stopMonitor(indexName)
+            logInfo("Index monitor task is cancelled")
         }
         errorCnt = 0 // Reset counter if no error
       } catch {
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
@@ -160,10 +160,10 @@ class FlintSparkIndexMonitor(
       logInfo(s"Scheduler trigger index monitor task for $indexName")
       try {
         if (isStreamingJobActive(indexName)) {
-          logInfo("Streaming job is still active")
-          flintMetadataLogService.recordHeartbeat(indexName)
-
-          if (!flintClient.exists(indexName)) {
+          if (flintClient.exists(indexName)) {
+            logInfo("Streaming job is still active")
+            flintMetadataLogService.recordHeartbeat(indexName)
+          } else {
             logWarning("Streaming job is active but data is deleted")
             stopStreamingJobAndMonitor(indexName)
           }
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkTransactionSupport.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkTransactionSupport.scala
@@ -46,14 +46,21 @@ trait FlintSparkTransactionSupport extends Logging {
    *   the type of the result produced by the operation block
    * @return
    *   Some(result) of the operation block if the operation is executed, or None if the operation
-   *   execution is bypassed due to missing index data
+   *   execution is bypassed due to index corrupted
    */
   def withTransaction[T](indexName: String, opName: String, forceInit: Boolean = false)(
       opBlock: OptimisticTransaction[T] => T): Option[T] = {
     logInfo(s"Starting index operation [$opName $indexName] with forceInit=$forceInit")
     try {
-      // Execute the action if data index exists or create index action (indicated by forceInit)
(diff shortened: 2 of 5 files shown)


noCharger (commented):
Will the PR also handle the issue where the create index fails with an opensearch timeout and the index status remains 'creating'?

seankao-az (approved):

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkTransactionSupport.scala
```diff
@@ -60,4 +79,42 @@ trait FlintSparkTransactionSupport { self: Logging =>
...
+  private def isIndexCorrupted(indexName: String): Boolean = {
+    val logEntry =
+      flintMetadataLogService
+        .getIndexMetadataLog(indexName)
+        .flatMap(_.getLatest)
+    val logEntryExists = logEntry.isPresent
+    val dataIndexExists = flintClient.exists(indexName)
+    val isCreatingOrVacuuming =
+      logEntry
+        .filter(e => e.state == EMPTY || e.state == CREATING || e.state == VACUUMING)
```
dai-chen, line 98: @noCharger We bypass corruption check for index in `CREATING` state here? This is mostly because of the concern mentioned in PR description.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
```diff
@@ -160,10 +160,10 @@ class FlintSparkIndexMonitor(
       logInfo(s"Scheduler trigger index monitor task for $indexName")
       try {
         if (isStreamingJobActive(indexName)) {
```
noCharger, line 162: nit: use `match` to avoid nested if else

noCharger (approved):

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexMonitor.scala
dai-chen: I tried the following changes but found IDEA or Scalafmt suggest to use if-else :) Any other idea?

```
        flintClient.exists(indexName) match {
          case true =>
            logInfo("Streaming job is still active")
            flintMetadataLogService.recordHeartbeat(indexName)
          case false =>
            logWarning("Streaming job is active but data is deleted")
            stopStreamingJobAndMonitor(indexName)
        }
```

noCharger: Something like this?

```
(isStreamingJobActive(indexName), flintClient.exists(indexName)) match {
  case (true, true) =>
    logInfo("Streaming job is active and index exists")
    flintMetadataLogService.recordHeartbeat(indexName)
    errorCnt = 0 // Reset counter if no error

  case (true, false) =>
    logWarning("Streaming job is active but index is deleted")
    stopStreamingJobAndMonitor(indexName)
    errorCnt = 0 // Reset counter if no error

  case (false, _) =>
    logError("Streaming job is not active. Cancelling monitor task")
    stopMonitor(indexName)
    logInfo("Index monitor task is cancelled")
}
```

dai-chen: Addressed in https://github.com/opensearch-project/opensearch-spark/pull/558/commits/7eb7eb334efabf1403db5b58930d651218846f50. Thanks!

PR #567, opened 2024-08-15
Add rate limiter for bulk request

ykmr1224:
### Description
- Add rate limiter for bulk request
- Without rate limit, it would send request until the destination OpenSearch throttle requests (especially when `NONE` refresh policy is used).
- Added config param `spark.datasource.flint.write.bulkRequestRateLimitPerNode` to specify the rate limit per node.
- The default is set to 0, which means no rate limit.

### Issues Resolved

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -527,6 +527,7 @@ In the index mapping, the `_meta` and `properties`field stores meta and schema i
 - `spark.datasource.flint.write.batch_size`: "The number of documents written to Flint in a single batch request. Default value is Integer.MAX_VALUE.
 - `spark.datasource.flint.write.batch_bytes`: The approximately amount of data in bytes written to Flint in a single batch request. The actual data write to OpenSearch may more than it. Default value is 1mb. The writing process checks after each document whether the total number of documents (docCount) has reached batch_size or the buffer size has surpassed batch_bytes. If either condition is met, the current batch is flushed and the document count resets to zero.
 - `spark.datasource.flint.write.refresh_policy`: default value is false. valid values [NONE(false), IMMEDIATE(true), WAIT_UNTIL(wait_for)]
+- `spark.datasource.flint.write.bulkRequestRateLimitPerNode`: default value is 0, which disables rate limit.
 - `spark.datasource.flint.read.scroll_size`: default value is 100.
 - `spark.datasource.flint.read.scroll_duration`: default value is 5 minutes. scroll context keep alive duration.
 - `spark.datasource.flint.retry.max_retries`: max retries on failed HTTP request. default value is 3. Use 0 to disable retry.
--- a/docs/index.md
+++ b/docs/index.md
@@ -521,13 +521,13 @@ In the index mapping, the `_meta` and `properties`field stores meta and schema i
 - `spark.datasource.flint.auth.username`: basic auth username.
 - `spark.datasource.flint.auth.password`: basic auth password.
 - `spark.datasource.flint.region`: default is us-west-2. only been used when auth=sigv4
-- `spark.datasource.flint.customAWSCredentialsProvider`: default is empty.   
+- `spark.datasource.flint.customAWSCredentialsProvider`: default is empty.
 - `spark.datasource.flint.write.id_name`: no default value.
 - `spark.datasource.flint.ignore.id_column` : default value is true.
 - `spark.datasource.flint.write.batch_size`: "The number of documents written to Flint in a single batch request. Default value is Integer.MAX_VALUE.
 - `spark.datasource.flint.write.batch_bytes`: The approximately amount of data in bytes written to Flint in a single batch request. The actual data write to OpenSearch may more than it. Default value is 1mb. The writing process checks after each document whether the total number of documents (docCount) has reached batch_size or the buffer size has surpassed batch_bytes. If either condition is met, the current batch is flushed and the document count resets to zero.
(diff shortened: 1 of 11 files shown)


flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
```diff
@@ -187,7 +193,7 @@ object FlintSparkConf {
       .doc("data source name")
       .createOptional()
   val CUSTOM_FLINT_METADATA_LOG_SERVICE_CLASS =
-    FlintConfig(FlintOptions.CUSTOM_FLINT_METADATA_LOG_SERVICE_CLASS)
+    FlintConfig(s"spark.datasource.flint.${FlintOptions.CUSTOM_FLINT_METADATA_LOG_SERVICE_CLASS}")
```
ykmr1224, line 196: It is not directly related, but I found this change is needed to make customFlintMetadataLogService config effective.

ykmr1224: Reverted this as it is addressed in another PR.

docs/index.md
```diff
@@ -527,6 +527,7 @@ In the index mapping, the `_meta` and `properties`field stores meta and schema i
 - `spark.datasource.flint.write.batch_size`: "The number of documents written to Flint in a single batch request. Default value is Integer.MAX_VALUE.
 - `spark.datasource.flint.write.batch_bytes`: The approximately amount of data in bytes written to Flint in a single batch request. The actual data write to OpenSearch may more than it. Default value is 1mb. The writing process checks after each document whether the total number of documents (docCount) has reached batch_size or the buffer size has surpassed batch_bytes. If either condition is met, the current batch is flushed and the document count resets to zero.
 - `spark.datasource.flint.write.refresh_policy`: default value is false. valid values [NONE(false), IMMEDIATE(true), WAIT_UNTIL(wait_for)]
+- `spark.datasource.flint.write.bulkRequestRateLimitPerNode`: default value is 0, which disables rate limit.
```
penghuo, line 530: 0.1 means 1 request per 10 seconds? could u add more in doc.

flint-core/src/main/scala/org/opensearch/flint/core/storage/BulkRequestRateLimiterHolder.java
```diff
@@ -0,0 +1,22 @@
...
+ * Hold shared instance of BulkRequestRateLimiter. This class is introduced to make
+ * BulkRequestRateLimiter testable and share single instance.
+ */
+public class BulkRequestRateLimiterHolder {
+
+  private static BulkRequestRateLimiter instance;
+
+  private BulkRequestRateLimiterHolder() {}
+
+  public synchronized static BulkRequestRateLimiter getBulkRequestRateLimiter(
```
penghuo, line 15: does Holder necessary? move to BulkRequestRateLimiter?

flint-core/src/main/java/org/opensearch/flint/core/RestHighLevelClientWrapper.java
```diff
@@ -55,13 +57,21 @@ public class RestHighLevelClientWrapper implements IRestHighLevelClient {
...
         this.client = client;
+        this.rateLimiter = rateLimiter;
     }
 
     @Override
     public BulkResponse bulk(BulkRequest bulkRequest, RequestOptions options) throws IOException {
-        return execute(OS_WRITE_OP_METRIC_PREFIX, () -> client.bulk(bulkRequest, options));
+      return execute(OS_WRITE_OP_METRIC_PREFIX, () -> {
+        try {
+          rateLimiter.acquirePermit();
```
penghuo, line 69: 1. Each bulk request contains multiple index request. if the throttle is on each index request, the bulk request limit may not help.
2. How does OpenSearch customer configure this paramater?

penghuo (commented):
During Index/mv refreshing, can user adjust this value if the customer under pressure.
My concern is that this configuration may be helpful during the testing stage but not in the production environment. We should investigate methods to dynamically control the ingestion rate.

docs/index.md
ykmr1224: It won't accept decimal value. We could support if we modify the rate limit period, but I think it would become complicated considering someone might specify a value like `1.23`. If we want to reduce the traffic less than 1 request/sec, we can reduce the size of batch.

flint-core/src/main/scala/org/opensearch/flint/core/storage/BulkRequestRateLimiterHolder.java
ykmr1224: As written in the comment, it is needed to use shared single instance, and make BulkRequestRateLimiter testable. If we directly make BulkRequestRateLimiter singleton, we cannot test it with different parameters.

flint-core/src/main/java/org/opensearch/flint/core/RestHighLevelClientWrapper.java
ykmr1224: 1. Rate limit on bulk request would limit the overall number of index request.
2. They can add it as an extra Spark parameter for now. We might want to add an attribute for Datasource so we can specify per datasource.

penghuo (approved):

ykmr1224:
Added `[Experimental]` to the doc.

PR #572, opened 2024-08-16
Retry bulk request to OpenSearch

ykmr1224:
### Description
- Retry bulk request to OpenSearch.
  - It retries only failed and retryable requests in the batch.
- There already is retry for other requests, but it won't be applied to bulk API, since bulk request itself will return 200 even if each request were throttled.
- This is to mitigate throttling when writing index to OpenSearch. When `NONE` refresh policy is used, bulk request will be responded quickly (even when the server is overloaded), and causes throttling.
- https://github.com/opensearch-project/opensearch-spark/pull/567 added rate limit, but we still need retry considering when the server is overloaded by other requests for short period of time.

### Issues Resolved


flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
```diff
@@ -0,0 +1,34 @@
...
+    if (Arrays.stream(bulkResponse.getItems()).anyMatch(itemResp -> !isCreateConflict(itemResp))) {
+      LOG.info("Found retryable failure in the bulk response");
+      return true;
+    }
+    return false;
+  }
+
+  private boolean isCreateConflict(BulkItemResponse itemResp) {
+    return itemResp.getOpType() == DocWriteRequest.OpType.CREATE && (itemResp.getFailure() == null
+        || itemResp.getFailure()
```
penghuo, line 31: nit: itemResp.getFailure().getStatus() in single line.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private boolean isRetryable(BulkResponse bulkResponse) {
+    if (Arrays.stream(bulkResponse.getItems()).anyMatch(itemResp -> !isCreateConflict(itemResp))) {
+      LOG.info("Found retryable failure in the bulk response");
+      return true;
+    }
+    return false;
+  }
+
+  private boolean isCreateConflict(BulkItemResponse itemResp) {
```
penghuo, line 29: there is no 429 exception?

flint-core/src/main/scala/org/opensearch/flint/core/http/FlintRetryOptions.java
```diff
@@ -71,13 +74,33 @@ public <T> RetryPolicy<T> getRetryPolicy() {
...
-            LOG.warning("Retrying failed request at #" + ex.getAttemptCount()))
+        .onFailedAttempt(FlintRetryOptions::onFailure)
+        .onRetry(FlintRetryOptions::onRetry)
         .build();
   }
 
+  public RetryPolicy<BulkResponse> getBulkRetryPolicy(CheckedPredicate<BulkResponse> resultPredicate) {
+    return RetryPolicy.<BulkResponse>builder()
+        // Using higher initial backoff to mitigate throttling quickly
+        .withBackoff(4, 30, SECONDS)
```
vamsi-amazon, line 85: Is this exponential backoff?
How will the progression of delay look like?

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
vamsi-amazon: Can you rename the method. createConflict is odd..does this mean the request is create and resulted in conflict. Is the intention to only retry requests with conflict failure?

vamsi-amazon:
If I understood correctly, whenever there is failure in bulk we will retry with exponential backoff...what is the retry policy earlier?

Why do we need separate backoff strategy apart from rate limiter?

Can you add parts of your design document to the PR description, so opensource users understand the change.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+
+  public BulkResponse bulkWithPartialRetry(RestHighLevelClient client, BulkRequest bulkRequest,
+      RequestOptions options) {
+    try {
+      final Holder<BulkRequest> nextRequest = new Holder<>(bulkRequest);
+      return Failsafe
+          .with(retryPolicy)
+          .get(() -> {
+            BulkResponse response = client.bulk(nextRequest.get(), options);
+            if (retryPolicy.getConfig().allowsRetries() && bulkItemErrorResultPredicate.test(
```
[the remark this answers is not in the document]
ykmr1224: Yes, this is coming from existing config: `retry.max_retries`. When it is set to 0, retry is disabled and it would return false.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+      LOG.severe("Request failed permanently. Re-throwing original exception.");
+
+      // unwrap original exception and throw
+      Throwable cause = ex.getCause();
+      throw new RuntimeException(cause);
+    }
+  }
+
+  // Holder class to let lambda expression update next BulkRequest
+  private static class Holder<T> {
```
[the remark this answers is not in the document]
ykmr1224: Thank you, looks usable. I'll change it.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+      return true;
+    }
+    return false;
+  }
+
+  private static boolean isItemRetryable(BulkItemResponse itemResponse) {
+    return itemResponse.isFailed() && !isCreateConflict(itemResponse);
+  }
+
+  private static boolean isCreateConflict(BulkItemResponse itemResp) {
```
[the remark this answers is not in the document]
ykmr1224: oh, I forgot to delete the class. Thanks.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+      this.item = item;
+    }
+  }
+
+  private BulkRequest getRetryableRequest(BulkRequest request, BulkResponse response) {
+    List<DocWriteRequest<?>> bulkItemRequests = request.requests();
+    BulkItemResponse[] bulkItemResponses = response.getItems();
+    BulkRequest nextRequest = new BulkRequest()
+        .setRefreshPolicy(request.getRefreshPolicy());
+    nextRequest.setParentTask(request.getParentTask());
```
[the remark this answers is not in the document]
ykmr1224: That indicate the parent task associated with this request. I was not able to find good description from the OpenSearch doc. It looks working like a tag for requests when checking from `_tasks` API. (we can filter tasks by parent taskId)
Copying the same value from original request to keep it same.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
ykmr1224: Here, it consider other than Conflict response for Create request is retryable.

flint-core/src/main/scala/org/opensearch/flint/core/http/FlintRetryOptions.java
ykmr1224: Yes, it is exponential backoff, and it would increase like 4 -> 16 -> 30 -> 30.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
ykmr1224: deleting the file

ykmr1224:
> If I understood correctly, whenever there is failure in bulk we will retry with exponential backoff...what is the retry policy earlier?

Originally, retry policy was effective only when whole request was failed. It was not applied when bulk request itself returned with 200, and each request failed.

> Why do we need separate backoff strategy apart from rate limiter?
> 
> Can you add parts of your design document to the PR description, so opensource users understand the change.

I put some description in the PR, but which part is missing or unclear?
I think retry is anyway needed for robustness. (Even if rate limit is well implemented, there could be throttling or other temporary issue time to time)
The rate limit might require some improvement for longer term.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final Logger LOG = Logger.getLogger(OpenSearchBulkRetryWrapper.class.getName());
+
+  private final RetryPolicy<BulkResponse> retryPolicy;
+
+  public OpenSearchBulkRetryWrapper(FlintRetryOptions retryOptions) {
+    this.retryPolicy = retryOptions.getBulkRetryPolicy(bulkItemErrorResultPredicate);
+  }
+
+  public BulkResponse bulkWithPartialRetry(RestHighLevelClient client, BulkRequest bulkRequest,
```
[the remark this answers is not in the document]
ykmr1224: Added

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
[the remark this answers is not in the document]
penghuo: nit, do we need to managed max_retries manually? does the RetryPolicy already handle it automatically?

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
penghuo:
```
itemResp.getOpType() == DocWriteRequest.OpType.CREATE && (itemResp.getFailure() == null
        || itemResp.getFailure().getStatus() == RestStatus.CONFLICT);
  }
```
Does conflict means throttled?

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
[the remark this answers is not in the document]
penghuo: I did not get it, tasks is OpenSearch internal concept, why the bulk request need to attach task info.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
ykmr1224: Here, `CONFLICT` means HTTP status `409 Conflict`, which indicates same request came to the same document at the same time, and we shouldn't retry. This logic is coming from original implementation to see the bulk request succeeded or not. `itemResp.getFailure() == null` is not needed here, and I'll fix it.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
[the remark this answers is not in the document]
ykmr1224: In this logic, it checks if retry is enabled so not to generate next retryable request when retry is disabled.

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkRetryWrapper.java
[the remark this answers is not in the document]
ykmr1224: We don't care task info, but as it is an attribute in BulkRequst, just inherit the value from original request so it would be consistent with original request. (inheriting as much as possible from the original request)

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/BulkItemErrorResultPredicate.java
ykmr1224: Fixed.

vamsi-amazon (approved):

PR #616, opened 2024-08-30
Update checkpoint location on alter path

noCharger:
### Description
Currently, there are two ways of creating a Flint index representation:

```
1. FlintSparkIndexFactory.create
2. FlintSparkIndexBuilder.create
```

The FlintSparkIndexFactory.create method doesn't call the option() method in FlintSparkIndexBuilder. Instead, it constructs options from metadata directly:

```
val indexOptions = FlintSparkIndexOptions(
  metadata.options.asScala.mapValues(_.asInstanceOf[String]).toMap)
```

This becomes problematic on the ALTER statement, which retrieves the index via describeIndex. The describeIndex method uses FlintSparkIndexFactory.create, bypassing the option() method call. As a result, the custom checkpoint location is not properly set for ALTER statements. This inconsistency needs to be addressed to ensure proper functionality across all index operations.

### Proposed Solution

Update the copyWithUpdate method to set the custom checkpoint location after FlintSparkIndexFactory.create is called in describeIndex method. This will ensure that the checkpoint location is correctly updated for ALTER statements, maintaining consistency with other index creation and modification methods.

### Steps to Reproduce
1. Create a batch job with auto_refresh=false
2. Create a streaming job with auto_refresh=true

```
Error:
Checkpoint location is required if spark.flint.index.checkpoint.mandatory option enabled"}}
java.lang.IllegalArgumentException: requirement failed: Checkpoint location is required if spark.flint.index.checkpoint.mandatory option enabled
```

### Expected Behavior

The ALTER statement should properly set and use the custom checkpoint location.

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
@@ -81,8 +81,9 @@ abstract class FlintSparkIndexBuilder(flint: FlintSpark) {
       index: FlintSparkIndex,
       updateOptions: FlintSparkIndexOptions): FlintSparkIndex = {
     val originalOptions = index.options
-    val updatedOptions =
-      originalOptions.copy(options = originalOptions.options ++ updateOptions.options)
+    val updatedOptions = updateOptionWithDefaultCheckpointLocation(
+      index.name(),
+      originalOptions.copy(options = originalOptions.options ++ updateOptions.options))
     val updatedMetadata = index
       .metadata()
       .copy(options = updatedOptions.options.mapValues(_.asInstanceOf[AnyRef]).asJava)
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
@@ -5,7 +5,7 @@
 
 package org.opensearch.flint.spark
 
-import java.util.{Collections, UUID}
+import java.util.Collections
 
 import scala.collection.JavaConverters.mapAsJavaMapConverter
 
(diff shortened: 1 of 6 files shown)


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
```diff
@@ -81,8 +81,9 @@ abstract class FlintSparkIndexBuilder(flint: FlintSpark) {
       index: FlintSparkIndex,
       updateOptions: FlintSparkIndexOptions): FlintSparkIndex = {
     val originalOptions = index.options
-    val updatedOptions =
-      originalOptions.copy(options = originalOptions.options ++ updateOptions.options)
+    val updatedOptions = updateOptionWithDefaultCheckpointLocation(
```
dai-chen, line 84: Is it better to move the core logic in `updateOptionWithDefaultCheckpointLocation` to `FlintSparkIndexOption.checkpointLocation(indexName)`? I just feel this default value logic belong to `FlintSparkIndexOption` abstraction.

integ-test/src/integration/scala/org/opensearch/flint/spark/FlintSparkCoveringIndexITSuite.scala
```diff
@@ -213,6 +213,59 @@ class FlintSparkCoveringIndexITSuite extends FlintSparkSuite {
...
+      checkAnswer(indexData, Seq())
+
+      var index = flint.describeIndex(testFlintIndex)
+      var checkpointLocation = index.get.options.checkpointLocation()
+      assert(checkpointLocation.isEmpty, "Checkpoint location should not be defined")
+
+      // 2. Update the spark conf with a custom checkpoint location
+      conf.setConfString(
+        FlintSparkConf.CHECKPOINT_LOCATION_ROOT_DIR.key,
+        checkpointDir.getAbsolutePath)
```
dai-chen, line 236: np: you can use `setFlintSparkConf` or add a `withFlintSparkConf` similar as `withSparkConf`.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
noCharger: refactored [c639be7](https://github.com/opensearch-project/opensearch-spark/pull/616/commits/c639be78675e40f5df81d8edc4a4dfaedf1d8f79)

> Is it better to move the core logic in `updateOptionWithDefaultCheckpointLocation` to `FlintSparkIndexOption.checkpointLocation(indexName)`? I just feel this default value logic belong to `FlintSparkIndexOption` abstraction.

dai-chen (approved):
Thanks for the changes!

PR #641, opened 2024-09-10
Disallow queries that reference catalog views

### LINKED ISSUE #534: [FEATURE] Add option to disallow catalog views
Catalog views cannot be trusted because they can inject custom code. To work with secure environments, an option should exist to disallow this SQL injection technique.

eirsep:
### Description
Disallow queries that reference catalog views

### Issues Resolved
#534  

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/DisallowCatalogViews.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/DisallowCatalogViews.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.flint.spark
+
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, View}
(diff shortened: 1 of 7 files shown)


noCharger (commented):
Can we add one IT for Sql as well?

penghuo (commented):
The code looks good.

While "Disable View Access" is a valid feature from a Managed Service perspective, it is not directly related to the core functionality of the opensearch-spark repository, which primarily provides a Spark Accelerator framework for enabling secondary indices to remote data stores.

Would it make sense to implement this feature in a separate repository?

dai-chen:
As @penghuo mentioned above, the changes are not directly related to Flint but it will disallow any SQL view created in Flint? I assume this is actually not the requirement for Flint Spark extension. I thought we just want to disable Glue Catalog View which may have the issue reported.

eirsep:
> As @penghuo mentioned above, the changes are not directly related to Flint but it will disallow any SQL view created in Flint? I assume this is actually not the requirement for Flint Spark extension. I thought we just want to disable Glue Catalog View which may have the issue reported.

What additional check can I add to ensure the View is actually a Glue catalog view?

penghuo:
@eirsep could we close this PR?

Issue #634, opened 2024-09-09
[BUG] delete index request to OpenSearch Serverless collection fails


ykmr1224:
**What is the bug?**
When OpenSearch Serverless collection is used as index store, it fails to delete index due to unrecognized parameters `master_timeout`, and `timeout`.
It is because those options are automatically set by the OpenSearch client library but serverless collection doesn't support them.

```
OpenSearch exception [type=illegal_argument_exception, reason=request [/flint_spark_catalog_default_ci_test_name_and_age_index] contains unrecognized parameters: [master_timeout], [timeout]]
OpenSearchStatusException[OpenSearch exception [type=illegal_argument_exception, reason=request [/flint_spark_catalog_default_ci_test_name_and_age_index] contains unrecognized parameters: [master_timeout], [timeout]]]
	at org.opensearch.rest.BytesRestResponse.errorFromXContent(BytesRestResponse.java:207)
	at org.opensearch.client.RestHighLevelClient.parseEntity(RestHighLevelClient.java:2228)
	at org.opensearch.client.RestHighLevelClient.parseResponseException(RestHighLevelClient.java:2205)
	at org.opensearch.client.RestHighLevelClient.internalPerformRequest(RestHighLevelClient.java:1924)
	at org.opensearch.client.RestHighLevelClient.performRequest(RestHighLevelClient.java:1877)
	at org.opensearch.client.RestHighLevelClient.performRequestAndParseEntity(RestHighLevelClient.java:1845)
	at org.opensearch.client.IndicesClient.delete(IndicesClient.java:118)
	at org.opensearch.flint.spark.FlintSparkSuite.$anonfun$deleteTestIndex$1(FlintSparkSuite.scala:82)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:38)
	at org.opensearch.flint.spark.FlintSparkSuite.deleteTestIndex(FlintSparkSuite.scala:64)
	at org.opensearch.flint.spark.FlintSparkSuite.deleteTestIndex$(FlintSparkSuite.scala:63)
	at org.opensearch.flint.spark.FlintSparkCoveringIndexITSuite.deleteTestIndex(FlintSparkCoveringIndexITSuite.scala:20)
[...]

PR #646, opened 2024-09-11
Disable timeout params for deleteIndex API only for serverless

ykmr1224:
### Description
- Disable timeout params for deleteIndex API only for serverless
- OpenSearch serverless does not accept `timeout` / `master_timeout` (= clusterManagerNodeTimeout) params, which is added by the RestHighLevelClient by default. (current version of `opensearch-java` library won't add it automatically, but we are still using `2.6.0` ([ref](https://github.com/opensearch-project/opensearch-spark/blob/fd3f82fc16273d47f17aa47e699aeaa8b4c3679d/build.sbt#L13-L15))

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/634

### DIFF
--- a/flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
@@ -9,6 +9,7 @@ import org.opensearch.action.admin.indices.delete.DeleteIndexRequest;
 import org.opensearch.client.RequestOptions;
 import org.opensearch.client.indices.CreateIndexRequest;
 import org.opensearch.client.indices.GetIndexRequest;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.flint.common.metadata.FlintMetadata;
 import org.opensearch.flint.core.FlintClient;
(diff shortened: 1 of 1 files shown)


noCharger (commented):
non-blocking: can we handle this in RestHighLevelClientWrapper and consider use DeleteIndexRequestBuilder?

ykmr1224:
> non-blocking: can we handle this in RestHighLevelClientWrapper and consider use DeleteIndexRequestBuilder?

RestHighLevelClientWrapper does not have reference to FlintOptions.
What do you mean by DeleteIndexRequestBuilder?

noCharger:
> > non-blocking: can we handle this in RestHighLevelClientWrapper and consider use DeleteIndexRequestBuilder?
> 
> RestHighLevelClientWrapper does not have reference to FlintOptions. What do you mean by DeleteIndexRequestBuilder?

Example:

```
public void deleteIndex(String indexName) {
    LOG.info("Deleting Flint index " + indexName);
    String osIndexName = sanitizeIndexName(indexName);
    try (IRestHighLevelClient client = createClient()) {
        DeleteIndexRequest request = buildDeleteIndexRequest(osIndexName);
        client.indices().delete(request, RequestOptions.DEFAULT);
    } catch (Exception e) {
        throw new IllegalStateException("Failed to delete Flint index " + osIndexName, e);
    }
}

private DeleteIndexRequest buildDeleteIndexRequest(String indexName) {
    DeleteIndexRequestBuilder builder = new DeleteIndexRequestBuilder(indexName);

    if (FlintOptions.SERVICE_NAME_AOSS.equals(options.getServiceName())) {
        // Disable timeouts for serverless environments
        builder.setClusterManagerNodeTimeout((TimeValue) null)
               .setTimeout((TimeValue) null);
    }

    return builder.build();
}

```

seankao-az (approved):
lgtm
+1 on Louis' idea on moving `new DeleteIndexRequest(osIndexName)` into build method.
but don't think it's necessary

ykmr1224:
I think current approach is more explicit and easier to read, though there could be some preference.

noCharger:
> I think current approach is more explicit and easier to read, though there could be some preference. btw, `DeleteIndexRequestBuilder` does not look like usable considering the constructor signature: `public DeleteIndexRequestBuilder(OpenSearchClient client, DeleteIndexAction action, String... indices)`

Actually, the DeleteIndexRequestBuilder from the library does not appear to be for external use; using the builder pattern externally is possible by extending it, but it is fine to leave it as is.

Issue #612, opened 2024-08-29
[BUG] PPL describe couldn't handle fully qualified name


seankao-az:
**What is the bug?**
`describe myglue_test.default.http_logs`
```
{
  "status": "FAILED",
  "error": "{\"Message\":\"Fail to run query. Cause: Invalid table name: myglue_test.default.http_logs Syntax: [ database_name. ] table_name\"}"
}
```

As a comparison, the following works:
`describe default.http_logs`
```
{
    "status": "SUCCESS",
    "schema": [
        {
            "name": "col_name",
            "type": "string"
        },
        {
            "name": "data_type",
            "type": "string"
        },
        {
            "name": "comment",
            "type": "string"
        }
    ],
    "datarows": [
        [
            "@timestamp",
            "timestamp",
            null
        ],
        [
            "clientip",
            "string",
            null
        ],
        [
            "request",
            "string",
            null
        ],
        [
            "status",
            "int",
            null
        ],
        [
            "size",
            "int",
            null
        ],
        [
            "year",
            "int",
            null
        ],
        [
            "month",
            "int",
            null
        ],
        [
            "day",
            "int",
            null
        ],
        [
            "# Partition Information",
            "",
            ""
        ],
        [
            "# col_name",
            "data_type",
            "comment"
        ],
        [
            "year",
            "int",
            null
        ],
        [
            "month",
            "int",
            null
        ],
        [
            "day",
            "int",
            null
        ],
        [
            "",
            "",
            ""
        ],
        [
            "# Detailed Table Information",
            "",
            ""
        ],
        [
            "Catalog",
            "spark_catalog",
            ""
        ],
        [
            "Database",
            "default",
            ""
        ],
        [
            "Table",
            "http_logs",
            ""
        ],
        [
            "Owner",
            "hadoop",
            ""
        ],
        [
            "Created Time",
            "Fri Aug 23 21:47:33 GMT 2024",
            ""
        ],
        [
            "Last Access",
            "UNKNOWN",
            ""
        ],
        [
            "Created By",
            "Spark 3.5.1-amzn-0",
            ""
        ],
        [
            "Type",
            "EXTERNAL",
            ""
        ],
        [
            "Provider",
            "json",
            ""
        ],
        [
            "Table Properties",
            "[isRegisteredWithLakeFormation=false]",
            ""
        ],
        [
            "Location",
            "s3://...",
            ""
        ],
        [
            "Serde Library",
[...]

PR #656, opened 2024-09-12
Fix ppl describe bug #612

YANG-DB:
### Description
fix describe ppl command to be aware of 3 parts FQN table that includes `schema.catalog.table`

### Issues Resolved
- https://github.com/opensearch-project/opensearch-spark/issues/612
- 


ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -132,6 +132,11 @@ public LogicalPlan visitRelation(Relation node, CatalystPlanContext context) {
                 identifier = new TableIdentifier(
                         node.getTableQualifiedName().getParts().get(1),
                         Option$.MODULE$.apply(node.getTableQualifiedName().getParts().get(0)));
+            } else if (node.getTableQualifiedName().getParts().size() == 3) {
```
LantaoJin, line 135: Seems the assumption here is relation names cannot contain `.`. But unfortunately, index name can contain `.`, ref https://opensearch.org/docs/latest/api-reference/index-apis/create-index/

PR #664, opened 2024-09-14
Add langType to FlintStatement model

noCharger:
### Description
Add langType to FlintStatement model

### Issues Resolved
N/A

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
@@ -13,6 +13,7 @@ import scala.concurrent.duration.{Duration, MINUTES}
 import scala.util.{Failure, Success, Try}
 
 import org.opensearch.flint.common.model.FlintStatement
+import org.opensearch.flint.common.scheduler.model.LangType
 import org.opensearch.flint.core.metrics.MetricConstants
 import org.opensearch.flint.core.metrics.MetricsUtil.incrementCounter
 import org.opensearch.flint.spark.FlintSpark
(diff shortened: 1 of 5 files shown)


YANG-DB (approved):

YANG-DB (approved):

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -71,7 +72,13 @@ case class JobOperator(
       instantiateStatementExecutionManager(commandContext, resultIndex, osClient)
 
     val statement =
-      new FlintStatement("running", query, "", queryId, currentTimeProvider.currentEpochMillis())
+      new FlintStatement(
+        "running",
+        query,
+        "",
+        queryId,
+        LangType.SQL,
```
seankao-az, line 80: Not a big concern but I feel like hard coded SQL isn't ideal. In principle our batch FlintJob could run even if provided with a PPL query.
However, if the language type isn't passed from async-query-core to the spark job by setting the spark conf, then really there's no way to get this lang type dynamically.

I know that our async-query-core use case always use FlintJob for Flint extension SQL queries only, so this could be fine

seankao-az (approved):

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
noCharger: Exactly, FlintJob is only used for SQL use case hense langType not passed to parameter https://github.com/opensearch-project/sql/blob/25064688db2d08e61579aaf7d21ff440c67605a4/async-query-core/src/main/java/org/opensearch/sql/spark/data/constants/SparkConstants.java#L88-L89. With warmpool the FlintStatement should come from async-query-core statement model.

Issue #416, opened 2024-07-10
[RFC] Direct Query External Query Scheduler


noCharger:
## Summary

This RFC proposes a new solution for managing automatic index refreshes in Flint, addressing the limitations of the current approach and providing improvements in user experience, cost management, and observability.

## Motivation

The current implementation of automatic index refreshes in Flint has several limitations:

**User Experience:** The ALTER clause does not support updating certain index options, such as the refresh_interval for auto-fresh indexes, requiring multiple queries.
**Cost Management:** Maintaining skipping indexes (SI), covering indexes (CI), or materialized views (MVs) with auto-refresh enabled requires a continuously running driver node, resulting in significant cost overhead when using EMR-S.
**Observability Challenges**: Users face challenges in understanding the index state and billing details when using auto-refresh.

## Proposed Solution

![image(2)](https://github.com/opensearch-project/opensearch-spark/assets/44716267/fbdb9487-e92c-4ba1-8b7b-0dd1113813d6)

The proposed solution involves the following modifications:

**Flint Index State:** After an index with auto refresh is created, it will remain in the ACTIVE state. Only during the refresh procedure does it move to the REFRESHING state.
**Query Scheduler Integration:** Flint will support external scheduler, and refresh operations will be scheduled based on the refresh_interval from the query.
**Scheduler Control:** Users can enable or disable the scheduler using the ALTER clause and change the refresh_interval directly.
**Streaming Job Management:** There is no need to cancel the streaming EMR job; instead, the external 
**Manual Execution:** Users can manually execute the REFRESH statement, but only one refresh operation can be active for any given index or MV at a time.

## Design Details

**SQL Syntax Changes**

A new field, `scheduler_mode`, will be introduced to control the scheduling behavior. When auto_refresh=true and `scheduler_mode=external`, the refresh is trigger by external scheduler.

**Scheduler Service**

A new scheduler service interface will be introduced, providing methods for scheduling, unscheduling, and updating jobs. Both the SQL Plugin and Flint will implement this interface, allowing for seamless integration with different components and services.

**Job Index**

The OpenSearch JobScheduler plugin will be used to manage the scheduled refresh jobs, as the default scheduler service implementation. Each `SI/CI/MV` will have a corresponding job index document, which will store the scheduling information and metadata.

**Scheduled Job Runner**

The `ScheduledJobRunner` interface will be implemented to handle the logic required for refreshing queries periodically. A locking service will be employed to ensure that job executions do not overlap.

[...]

Issue #622, opened 2024-09-04
Optimization: Use external scheduler with configurable interval


noCharger:
Flint can dynamically select between using the spark scheduler or the external scheduler based on the interval threshold level.

Issue #418, opened 2024-07-10
Introduce scheduling service


noCharger:

dblock:
[Catch All Triage - [1](https://github.com/dblock), [2](https://github.com/krisfreedain), [3](https://github.com/andrross), [4](https://github.com/reta)]

PR #668, opened 2024-09-16
Introduce async query scheduler

noCharger:
### Description
1. Introduce `AsyncQueryScheduler` interface and `AsyncQuerySchedulerRequest` request model to support external scheduler with methods for scheduling, unscheduling, and updating refresh index job.
2. Add `AsyncQuerySchedulerBuilder` to support custom implementations of `AsyncQueryScheduler`, via new config `spark.datasource.flint.customFlintSchedulerClass`
3. Add default external scheduler via opensearch job-scheduler plugin. Interval granularity of `OpenSearchAsyncQueryScheduler` will be in minutes.
3. Introduce new configs `spark.flint.job.externalScheduler.enabled` and `spark.flint.job.externalScheduler.interval`
to dynamically select between using the spark scheduler or the external scheduler based on the interval threshold level.

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/416
https://github.com/opensearch-project/opensearch-spark/issues/418
https://github.com/opensearch-project/opensearch-spark/issues/622

### Follow-up 
1. Add metrics on refresh
2. Add scheduler support on recover 


flint-commons/src/main/scala/org/opensearch/flint/common/scheduler/model/AsyncQuerySchedulerRequest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import lombok.Builder;
+import lombok.Data;
+
+import java.time.Instant;
+
+/** Represents a job request for a scheduled task. */
+@Builder
+@Data
+public class AsyncQuerySchedulerRequest {
```
dai-chen, line 16: Should we move all these classes to a new scheduler module? Because it's not directly related to Flint core and cannot be used by Flint core lib user (specific Flint index and its refresh is defined in Flint Spark layer)?

flint-core/src/main/java/org/opensearch/flint/core/scheduler/AsyncQuerySchedulerBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+  public enum AsyncQuerySchedulerAction {
+    SCHEDULE,
+    UPDATE,
+    UNSCHEDULE,
+    REMOVE
+  }
+
+  public static AsyncQueryScheduler build(FlintOptions options) {
+    String className = options.getCustomAsyncQuerySchedulerClass();
+    logger.info("className: {}", className);
```
dai-chen, line 33: np: more specific like `...info('AsyncQueryScheduler className: ...')` ?

flint-core/src/main/java/org/opensearch/flint/core/scheduler/OpenSearchAsyncQueryScheduler.java
```diff
@@ -0,0 +1,288 @@
...
+        } else if (e.status() == RestStatus.NOT_FOUND) {
+            throw new IllegalArgumentException("Job: " + jobId + " doesn't exist", e);
+        } else {
+            throw new RuntimeException("Unexpected OpenSearch exception while handling job: " + jobId, e);
+        }
+    }
+
+    private void handleException(String message, String identifier, Throwable e) {
+        if (e instanceof IllegalArgumentException) {
+            throw (IllegalArgumentException) e;
```
dai-chen, line 215: what does this mean?

flint-core/src/main/java/org/opensearch/flint/core/scheduler/util/IntervalSchedulerParser.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.spark.sql.execution.streaming.Triggers;
+import org.opensearch.jobscheduler.spi.schedule.IntervalSchedule;
+import org.opensearch.jobscheduler.spi.schedule.Schedule;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+
+public class IntervalSchedulerParser {
+
+  public static Schedule parse(Object schedule) {
```
dai-chen, line 17: add missing Javadoc? Could you elaborate a little why schedule can be different type?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
```diff
@@ -156,17 +168,35 @@ abstract class FlintSparkIndexBuilder(flint: FlintSpark) {
...
+    // Update scheduler mode by default
+    if (options.autoRefresh() && !updatedOptions.contains(SCHEDULER_MODE.toString)) {
+      val externalSchedulerEnabled = flintSparkConf.isExternalSchedulerEnabled
+      // If refresh interval is not set, or it is set but the interval is smaller than the threshold, use spark internal scheduler
+      val shouldUseExternalScheduler = options.refreshInterval().exists { interval =>
+        parseStringSchedule(
+          flintSparkConf.externalSchedulerIntervalThreshold()).getInterval >= parseStringSchedule(
+          interval).getInterval
+      }
+      if (externalSchedulerEnabled && shouldUseExternalScheduler) {
```
dai-chen, line 193: wondering why we decide to use external scheduler without considering threshold elsewhere? Did I miss anything?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/refresh/FlintSparkIndexRefresh.scala
```diff
@@ -86,7 +86,7 @@ object FlintSparkIndexRefresh {
   def create(indexName: String, index: FlintSparkIndex): FlintSparkIndexRefresh = {
     val options = index.options
     // TODO: Refactor the refresh class names and RefreshMode enum to distinguish Spark side physical trigger setup from the logical Flint index option.
-    if (SchedulerMode.EXTERNAL == options.schedulerMode()) {
+    if (options.isExternalSchedulerEnabled()) {
```
dai-chen, line 89: Following my previous comment and wondering if possible to have new abstraction or at least do this check at single place. Currently it seems we do this check on each layer, from AST builder to Flint Spark API to this lower level impl class. Just feel it's a little hard to maintain in future.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/scheduler/util/RefreshQueryGenerator.scala
```diff
@@ -0,0 +1,38 @@
...
+   *   if the provided parameters are invalid
+   */
+  def generateRefreshQuery(index: FlintSparkIndex): String = {
+    index match {
+      case skippingIndex: FlintSparkSkippingIndex =>
+        s"REFRESH SKIPPING INDEX ON ${skippingIndex.tableName}"
+      case coveringIndex: FlintSparkCoveringIndex =>
+        s"REFRESH INDEX ${index.name()} ON ${coveringIndex.tableName}"
+      case _: FlintSparkMaterializedView =>
+        s"REFRESH MATERIALIZED VIEW ${index.name()}"
```
dai-chen, line 32: Is this supposed to be `mvName`? I think `index.name()` is the full Flint index name. Any IT to cover this?

flint-spark-integration/src/test/scala/org/opensearch/flint/spark/scheduler/util/RefreshQueryGeneratorSuite.scala
```diff
@@ -0,0 +1,52 @@
...
+    when(mockIndex.name()).thenReturn("testIndex")
+    when(mockIndex.tableName).thenReturn("testTable")
+
+    val result = RefreshQueryGenerator.generateRefreshQuery(mockIndex)
+    result shouldBe "REFRESH INDEX testIndex ON testTable"
+  }
+
+  test("generateRefreshQuery should return correct query for FlintSparkMaterializedView") {
+    val mockIndex = mock(classOf[FlintSparkMaterializedView])
+    when(mockIndex.name()).thenReturn("testMV")
```
dai-chen, line 38: Following last comment, I guess this assumption is wrong and we need to create real object in this UT.

integ-test/src/integration/scala/org/opensearch/flint/spark/FlintSparkCoveringIndexSqlITSuite.scala
```diff
@@ -155,6 +155,11 @@ class FlintSparkCoveringIndexSqlITSuite extends FlintSparkSuite {
            | )
            | """.stripMargin)
 
+      val indexData = flint.queryIndex(testFlintIndex)
```
dai-chen, line 158: Should we add more assertion on external scheduler state? Otherwise these IT won't cover most of the changes in this PR?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -122,7 +133,11 @@ class FlintSpark(val spark: SparkSession) extends FlintSparkTransactionSupport w
               flintClient.createIndex(indexName, metadata.copy(latestId = Some(latest.id)))
               flintIndexMetadataService
                 .updateIndexMetadata(indexName, metadata.copy(latestId = Some(latest.id)))
-            })
+            }
+            if (isExternalSchedulerEnabled(index)) {
+              handleAsyncQueryScheduler(index, AsyncQuerySchedulerAction.SCHEDULE)
+            }
+          })
```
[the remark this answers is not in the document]
noCharger: RefreshIndex will perform the actual refresh action

flint-commons/src/main/scala/org/opensearch/flint/common/scheduler/model/AsyncQuerySchedulerRequest.java
noCharger: The interface and request model is in flint-commons, the implementation is in flint-core

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/refresh/FlintSparkIndexRefresh.scala
noCharger: Originally I thought a new implementation of

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/refresh/AutoIndexRefresh.scala
```diff
@@ -109,7 +109,9 @@ class AutoIndexRefresh(indexName: String, index: FlintSparkIndex)
         .addCheckpointLocation(options.checkpointLocation(), flintSparkConf.isCheckpointMandatory)
         .addRefreshInterval(options.refreshInterval())
         .addAvailableNowTrigger(
-          SchedulerMode.EXTERNAL == options.schedulerMode() || options.incrementalRefresh())
+          options
+            .isExternalSchedulerEnabled() || options
+            .incrementalRefresh())
```
seankao-az, line 114: These line formatting is a bit confusing. Line change makes it look like it should be a chain, but they're condition A OR condition B.

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/scheduler/util/RefreshQueryGenerator.scala
```diff
@@ -0,0 +1,38 @@
...
+   *   the generated scheduled query string
+   * @throws IllegalArgumentException
+   *   if the provided parameters are invalid
+   */
+  def generateRefreshQuery(index: FlintSparkIndex): String = {
+    index match {
+      case skippingIndex: FlintSparkSkippingIndex =>
+        s"REFRESH SKIPPING INDEX ON ${skippingIndex.tableName}"
+      case coveringIndex: FlintSparkCoveringIndex =>
+        s"REFRESH INDEX ${index.name()} ON ${coveringIndex.tableName}"
```
seankao-az, line 30: should be coveringIndex.indexName

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/scheduler/util/RefreshQueryGenerator.scala
seankao-az: +1 to mvName.
Also np: some case uses `index` and some case uses `skippingIndex`, `coveringIndex`. I think they could all just be `index`

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -238,16 +253,22 @@ class FlintSpark(val spark: SparkSession) extends FlintSparkTransactionSupport w
...
-            true
+            if (isExternalSchedulerEnabled(index.get)) {
+              handleAsyncQueryScheduler(index.get, AsyncQuerySchedulerAction.UNSCHEDULE)
+              true
+            } else {
+              // TODO: share same transaction for now
+              flintIndexMonitor.stopMonitor(indexName)
+              stopRefreshingJob(indexName)
+              true
+            }
```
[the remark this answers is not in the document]
seankao-az: I'm thinking would it be a good idea to extend `FlintSparkIndexRefresh` or `AutoIndexRefresh`, and encapsulate the scheduler logic inside it. So we continue to use refreshIndex() api.

`FlintSparkIndexRefresh.start()` would schedule the job.
Similarly, `stopRefreshingJob` and unschedule job will need to be abstracted behind a new interface `FlintSparkIndexRefresh.stop()`
And `cleanUp()` for vacuumIndex and `update()` for updateIndex

Something like:
```
// for deleteIndex
val indexRefresh = FlintSparkIndexRefresh.create(indexName, null) // there's no FlintSparkIndex object for delete
// ...
indexRefresh.stop() // no-op for most refresh. Only streaming / scheduler will use this.
```
This can remove most condition checks in FlintSpark, AST builder, etc.

Although FlintIndexMonitor would still need the check, I think.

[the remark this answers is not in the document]
noCharger: > I'm thinking would it be a good idea to extend `FlintSparkIndexRefresh` or `AutoIndexRefresh`, and encapsulate the scheduler logic inside it. So we continue to use refreshIndex() api.

I investigated this option but it's not suitable because `FlintSparkIndexRefresh.start()` will perform the actual refresh action but the scheduler will only "plan" for the future refresh

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexBuilder.scala
noCharger: `updateOptionsWithDefaults` will update the option object based on threshold.

flint-core/src/main/java/org/opensearch/flint/core/scheduler/OpenSearchAsyncQueryScheduler.java
noCharger: > what does this mean?

Rethrow this at top level since some IllegalArgumentException comes from the response handler

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/scheduler/util/RefreshQueryGenerator.scala
noCharger: > I think they could all just be index

tableName, indexName and mvName belong to the child class, can't access from parent

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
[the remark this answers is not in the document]
noCharger: > Looks like most if-else added has a pattern (except vacuum index):
```
if external scheduler enabled:
   asyncQueryScheduler.XXXaction
else:
   index monitor.start/stop
   start/stop refresh job
```
Brainstorming: Comparing external scheduler behavior with internal scheduler 
createIndex -> SCHEDULE (create job) VS no AI (since refresh is another transaction) 
vacuumIndex ->  REMOVE (delete job) VS no AI
deleteIndex / updateIndexAutoToManual -> UNSCHEDULE (disable job)  VS stop streaming job
updateIndexManualToAuto -> UPDATE (update job) VS start streaming job

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/refresh/FlintSparkIndexRefresh.scala
noCharger: > Currently it seems we do this check on each layer, from AST builder to Flint Spark API to this lower level impl class.

I think the reason here is scheduler mode belongs to auto_refresh and `options.autoRefresh()` is called in all layers

flint-core/src/main/scala/org/opensearch/flint/core/FlintOptions.java
```diff
@@ -177,6 +179,12 @@ public String getDataSourceName() {
     return options.getOrDefault(DATA_SOURCE_NAME, "");
   }
 
+  public String getClientId() {
```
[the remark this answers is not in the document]
noCharger: renamed to `getAWSAccountId` with javadoc

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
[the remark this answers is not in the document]
noCharger: Introduced abstraction in 0f7ebaf601bf8b414afebda5cec07440efe25a12
FlintSparkJobSchedulingService
-> FlintSparkJobExternalSchedulingService
-> FlintSparkJobInternalSchedulingService

flint-commons/src/main/scala/org/opensearch/flint/common/scheduler/model/AsyncQuerySchedulerRequest.java
noCharger: updated commit 0f7ebaf601bf8b414afebda5cec07440efe25a12 to move the implementation to flint-spark-integration as discussed offline

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/refresh/FlintSparkIndexRefresh.scala
dai-chen: Feel free to open issue to track the ideas. Thanks!

flint-core/src/main/java/org/opensearch/flint/core/scheduler/util/IntervalSchedulerParser.java
noCharger: resolved on commit https://github.com/opensearch-project/opensearch-spark/commit/0f7ebaf601bf8b414afebda5cec07440efe25a12 to simplify the parsing logic

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
```diff
@@ -122,7 +135,9 @@ class FlintSpark(val spark: SparkSession) extends FlintSparkTransactionSupport w
               flintClient.createIndex(indexName, metadata.copy(latestId = Some(latest.id)))
               flintIndexMetadataService
                 .updateIndexMetadata(indexName, metadata.copy(latestId = Some(latest.id)))
-            })
+            }
+            jobSchedulingService.handleJob(index, AsyncQuerySchedulerAction.SCHEDULE)
```
dai-chen, line 139: Still have a little confusion here: we schedule without checking if index is auto refreshed? Also the Spark impl will start index monitor at this point (earlier than actually starting Spark streaming job in `refreshIndex`)?

flint-spark-integration/src/test/scala/org/opensearch/flint/spark/scheduler/util/RefreshQueryGeneratorSuite.scala
noCharger: 0f7ebaf601bf8b414afebda5cec07440efe25a12 updated SQL ITs with AsyncQuerySchedulerForSqlIT which invokes the generated refresh query in one pass

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
noCharger: my bad... based on https://github.com/opensearch-project/opensearch-spark/pull/668#discussion_r1769480827, this should be no AI for internal scheduler

dai-chen: I had a previous comment regarding IT. So this is not covered by IT in this Spark repo?

noCharger: let me check if i can spy the index monitor

dai-chen: FYI, I recall we disable index monitor by default in base IT class.

noCharger: @dai-chen  Added a IT on 0c53dc32b12ea27635d8e15e70e6bc6ee38c2c93

noCharger:
CI passed on [6831551](https://github.com/opensearch-project/opensearch-spark/pull/668/commits/68315516e139f636f2c1a3ddda300a3fc6b07f3d)

dai-chen (approved):
Thanks for the changes!

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
noCharger: > @dai-chen Added a IT on [0c53dc3](https://github.com/opensearch-project/opensearch-spark/commit/0c53dc32b12ea27635d8e15e70e6bc6ee38c2c93)

revert back to [6831551](https://github.com/opensearch-project/opensearch-spark/pull/668/commits/68315516e139f636f2c1a3ddda300a3fc6b07f3d)

Issue #593, opened 2024-08-21
[BUG] FlintDataType cannot serialize TIMESTAMP_NTZ


engechas:
**What is the bug?**
Creating a covering index against a field with the type TIMESTAMP_NTZ throws an exception during serialization in the FlintDataType class.

```
24/08/21 23:01:33 ERROR FlintSpark: Failed to execute index operation [Create Flint index flint_hurneyttestsource_amazon_security_lake_glue_db_us_east_1_amazon_security_lake_table_us_east_1_vpc_flow_2_0_hurneyttest1_index]
java.lang.IllegalStateException: unsupported data type: TIMESTAMP_NTZ
	at org.apache.spark.sql.flint.datatype.FlintDataType$.serializeField(FlintDataType.scala:166) ~[opensearch-spark-standalone_2.12-latest.jar:0.5.0-SNAPSHOT]
	at org.apache.spark.sql.flint.datatype.FlintDataType$.$anonfun$serializeJValue$1(FlintDataType.scala:121) ~[opensearch-spark-standalone_2.12-latest.jar:0.5.0-SNAPSHOT]
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286) ~[scala-library-2.12.17.jar:?]
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36) ~[scala-library-2.12.17.jar:?]
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33) ~[scala-library-2.12.17.jar:?]
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:198) ~[scala-library-2.12.17.jar:?]
	at scala.collection.TraversableLike.map(TraversableLike.scala:286) ~[scala-library-2.12.17.jar:?]
	at scala.collection.TraversableLike.map$(TraversableLike.scala:279) ~[scala-library-2.12.17.jar:?]
	at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:198) ~[scala-library-2.12.17.jar:?]
	at org.apache.spark.sql.flint.datatype.FlintDataType$.serializeJValue(FlintDataType.scala:121) ~[opensearch-spark-standalone_2.12-latest.jar:0.5.0-SNAPSHOT]
[...]

PR #673, opened 2024-09-17
Add support for serializing TimestampNTZType

engechas:
### Description
Adds support for serializing TimestampNTZType by considering it the same case as a TimestampType

### Issues Resolved
N/A

### DIFF
--- a/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
+++ b/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
@@ -144,7 +144,7 @@ object FlintDataType {
       case FloatType => JObject("type" -> JString("float"))
 
       // Date
-      case TimestampType =>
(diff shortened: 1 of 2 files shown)


noCharger (approved):

dai-chen (approved):
Thanks for the fix! Just wonder there is no difference between the two types for OpenSearch? Or this is just a quick fix?

engechas:
>Thanks for the fix! Just wonder there is no difference between the two types for OpenSearch? Or this is just a quick fix?

There doesn't seem to be. Both types are ultimately dates in OpenSearch and use the same format, one just has a TZ attached whereas the other doesn't

dai-chen:
@noCharger Not sure whether we want to backport this. Please label accordingly.

dai-chen:
@engechas Does this PR fix https://github.com/opensearch-project/opensearch-spark/issues/593 ?

engechas:
>@engechas Does this PR resolve https://github.com/opensearch-project/opensearch-spark/issues/593 ?

Yes, will close it

PR #682, opened 2024-09-19
Optional support for getAllIndexMetadata by index pattern

seankao-az:
### Description
`FlintIndexMetadataService` used to use index pattern (using wildcard character `*`) to fetch index metadata for indexes matching the pattern. This change makes supporting getting by pattern optional. Instead, the service could fetch index metadata for full index names.

This does not change the `FlintSpark.describeIndexes` api, which still takes an index pattern as argument.

#### Alternative Considered
Alternative is to not use index pattern in `FlintIndexMetadataService.getAllIndexMetadata` at all, and instead always take full index names. However, this adds number of requests overhead for services that do allow for pattern matching, e.g. OpenSearch implementation `FlintOpenSearchIndexMetadataService`. Instead of making 1 request, a pattern that matches to N indexes would require N requests to fetch their metadata, if getting by pattern is removed completely.

### Issues Resolved
* #371 

### DIFF
--- a/flint-commons/src/main/scala/org/opensearch/flint/common/metadata/FlintIndexMetadataService.java
+++ b/flint-commons/src/main/scala/org/opensearch/flint/common/metadata/FlintIndexMetadataService.java
@@ -26,13 +26,21 @@ public interface FlintIndexMetadataService {
   FlintMetadata getIndexMetadata(String indexName);
 
   /**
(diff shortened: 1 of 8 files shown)


dai-chen (approved):

seankao-az:
removed some unused import

noCharger (commented):
@seankao-az will this backport to 0.5?

PR #685, opened 2024-09-20
Ppl help command

YANG-DB:
### Description
Add PPL `-help` command to assist grammar explaining and assisting users on different dialects and semantic

```sql
search -help
```

results in:
```
SEARCH Command:

Syntax:
   (SEARCH)? fromClause
   | (SEARCH)? fromClause logicalExpression
   | (SEARCH)? logicalExpression fromClause

Description:
The SEARCH command is used to retrieve data from a specified source. It can be used with or without additional filters.
- You can specify the data source using the FROM clause.
- You can add filters using logical expressions.
- The order of FROM clause and logical expression can be interchanged.

```

### Issues Resolved
- https://github.com/opensearch-project/opensearch-spark/issues/658


YANG-DB:
@LantaoJin 
I would appreciate your feedback on the approach i've drafter here for adding `help` parameter attribute for each command, for example the following would provide information and examples for the `search` command:
```sql
search -help
```
This would be done for all commands we want to add the `help` description to.
Implementation conceps - 
 - Added a dedicated `AstCommandDescriptionVisitor` which is an additional visitor (to the ASTBuilder ) that is only responsible of the commands decoration visiting.
 - The result is that `AstStatementBuilder` now hold a chain of visitors that run on each query and produce a plan (or not if they dont apply)
 - `FlintPPLSparkExtensions` was added with `injectPlannerStrategy` to override the standard queries in case of a `PrintLiteralCommandDescriptionLogicalPlan` plan and to avoid reaching to the underlying SQL engine and responding with the actual planner's text that was generated during the former stage

Please let me know if this sounds reasonable and if you have any suggestions to make this simpler
Thanks

LantaoJin:
Hi @YANG-DB , I haven't get the point what is the help command for. 
> The result is that AstStatementBuilder now hold a chain of visitors that run on each query and produce a plan (or not if they dont apply)

Seems you are going to add a `debug` command to display the translated Spark unresolved plan tree. `help` sounds like a customer towards command to show usage of PPL. 

If your proposal is for `debug` purpose, I think the display info not only includes the unresolved plan, but also the physical plan (which will reach to the underlying SQL engine) with a flag.

YANG-DB:
@YANG-DB plz resolve conflict

YANG-DB:
closing this as it has not gained sufficient attention

PR #689, opened 2024-09-24
Add resultIndex to session manager extension

noCharger:
### Description
Add resultIndex to session manager extension constructor. In this case custom session manager can work with default QueryResultWriter to persist result to OpenSearch

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
@@ -983,7 +983,8 @@ object FlintREPL extends Logging with FlintJobExecutor {
       resultIndexOption: Option[String]): SessionManager = {
     instantiate(
       new SessionManagerImpl(spark, resultIndexOption),
-      spark.conf.get(FlintSparkConf.CUSTOM_SESSION_MANAGER.key, ""))
+      spark.conf.get(FlintSparkConf.CUSTOM_SESSION_MANAGER.key, ""),
+      resultIndexOption.getOrElse(""))
   }
 
   private def instantiateStatementExecutionManager(
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/QueryResultWriterImpl.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/QueryResultWriterImpl.scala
@@ -9,11 +9,13 @@ import org.opensearch.flint.common.model.FlintStatement
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.FlintJob.writeDataFrameToOpensearch
+import org.apache.spark.sql.flint.config.FlintSparkConf
 
 class QueryResultWriterImpl(context: Map[String, Any]) extends QueryResultWriter with Logging {
 
   private val resultIndex = context("resultIndex").asInstanceOf[String]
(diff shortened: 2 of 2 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
```diff
@@ -983,7 +983,8 @@ object FlintREPL extends Logging with FlintJobExecutor {
       resultIndexOption: Option[String]): SessionManager = {
     instantiate(
       new SessionManagerImpl(spark, resultIndexOption),
-      spark.conf.get(FlintSparkConf.CUSTOM_SESSION_MANAGER.key, ""))
+      spark.conf.get(FlintSparkConf.CUSTOM_SESSION_MANAGER.key, ""),
+      resultIndexOption)
```
dai-chen, line 987: Just to confirm: custom session manager constructor in Java can accept Scala's Option[String]?

spark-sql-application/src/main/scala/org/apache/spark/sql/QueryResultWriterImpl.scala
```diff
@@ -9,11 +9,12 @@ import org.opensearch.flint.common.model.FlintStatement
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.FlintJob.writeDataFrameToOpensearch
+import org.apache.spark.sql.flint.config.FlintSparkConf
 
 class QueryResultWriterImpl(context: Map[String, Any]) extends QueryResultWriter with Logging {
 
   private val resultIndex = context("resultIndex").asInstanceOf[String]
-  private val osClient = context("osClient").asInstanceOf[OSClient]
+  private val osClient = new OSClient(FlintSparkConf().flintOptions())
```
dai-chen, line 17: Sorry could you elaborate this change?

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
noCharger: Java project needs to add Scala's library. For example
```
import scala.Option;
import scala.Some;
import scala.None$;

MyClass myClass = new MyClass(Some.apply("Hello")); // using Scala's Some
```
But I think it makes more sense to convert this as String

spark-sql-application/src/main/scala/org/apache/spark/sql/QueryResultWriterImpl.scala
noCharger: Default session manager impl keeps an OSClient in context map and expose it to OpenSearch QueryResultWriter. For the custom session manager, it doesn't make sense to do so to work with  OpenSearch QueryResultWriter.

seankao-az: A bit unintuitive.
Could we add a comment here? Because if we're just reading the code (our OpenSearch implementations) we might wonder why we're not using the osClient available in context.
If I'm understanding correctly, here we're not using context("osClient") because we can't rely on the session manager implementation to put osClient in context?

noCharger: sure, comment added

dai-chen (approved):

PR #701, opened 2024-09-25
[REFACTOR] Move DF reformat from StatementExecutionManagerImpl to QueryResultWriterImpl

noCharger:
### Description
Move DF reformat from `StatementExecutionManagerImpl` to `QueryResultWriterImpl` so that people use custom `StatementExecutionManager` still can work with default `QueryResultWriterImpl` and write result to OpenSearch result index for custom data source.

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
@@ -810,8 +810,13 @@ object FlintREPL extends Logging with FlintJobExecutor {
     } else {
       val futureQueryExecution = Future {
         val startTime = System.currentTimeMillis()
+        // Execute the statement and get the resulting DataFrame
+        // This step may involve Spark transformations, but not necessarily actions
         val df = statementsExecutionManager.executeStatement(flintStatement)
-        queryResultWriter.reformatDataFrame(df, flintStatement, startTime)
+        // Process the DataFrame, applying any necessary transformations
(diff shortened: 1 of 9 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
```diff
@@ -801,7 +809,9 @@ object FlintREPL extends Logging with FlintJobExecutor {
         startTime)
     } else {
       val futureQueryExecution = Future {
-        statementsExecutionManager.executeStatement(flintStatement)
+        val startTime = System.currentTimeMillis()
+        val df = statementsExecutionManager.executeStatement(flintStatement)
+        queryResultWriter.reformatDataFrame(df, flintStatement, startTime)
```
dai-chen, line 814: just wonder why query result writer has to expose this new API and cannot do this in `writeDataFrame` API? Anything happen after this and before writeDataFrame called?

dai-chen (approved):

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
noCharger: > just wonder why query result writer has to expose this new API and cannot do this in `writeDataFrame` API? Anything happen after this and before writeDataFrame called?

Reformat data will collect results on driver node as an termination state. If only`spark.sql()` called without a termination on the same thread, would it cause any issue?

seankao-az (approved):

PR #706, opened 2024-09-26
PPL `cidr` function

salyh:
### Description
Implement cidr function to check wether an ip address is with a given cidr

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/671


salyh:
@YANG-DB @vamshin @anirudha 

I pushed the scaffold for the CIDR function.
The real implementation which validates the ip address and check if its within the cidr is missing because we need to discuss the introduction of a new 3rd party library:

I suggest to add a dependency to https://github.com/seancfoley/IPAddress because this library seems to provide ip validation and in range checking for ipv4 and ipv6 addresses/cidrs. The library is Apache licensed.

Other options are the following libraries or code snippets (or of course manual implementation with java standard library which is error prone in my view):

- https://www.baeldung.com/java-check-ip-address-range
- https://github.com/apache/commons-net (ipv4 only)
- https://github.com/edazdarevic/CIDRUtils/blob/master/CIDRUtils.java (MIT license)

Let me know what you think and what the general process is to discuss new dependencies

YANG-DB:
Hi @salyh 
I agree - let's use https://github.com/seancfoley/IPAddress library since it's in use in our SQL repo already

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/expression/Cidr.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+public class Cidr extends UnresolvedExpression {
+    private UnresolvedExpression ipAddress;
+    private UnresolvedExpression cidrBlock;
+
+    public Cidr(UnresolvedExpression ipAddress, UnresolvedExpression cidrBlock) {
+        this.ipAddress = ipAddress;
+        this.cidrBlock = cidrBlock;
```
LantaoJin, line 19: Please merge the latest code base and refactor this class with Lombok annotations. Ref https://github.com/opensearch-project/opensearch-spark/commit/38ca314034021c34765713c016bcf3cef775efab

LantaoJin:
Hi @salyh 
Can you see if the [Guava library](https://guava.dev/releases/19.0/api/docs/com/google/common/net/InetAddresses.html) can meet our needs?

salyh:
> Hi @salyh Can you check if the [Guava library](https://guava.dev/releases/19.0/api/docs/com/google/common/net/InetAddresses.html) can meet our needs?

I checked but it does not support cidr checks. https://github.com/seancfoley/IPAddress  is also already used in the sql project.

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/expression/Cidr.java
salyh: done

salyh:
> Hi @salyh I agree - let's use https://github.com/seancfoley/IPAddress library since it's in use in our SQL repo already

done

YANG-DB:
@salyh plz resolve conflict
thanks

salyh:
> @salyh plz resolve conflict thanks

done

salyh:
@YANG-DB @vamshin @anirudha

Regarding validation of the CIDR function parameters:

The [IPAddress](https://github.com/seancfoley/IPAddress) library is very lenient regarding validation of ip addresss and cidr blocks.

I suggest we at least disallow partial ip address segements and empty or null adresses:

```
  IPAddressStringParameters valOptions = new IPAddressStringParameters.Builder()
                .allowEmpty(false)
                .setEmptyAsLoopback(false)
                .allow_inet_aton(false)
                .allowSingleSegment(false)
                .toParams();
```

Then the remaining question is wether we should enforce other syntax checks like which the library is not enforcing:

- check that cidr block is a really a in cicdr notation (containing a '/')
- check that ip address and cidr are either both ipv6 or both ipv4
- define the behaviour for edge cases like 0.0.0.0 (or just rely on the api)
- define handling of special ip adresses like multicast, anycast, 255.*.*.* etc

My suggestion is to only 

- disallow partial ip address segements and empty or null adresses (like already implemented) and leave everything else 
- disallow mixed ipv4/ipv6 adresses and cidr blocks

and leave everything else to the library.

YANG-DB:
@salyh 
please add the needed documentation in the following locations:
 - [ppl commands list](https://github.com/opensearch-project/opensearch-spark/tree/main/docs/ppl-lang)
 - [ppl functions list](https://github.com/opensearch-project/opensearch-spark/tree/main/docs/ppl-lang/functions)
 - [ppl examples doc](https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/PPL-Example-Commands.md)

YANG-DB:
> @YANG-DB @vamshin @anirudha
> 
> Regarding validation of the CIDR function parameters:
> 
> The [IPAddress](https://github.com/seancfoley/IPAddress) library is very lenient regarding validation of ip addresss and cidr blocks.
> 
> I suggest we at least disallow partial ip address segements and empty or null adresses:
> 
> ```
>   IPAddressStringParameters valOptions = new IPAddressStringParameters.Builder()
>                 .allowEmpty(false)
>                 .setEmptyAsLoopback(false)
>                 .allow_inet_aton(false)
>                 .allowSingleSegment(false)
>                 .toParams();
> ```
> 
> Then the remaining question is wether we should enforce other syntax checks like which the library is not enforcing:
> 
> * check that cidr block is a really a in cicdr notation (containing a '/')
> * check that ip address and cidr are either both ipv6 or both ipv4
> * define the behaviour for edge cases like 0.0.0.0 (or just rely on the api)
> * define handling of special ip adresses like multicast, anycast, 255._._.* etc
> 
> My suggestion is to only
> 
> * disallow partial ip address segements and empty or null adresses (like already implemented) and leave everything else
> * disallow mixed ipv4/ipv6 adresses and cidr blocks
> 
> and leave everything else to the library.

@salyh 
Thanks for this review - I support your suggestion
- disallow partial ip address segements and empty or null adresses (like already implemented) and leave everything else
- disallow mixed ipv4/ipv6 adresses and cidr blocks

In addition plz document the additional validations as future work:

```
check that cidr block is a really a in cicdr notation (containing a '/')
check that ip address and cidr are either both ipv6 or both ipv4
define the behaviour for edge cases like 0.0.0.0 (or just rely on the api)
define handling of special ip adresses like multicast, anycast, 255...* etc
```
Thanks

YANG-DB (commented):
please add the needed documentation in the following locations:

[ppl planning list](https://github.com/opensearch-project/opensearch-spark/tree/main/docs/ppl-lang/planning)

Once we approve we'll also add it onto the other docs:
 - https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/PPL-Example-Commands.md
 - https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/README.md

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
```diff
@@ -0,0 +1,63 @@
...
+                .allowEmpty(false)
+                .setEmptyAsLoopback(false)
+                .allow_inet_aton(false)
+                .allowSingleSegment(false)
+                .toParams();
+
+        @Override
+        public Boolean apply(String ipAddress, String cidrBlock) {
+
+            IPAddressString parsedIpAddress = new IPAddressString(ipAddress, valOptions);
```
penghuo, line 36: Should we leverage [InetAddresses](https://github.com/opensearch-project/OpenSearch/blob/main/libs/common/src/main/java/org/opensearch/common/network/InetAddresses.java) in opensearch-common package to parse IpAddress and CIDR?

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
```diff
@@ -0,0 +1,63 @@
...
+            try {
+                parsedIpAddress.validate();
+            } catch (AddressStringException e) {
+                throw new RuntimeException("The given ipAddress '"+ipAddress+"' is invalid. It must be a valid IPv4 or IPv6 address. Error details: "+e.getMessage());
+            }
+
+            IPAddressString parsedCidrBlock = new IPAddressString(cidrBlock, valOptions);
+
+            try {
+                parsedCidrBlock.validate();
```
penghuo, line 47: Could we align the validation with OpenSearch?

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -476,6 +477,10 @@ booleanFunctionCall
    : conditionFunctionBase LT_PRTHS functionArgs RT_PRTHS
    ;
 
+cidrFunctionCall
+   : CIDR LT_PRTHS ipAddress = functionArg COMMA cidrBlock = functionArg RT_PRTHS
```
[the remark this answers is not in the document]
dr-lilienthal: You mean instead of “`where CIDR(src_ip, ‘192.168.0.0/24’)`” that the command “`where CIDRMATCH(src_ip, ‘192.168.0.0/24’)`” is better understandable? Should we change this accordingly?

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
dr-lilienthal: I don't think there's anything wrong with it. It is a well-maintained package

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
penghuo: yes, should we use cidrmatch instead of cidr as function name.

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
penghuo: I am not concerning the package is well-maintained or not.
My point is to be consistent with OpenSearch's parsing and validation behavior.

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
YANG-DB: @dr-lilienthal I agree with @penghuo 
lets consolidate for using the same functionality
thanks

YANG-DB:
@dr-lilienthal can u plz resolve the conflicts and also signOff the commits ?( DCO failure )
thanks

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
salyh: When I read this and [the above thread](https://github.com/opensearch-project/opensearch-spark/pull/706#discussion_r1797101977) correct the idea is to remove the https://github.com/seancfoley/IPAddress lib dependency and instead use classes of the `org.opensearch.common.network`, right?

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
salyh: @penghuo Can you provide a link to code or docs where we can see the `OpenSearch's parsing and validation behavior.`?

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
salyh: @dr-lilienthal can you rename cidr to cidrmatch?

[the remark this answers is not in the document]
dr-lilienthal: @salyh yes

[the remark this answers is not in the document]
dr-lilienthal: done @salyh

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
penghuo: https://github.com/opensearch-project/OpenSearch/blob/main/libs/common/src/main/java/org/opensearch/common/network/InetAddresses.java#L358
```
    public static InetAddress forString(String ipString) {
        byte[] addr = ipStringToBytes(ipString);

        // The argument was malformed, i.e. not an IP string literal.
        if (addr == null) {
            throw new IllegalArgumentException(String.format(Locale.ROOT, "'%s' is not an IP string literal.", ipString));
        }

        return bytesToInetAddress(addr);
    }
```

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
penghuo: Exactally.

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
salyh: @penghuo Thanks. But please see https://github.com/opensearch-project/opensearch-spark/pull/706#discussion_r1800818727

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
YANG-DB: @penghuo please assist...

penghuo: @salyh for cidr match, I want to confirm if the behavior of the term query is as expected. If it is, I can help locate the relevant code.
```
GET my-index-000001/_search
{
  "query": {
    "term": {
      "ip_addr": "192.168.0.0/16"
    }
  }
}
```

salyh: @penghuo yes it is. Is the relevant code this?: https://github.com/opensearch-project/OpenSearch/blob/119abaff95a7f54affea6c844fad7ce3c8360155/server/src/main/java/org/opensearch/index/mapper/IpFieldMapper.java#L248

If so I try to re-implement based on `org.apache.lucene.document.InetAddressPoint.newPrefixQuery()` which evalutes to something like 

```
byte[] lower = value.getAddress();
    byte[] upper = value.getAddress();
    for (int i = prefixLength; i < 8 * lower.length; i++) {
      int m = 1 << (7 - (i & 7));
      lower[i >> 3] &= ~m;
      upper[i >> 3] |= m;
    }
```

penghuo: IpField term query is rewritten as PointRangeQuery in lucene. I can think 2 approach
1. read value from child operator, index into Lucene, and run range query.
2. leverage [parseCidr logic](https://github.com/opensearch-project/OpenSearch/blob/main/libs/common/src/main/java/org/opensearch/common/network/InetAddresses.java#L395), and perform match query using https://github.com/seancfoley/IPAddress.

salyh: @penghuo @YANG-DB 
I would vote for 2. because 1. is too complex for almost no benefit
But I cannot access `org.opensearch.common.network.InetAddresses` class because its seems not part of the dependencies. While trying to add 'opensearch-commons' lib as dependency i noticed that in build.sbt 

```
lazy val opensearchVersion = "2.6.0"
lazy val opensearchMavenVersion = "2.6.0.0"
```

is set whereas a) the org.opensearch.common.network.InetAddresses was not part of the libarary in this version and b) it seems odd to compile and build against 2.6.0 which is very old

penghuo: what is the difference of `parsedCidrBlock.validate();` vs OpenSearch implementation? can we leverage `parsedCidrBlock.validate();` now and upgrade later?

salyh: sure, thats the easiest way and already implemented. I add tests and check what the differences are ...

ppl-spark-integration/src/main/antlr4/OpenSearchPPLLexer.g4
```diff
@@ -396,6 +396,7 @@ LIKE:                               'LIKE';
 ISNULL:                             'ISNULL';
 ISNOTNULL:                          'ISNOTNULL';
 ISPRESENT:                          'ISPRESENT';
+CIDRMATCH:                          'CIDRMATCH';
```
YANG-DB, line 399: Add CIDRMATCH to `keywordsCanBeId`

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
```diff
@@ -0,0 +1,63 @@
...
+
+import inet.ipaddr.AddressStringException;
+import inet.ipaddr.IPAddressString;
+import inet.ipaddr.IPAddressStringParameters;
+import scala.Function2;
+import scala.Serializable;
+import scala.runtime.AbstractFunction2;
+
+
+public class SerializableUdf {
```
YANG-DB, line 16: why not use an interface or abstract class in that case  ?

YANG-DB (changes requested):

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
salyh: Thats imho a pure style discussion. I prefer private constructors because abstract classes are intended for subclassing and interfaces for multiple inheritance. But I am happy to convert it into a abstract class or an interface.

ppl-spark-integration/src/main/antlr4/OpenSearchPPLLexer.g4
salyh: done

ppl-spark-integration/src/main/java/org/opensearch/sql/expression/function/SerializableUdf.java
salyh: Changed to an interface

salyh:
> @salyh @dr-lilienthal can u plz resolve the conflicts ? thanks

@YANG-DB done

docs/ppl-lang/functions/ppl-ip.md
```diff
@@ -0,0 +1,35 @@
...
+    os> source=ipsv6 | where cidrmatch(ip, '2003:db8::/32') | fields ip
+    fetched rows / total rows = 1/1
+    +-----------------------------------------+
+    | ip                                      |
+    |-----------------------------------------|
+    | 2003:0db8:0000:0000:0000:0000:0000:0000 |
+    +-----------------------------------------+
+
+Note:
+ - `ip` can be an IPv4 or an IPv6 address
```
YANG-DB, line 32: @salyh please explain 'ip' is a column name and 'cidr' is a literal value

salyh: according to antlr definition its both a `functionArg` (which makes sense imho)

YANG-DB (approved):

Issue #672, opened 2024-09-16
[FEATURE]Add `iplocation` function to PPL for IP address geolocation


YANG-DB:
**Description:**
We propose adding an `geoip` function to OpenSearch's Piped Processing Language (PPL) and SQL to provide built-in IP address geolocation capabilities.
This feature would be similar to functionality used in OpenSearch's geospatial feature, enhancing PPL's ability to enrich log data with geographical information based on IP addresses.

**Proposed Functionality:**

1. The 'geoip' function should take an IP address as input and return geographical information.
2. It should support both IPv4 and IPv6 addresses.
3. The function should return multiple fields including country, region, city, latitude, longitude, and others as available.
4. It should allow users to specify which geolocation fields to include in the output.
5. The function should use a regularly updated IP geolocation database for accuracy.

**Example Usage:**
```sql
... | eval geolocation = geoip(ip_field)
```

This would add a new field 'geolocation' with all available location information for the IP address in 'ip_field'.

```sql
... | eval country = geoip(ip_field, "country")
... | eval lat = geoip(ip_field, "lat"), lon = iplocation(ip_field, "lon")
```

This would add new fields with specific geolocation information.

```sql
... | eval location_info = geoip(ip_field, "country,region,city,lat,lon")
```

This would add a new field 'location_info' with multiple pieces of geolocation data.

**Additional considerations**
- Allow for registering a DB table that allows resolving the IP to Geo
- Adding a generic way to register the IP to Geo location resolving mechanism / service
    -  Adding auth tokens for calling such service

dblock:
[Catch All [Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta),  [3](https://github.com/krisfreedain),  [4](https://github.com/lukas-vlcek/)]

abdullahdevrel:
Just passing by - considering the discussion is around GeoIP functionality, would it be possible to add support for our ([IPinfo](https://ipinfo.io/)) IP databases as well? We have a bunch of IP databases that come in MMDB data file format, which also include our free IP database (https://ipinfo.io/products/free-ip-database). Our free IP databases include the [IP to Country ASN/ISP database](https://ipinfo.io/developers/ip-to-country-asn-database) that comes with full accuracy, no EULA, and allows commercial re-distribution through only providing an attribute.

YANG-DB:
[...]

PR #781, opened 2024-10-16
PPL geoip function - old reference PR

salyh:
### Description
PPL geoip function

### Issues Resolved
https://github.com/opensearch-project/opensearch-spark/issues/672

### DIFF
--- a/docs/ppl-lang/planning/ppl-geoip.md
+++ b/docs/ppl-lang/planning/ppl-geoip.md
@@ -1,0 +1,39 @@
+## geoip syntax proposal
+
+geoip function to add information about the geographical location of an IPv4 or IPv6 address
+
+1. **Proposed syntax**
+    - `... | eval geoinfo = geoip([datasource,] ipAddress [,properties])`
+        - generic syntax     
+    - `... | eval geoinfo = geoip(ipAddress)`
+        - use the default geoip datasource
+    - `... | eval geoinfo = geoip("abc", ipAddress)`
+        - use the "abc" geoip datasource
+    - `... | eval geoinfo = geoip(ipAddress, "city,lat,lon")`
+        -  use the default geoip datasource, retrieve only city, lat and lon
+    - `... | eval geoinfo = geoip("abc", ipAddress, "city,lat,lon")`
+        - use the "abc" geoip datasource, retrieve only city, lat and lon
+
+
+2. **Proposed wiring with the geoip database**
+    - Leverage the functionality of the ip2geo processor
+      - ip2geo processor configuration, functionality and code will be used
+      - Prerequisite for the geoip is that ip2geo processor is configured properly
+      - See https://opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/
+
+
+### New syntax definition in ANTLR
+
+```ANTLR
+  
+// functions
+evalFunctionCall
+   : evalFunctionName LT_PRTHS functionArgs RT_PRTHS
+   | geoipFunction
+   ;  
+  
+geoipFunction
+   : GEOIP LT_PRTHS (datasource = functionArg COMMA)? ipAddress = functionArg (COMMA properties = stringLiteral)? RT_PRTHS
+   ;
+```
+
--- a/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
+++ b/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
@@ -465,6 +465,7 @@ wcFieldExpression
 // functions
 evalFunctionCall
    : evalFunctionName LT_PRTHS functionArgs RT_PRTHS
+   | geoipFunction
    ;
 
 // cast function
@@ -781,6 +782,10 @@ coalesceFunctionName
    : COALESCE
    ;
 
+geoipFunction
+   : GEOIP LT_PRTHS (datasource = functionArg COMMA)? ipAddress = functionArg (COMMA properties = stringLiteral)? RT_PRTHS
+   ;
+
 // operators
  comparisonOperator
    : EQUAL
@@ -1051,4 +1056,5 @@ keywordsCanBeId
    | FULL
    | SEMI
    | ANTI
+   | GEOIP
    ;
--- a/ppl-spark-integration/src/main/antlr4/OpenSearchPPLLexer.g4
+++ b/ppl-spark-integration/src/main/antlr4/OpenSearchPPLLexer.g4
@@ -372,6 +372,9 @@ TYPEOF:                             'TYPEOF';
 //OTHER CONDITIONAL EXPRESSIONS
 COALESCE:                           'COALESCE';
 
+//GEOLOCATION FUNCTIONS
+GEOIP:                              'GEOIP';
+
 // RELEVANCE FUNCTIONS AND PARAMETERS
 MATCH:                              'MATCH';
 MATCH_PHRASE:                       'MATCH_PHRASE';


salyh:
@YANG-DB @penghuo @lukasz-soszynski-eliatra @kt-eliatra @dr-lilienthal PPL geoip function syntax proposal

Please review and comment if needed

YANG-DB:
@LantaoJin @penghuo @dai-chen please add u'r comments here

salyh:
> @salyh thanks on my end looking good - only question is the `datasource` definition - how does the user creates a new `datasource, can/should the `datasource` be one of the following:
> 
>     * external API
> 
>     * table
> 
>     * file
>       Lets add these capabilities ...

@YANG-DB @LantaoJin @penghuo @dai-chen 

datasource refers to the datasources provided by the ip2geo processor https://opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/

My idea was to leveage the already present capabilities of the [ip2geo processor](https://opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/) and extend it as necessary. Because it would avoid code duplication. When I read the docs of the processor correct, GeoLite2 is supported by now. I propose to add the above requested functionality to the processor. Then we try to use the code from the processor (as a library) to solve this issue. Otherwise we would reinvent the wheel.

docs/ppl-lang/planning/ppl-geoip.md
```diff
@@ -0,0 +1,39 @@
...
+
+
+2. **Proposed wiring with the geoip database**
+    - Leverage the functionality of the ip2geo processor
+      - ip2geo processor configuration, functionality and code will be used
+      - Prerequisite for the geoip is that ip2geo processor is configured properly
+      - See https://opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/
+
+
+### New syntax definition in ANTLR
```
LantaoJin, line 25: As a user doc, could we remove the ANTLR related information, what here needs is just a user readable syntax definition.

LantaoJin: Oh, the file is located ppl-lang/planning folder. I think we need an user doc instead of design doc. We can copy this design information to the PR description or issue description. And you can easily update your design without any code changes.

docs/ppl-lang/planning/ppl-geoip.md
```diff
@@ -0,0 +1,39 @@
+## geoip syntax proposal
+
+geoip function to add information about the geographical location of an IPv4 or IPv6 address
+
+1. **Proposed syntax**
+    - `... | eval geoinfo = geoip([datasource,] ipAddress [,properties])`
```
LantaoJin, line 6: what does `datasource` mean? Could you use other meaningful usage rather that `abc`? `abc` is not straightforward user case.

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -781,6 +782,10 @@ coalesceFunctionName
    : COALESCE
    ;
 
+geoipFunction
+   : GEOIP LT_PRTHS (datasource = functionArg COMMA)? ipAddress = functionArg (COMMA properties = stringLiteral)? RT_PRTHS
```
LantaoJin, line 786: > datasource = functionArg

Any case for supporting `functionArg` instead of `stringLiteral`. Again, what does `datasource` mean?

docs/ppl-lang/planning/ppl-geoip.md
salyh: datasource refers to the datasources provided by the ip2geo processor, see https://github.com/opensearch-project/opensearch-spark/pull/781#issuecomment-2420675538

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
salyh: no particular case for `functionArg`. We can change it to `stringLiteral` for sure

datasource refers to the datasources provided by the ip2geo processor, see https://github.com/opensearch-project/opensearch-spark/pull/781#issuecomment-2420675538

docs/ppl-lang/planning/ppl-geoip.md
salyh: @YANG-DB I did unterstand that the `ppl-lang/planning` folder is for design docs. When the design is approved it will be converted to a userdoc and moved out of the `planning` sub-folder. Am I wrong with that assumption?

LantaoJin: I thought it was for roadmap. I prefer to using GitHub PR description and comments. It's fine to add design docs in code if you prefer to. But these docs are not for users, we need to write a new one later.

YANG-DB: @salyh yes - once design doc is approved we will move it with necessary updates to the regular commands folder

salyh:
> HI @salyh,
> 
> Currently looking into implementation if this feature. I saw above conversation on planning to leverage ip2geo functionality within geospatial plugin. Did you have a clear idea on how this can be done?
> 
> From my understanding to leverage ip2geo, one must first create a ip2geo processor then attach it to a opensearch pipeline. I am not too sure how the Spark plugin could access this functionality. Some pointers would be much appreciated :)
> 
> Thanks!

The idea is not to leverage the processor as a processor but to reuse its code (either by copying it - not ideal - , or by separating them into a kind of shared commons library or by adding them as a dependency on code level).

kenrickyap:
Sounds good my current idea is to create an spi for the plugin, as was done for job-scheduler (allowing the geospatial plugin to implement ExtensiblePlugin) am still working on a design. This is to allow the sql plugin to access the ip2geo functionality (am also working on the iplocation functionality there).

Would you know if the Spark project would be able to leverage the SPI since it is not a plugin? If we are to expose the ip2geo functionality I would rather create something that works both for the sql-plugin and Spark.

kenrickyap:
Hi @salyh, spent a bit more time investigating this ticket. currently plan is to rewrite the geo2ip functionality within the spark project. The reason for this is that currently the geo2ip processor with the geospatial plugin is intwined with using the job scheduler extension. From my understanding there is not way to leverage usage of this extension in spark so a lot of the functionality would not be able to be migrated.

The rewritten geo2ip within spark will:
- create spark temp table storing data from geoip datasource if table does not already exist
- retrieve information from said table based on provided ips.

The current design question is if it would be better to use spark UDF or reflection to call [java_method](https://spark.apache.org/docs/3.5.3/api/sql/index.html#java_method) to call the rewritten geo2ip functionality?

kenrickyap:
> Hi @salyh, spent a bit more time investigating this ticket. currently plan is to rewrite the geo2ip functionality within the spark project. The reason for this is that currently the geo2ip processor with the geospatial plugin is intwined with using the job scheduler extension. From my understanding there is no way to leverage usage of this extension in spark so a lot of the functionality would not be able to be migrated.
> 
> The rewritten geo2ip within spark will:
> 
> * create spark temp table storing data from geoip datasource if table does not already exist
> * retrieve information from said table based on provided ips.
> 
> The current design question is if it would be better to use spark UDF or reflection to call [java_method](https://spark.apache.org/docs/3.5.3/api/sql/index.html#java_method) to call the rewritten geo2ip functionality?

Will follow CIDR https://github.com/opensearch-project/opensearch-spark/pull/706/files approach and use UDF

YANG-DB (changes requested):
Only a reference PR - DO NO MERGE !

Issue #740, opened 2024-10-03
[BUG][SanityTest] stats `by` a high cardinality field will cause writing job fail with "size exceed limitation"


LantaoJin:
**What is the bug?**
When group by a high cardinality field, such as `clientip`, the spark writing job will fail "Request size exceeded 10485760 bytes" .

<img width="1634" alt="Screenshot 2024-10-01 at 17 33 09" src="https://github.com/user-attachments/assets/d15cd8c1-1f88-4a5a-868d-191c60005e11">
<img width="1651" alt="Screenshot 2024-10-01 at 15 12 07" src="https://github.com/user-attachments/assets/427b89c4-8d25-4c0d-bf53-183a630b70c4">

**How can one reproduce the bug?**
1. Go to Workbench page. For example https://search-flint05-sanity-h5if7yelmxws5hc35cap2lauf4.us-east-1.es-integ.amazonaws.com/_dashboards/app/opensearch-query-workbench#/
2. Execute a PPL query `source = myglue_test.default.http_logs | stats avg(size) by clientip`
3. Open the EMR-S job in Spark UI.

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

YANG-DB:
@LantaoJin what do you suggest here ? 
should we rewrite the query in such high carnality cases ?

YANG-DB:
[fieldsummary](https://github.com/opensearch-project/opensearch-spark/issues/662) should provide statistical info about fields and assisting the cost estimation of the query including cardinality

PR #796, opened 2024-10-21
PPL `tablesample`  command for faster approximation statements

YANG-DB:
### Description
Add a new sample command (`tablesample`) to reduce amount of scanned data points and allow approximation of a statement when faster sample based results if favour of exact long running results 

```sql
source = testTable TABLESAMPLE(50 percent) | rare address
```

### Issues Resolved
- https://github.com/opensearch-project/opensearch-spark/issues/740


YANG-DB:
@LantaoJin @penghuo I'd like u'r comments on this direction for approximation based queries...

YANG-DB:
The following example demonstrates how to sample 50% of the data from the table and then perform aggregation (finding rare occurrences of address).

PPL query:

    os> source = account  TABLESAMPLE(75 percent) | top 3 country by occupation

This query samples 75% of the records from account table, then retrieves the top 3 countries grouped by occupation

```sql
SELECT *
FROM (
         SELECT country, occupation, COUNT(country) AS count_country
         FROM account
                  TABLESAMPLE(75 PERCENT)
         GROUP BY country, occupation
         ORDER BY COUNT(country) DESC NULLS LAST
             LIMIT 3
     ) AS subquery
    LIMIT 3;
```
Logical Plan Equivalent:

```sql
'Project [*]
+- 'GlobalLimit 3
   +- 'LocalLimit 3
      +- 'Sort ['COUNT('country) AS count_country#68 DESC NULLS LAST], true
         +- 'Aggregate ['country, 'occupation AS occupation#67], ['COUNT('country) AS count_country#66, 'country, 'occupation AS occupation#67]
            +- 'Sample 0.0, 0.75, false, 0
               +- 'UnresolvedRelation [account], [], false

```

By introducing the `TABLESAMPLE` instruction into the source command, one can now sample data as part of your queries and reducing the amount of data being scanned thereby converting precision with performance.

The `percent` parameter will give the actual approximation of the true value with the needed trade of between accuracy and performance.

YANG-DB:
The following example demonstrates how to sample 50% of the data from the table and then perform aggregation (finding rare occurrences of address).

PPL query:

    os> source = account  TABLESAMPLE(75 percent) | top 3 country by occupation

This query samples 75% of the records from account table, then retrieves the top 3 countries grouped by occupation

```sql
SELECT *
FROM (
         SELECT country, occupation, COUNT(country) AS count_country
         FROM account
                  TABLESAMPLE(75 PERCENT)
         GROUP BY country, occupation
         ORDER BY COUNT(country) DESC NULLS LAST
             LIMIT 3
     ) AS subquery
    LIMIT 3;
```
Logical Plan Equivalent:

```sql
'Project [*]
+- 'GlobalLimit 3
   +- 'LocalLimit 3
      +- 'Sort ['COUNT('country) AS count_country#68 DESC NULLS LAST], true
         +- 'Aggregate ['country, 'occupation AS occupation#67], ['COUNT('country) AS count_country#66, 'country, 'occupation AS occupation#67]
            +- 'Sample 0.0, 0.75, false, 0
               +- 'UnresolvedRelation [account], [], false

```

By introducing the `TABLESAMPLE` instruction into the source command, one can now sample data as part of your queries and reducing the amount of data being scanned thereby converting precision with performance.

The `percent` parameter will give the actual approximation of the true value with the needed trade of between accuracy and performance.

LantaoJin:
high level question, will below query work as expected?
> os> source = account1 TABLESAMPLE(75 percent), account2 TABLESAMPLE(10 percent) | top 3 country by occupation

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -160,11 +160,18 @@ public LogicalPlan visitRelation(Relation node, CatalystPlanContext context) {
                             true,
                             DescribeRelation$.MODULE$.getOutputAttrs()));
         }
+        //populate table sampling
+        context.withSampling(node.getTablesampleContext());
```
LantaoJin, line 164: should this line move after L169?

LantaoJin:
For long-term thinking, how above just rename `TABLESAMPLE` to `SAMPLE`, seems sample could be applied not only table but also any plan node. 
And it just brings me why not as a new command? For example, `source = t status=200 | sample 50 percent | ...`
has very different semantic with `source = t sample(50 percent) status=200 | ...`

LantaoJin:
Is `Sample` a non-deterministic operator? It may prevent DSL pushdown in sql. Not sure how it works with Spark Pushdown optimizations such as FilterPushdown or AggregatePushdown, etc. So it might be a performance barrier which is out of user's expectation. We need to confirm and highlight it in doc if yes.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
YANG-DB: good point - I'll refactor this part

YANG-DB:
> For long-term thinking, how above just rename `TABLESAMPLE` to `SAMPLE`, seems sample could be applied not only table but also any plan node. And it just brings me why not as a new command? For example, `source = t status=200 | sample 50 percent | ...` has very different semantic with `source = t sample(50 percent) status=200 | ...`

I agree about the renaming but I think it should be attached to the index since its clear the sampling is the first action and has no prior operation since the pipe operation has no specific restriction and the sample operation needs to very clear it operates directly on the index/table

YANG-DB:
closing this as [this PR](https://github.com/opensearch-project/opensearch-spark/pull/879) give a better more precise answer to the issue at hand

PR #817, opened 2024-10-25
Add checkpoint.delete.processingTime metric

ykmr1224:
### Description
- Add checkpoint.delete.processingTime metric

Tested in my test cluster
checkpoint delete processingTime

### Related Issues
n/a

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricsUtil.java
+++ b/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricsUtil.java
@@ -9,6 +9,7 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import java.util.function.Supplier;
 import org.apache.spark.SparkEnv;
 import org.apache.spark.metrics.source.FlintMetricSource;
 import org.apache.spark.metrics.source.FlintIndexMetricSource;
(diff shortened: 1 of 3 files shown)


flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricsUtil.java
```diff
@@ -133,6 +134,21 @@ public static void addHistoricGauge(String metricName, final long value) {
         }
     }
 
+    /**
+     * Automatically emit latency metric as Historic Gauge for the execution of supplier
+     * @param supplier the lambda to be metered
+     * @param metricName name of the metric
+     * @return value returned by supplier
+     */
+    public static <T> T withLatencyAsHistoricGauge(Supplier<T> supplier, String metricName) {
```
seankao-az, line 143: This might be too nit-picky, but do you think it might be better to swap the order of arguments supplier and metricName?

In case we need to measure latency for larger code blocks, using it will look like:
```
withLatencyAsHistoricGauge(() => {
  // some
  // larger
  // code
  // block
}, MetricConstants.SOME_OPERATION_METRIC)
```

Personally I think this is more natural to use:
```
withLatencyAsHistoricGauge(MetricConstants.SOME_OPERATION_METRIC, () => {
  // some
  // larger
  // code
  // block
})
```

This usage would look more similar to scala syntax if higher order function can be used, which allow for
```
withLatencyAsHistoricGauge(MetricConstants.SOME_OPERATION_METRIC) {
  // some
  // operation
  // code
}
```

seankao-az (approved):
approve with non-blocking minor comment

flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricsUtil.java
ykmr1224: Thanks for the comment.
I changed the order to current one since I saw possibility to extend it to take variable args for multiple metricName to be emitted. (I did it for OpenSearch client metrics)

PR #825, opened 2024-10-28
Add index state metrics

ykmr1224:
### Description
- Add index state metrics
  - `indexState. updatedTo.{state}.count`: count for the index's final log state. emitted after optimistic transaction when it is successful.
  - `initialConditionCheck.failed.{state}.count`: count for initial condition check failure per index state.

Tested with creating auto refresh MV, and trying manual refresh:
indexState metrics

### Related Issues
n/a

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
+++ b/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
@@ -161,9 +161,9 @@ public final class MetricConstants {
     public static final String CHECKPOINT_DELETE_TIME_METRIC = "checkpoint.delete.processingTime";
 
     /**
-     * Metric for tracking the index state transitions
+     * Metric prefix for tracking the index state transitions
      */
-    public static final String INDEX_STATE_METRIC_PREFIX = "indexState.";
+    public static final String INDEX_STATE_UPDATED_TO_PREFIX = "indexState.updatedTo.";
 
     /**
      * Metric for tracking the index state transitions
--- a/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
+++ b/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
@@ -160,6 +160,16 @@ public final class MetricConstants {
      */
     public static final String CHECKPOINT_DELETE_TIME_METRIC = "checkpoint.delete.processingTime";
 
+    /**
+     * Metric for tracking the index state transitions
+     */
+    public static final String INDEX_STATE_METRIC_PREFIX = "indexState.";
+
+    /**
+     * Metric for tracking the index state transitions
+     */
+    public static final String INITIAL_CONDITION_CHECK_FAILED_PREFIX = "initialConditionCheck.failed.";
+
     private MetricConstants() {
         // Private constructor to prevent instantiation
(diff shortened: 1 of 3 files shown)


flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
```diff
@@ -160,6 +160,16 @@ public final class MetricConstants {
      */
     public static final String CHECKPOINT_DELETE_TIME_METRIC = "checkpoint.delete.processingTime";
 
+    /**
+     * Metric for tracking the index state transitions
+     */
+    public static final String INDEX_STATE_METRIC_PREFIX = "indexState.";
```
noCharger, line 166: what's the expection of this one? let's say there's a transcation call on single index every 1 minute, then sum of 5 mins metrics will goes to 5?

noCharger (commented):
Also can you explain the screenshot - why does "refreshing" has continuous value?

flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
ykmr1224: Yes, it will emit datapoint for each update, and the sum would be 5 in that case.
Refreshing has continuous value, since it was refreshing every minute.

noCharger: > Yes, it will emit datapoint for each update, and the sum would be 5 in that case.
but there's only one index, does it make sense to have 5?

> Refreshing has continuous value, since it was refreshing every minute.
if you are refering to the heartbeat, there's already metric on it

ykmr1224: As it hooks status update by the `finalLog`, it emits data point for each OptimisticTransaction.
Our main interest is active/failed state.
Basically I don't want to make complex rule/logic for emitting metrics.
If you have better idea, please let me know.

seankao-az: I guess it's not clear what `indexState.{state}.count` should mean. I would interpret this as counting the number of index in each state at any given time. But updating in final log will bloat the refresh state index count?

noCharger (approved):

flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
ykmr1224: Fixed the metric name to `indexState.updatedTo.{state}.count`

dai-chen (approved):
Any UT to verify the metric counter? Also I forgot if we document all these metrics somewhere?

ykmr1224:
> Any UT to verify the metric counter? Also I forgot if we document all these metrics somewhere?

Added UT for the metrics for this class.
There is no separate doc for metrics, but I think `MetricConstants` class and its Javadoc works as dox&index.

noCharger:
Will HistoricGauge with value one aggregated on client side?

ykmr1224:
> Will HistoricGauge with value one aggregated on client side?

It will be aggregated at CloudWatch side.

Issue #804, opened 2024-10-23
[BUG] Infinite query loop when result metadata in unavailable


noCharger:
**What is the bug?**
Infinite query loop when result metadata in unavailable
 
**How can one reproduce the bug?**
Try to reject result metadata access.

**What is the expected behavior?**
Query loop should exit

noCharger:
RCA:
1. Any unhandled exception from opensearch client will be considered as 500, we should handle the case where explicit status code returned
2. The `checkindex` logic is included query loop. It can be optimized bying moving it from statement level to session level since query from different data source will start a new session.

PR #828, opened 2024-10-29
Optimize prepareStatementExecution request freq

noCharger:
### Description
Optimize prepareStatementExecution request freq. 

Before the change:

- prepareStatementExecution will run k times, depends dynamically on session duration and query execution time.
- getNext will run k times to create new reader clients

After the change:

- prepareStatementExecution will run 1 time before the query loop
- getNext will behave the same, to avoid concurrent issue of using one reader client within same session 

Also tested on EMR to verify number of times.

### Related Issues
https://github.com/opensearch-project/opensearch-spark/issues/804

### Check List
- [X] Implemented unit tests
- [X] Commits are signed per the DCO using `--signoff`

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/StatementExecutionManagerImpl.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/StatementExecutionManagerImpl.scala
@@ -6,7 +6,7 @@
 package org.apache.spark.sql
 
 import org.opensearch.flint.common.model.FlintStatement
-import org.opensearch.flint.core.storage.OpenSearchUpdater
+import org.opensearch.flint.core.storage.{FlintReader, OpenSearchUpdater}
 import org.opensearch.search.sort.SortOrder
 
 import org.apache.spark.internal.Logging
@@ -29,8 +29,8 @@ class StatementExecutionManagerImpl(commandContext: CommandContext)
     context("flintSessionIndexUpdater").asInstanceOf[OpenSearchUpdater]
 
   // Using one reader client within same session will cause concurrency issue.
-  // To resolve this move the reader creation and getNextStatement method to mirco-batch level
-  private val flintReader = createOpenSearchQueryReader()
+  // To resolve this move the reader creation to getNextStatement method at mirco-batch level
+  private var currentReader: Option[FlintReader] = None
 
   override def prepareStatementExecution(): Either[String, Unit] = {
     checkAndCreateIndex(osClient, resultIndex)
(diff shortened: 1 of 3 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/StatementExecutionManagerImpl.scala
```diff
@@ -29,8 +29,8 @@ class StatementExecutionManagerImpl(commandContext: CommandContext)
     context("flintSessionIndexUpdater").asInstanceOf[OpenSearchUpdater]
 
   // Using one reader client within same session will cause concurrency issue.
-  // To resolve this move the reader creation and getNextStatement method to mirco-batch level
-  private val flintReader = createOpenSearchQueryReader()
+  // To resolve this move the reader creation to getNextStatement method at mirco-batch level
+  private var currentReader: Option[FlintReader] = None
```
dai-chen, line 33: If we create new reader for each statement, should we make this local in `getNextStatement` and close it there?

noCharger:
> Is it possible to reproduce the issue in UT or IT and verify the fix?

There's UT in https://github.com/opensearch-project/opensearch-spark/pull/828/files#diff-19a51bd4a67b635f750a1f694cf276cd220c549fa48f80e1018782b64d32ec61, are you expect something different?

spark-sql-application/src/main/scala/org/apache/spark/sql/StatementExecutionManagerImpl.scala
noCharger: > If we create new reader for each statement, should we make this local in `getNextStatement` and close it there?

it's not for each statement, but a batch of statements. FlintReader follows iteractor pattern and getNextStatement will reuse reader multiple times.

dai-chen:
> > Is it possible to reproduce the issue in UT or IT and verify the fix?
> 
> There's UT in https://github.com/opensearch-project/opensearch-spark/pull/828/files#diff-19a51bd4a67b635f750a1f694cf276cd220c549fa48f80e1018782b64d32ec61, are you expect something different?

I mean is there UT to cover this case? The existing UT passed even if there was infinity loop bug right?

dai-chen (approved):

noCharger:
> > > Is it possible to reproduce the issue in UT or IT and verify the fix?
> > 
> > 
> > There's UT in https://github.com/opensearch-project/opensearch-spark/pull/828/files#diff-19a51bd4a67b635f750a1f694cf276cd220c549fa48f80e1018782b64d32ec61, are you expect something different?
> 
> I mean is there UT to cover this case? The existing UT passed even if there was infinity loop bug right?


The existing UT expect prepareStatementExecution running k times, and we changed it to be one here

Issue #831, opened 2024-10-29
[FEATURE]PPL new `trendline` command ( WMA)


salyh:
This is the WMA part of the new spark PPL trendline command

See https://github.com/opensearch-project/opensearch-spark/issues/655 for details

salyh:
See https://github.com/eliatra/opensearch-spark/pull/7 for an prototype

salyh:
Consider https://github.com/opensearch-project/opensearch-spark/pull/833#issuecomment-2448892431 when implementing WMA

andy-k-improving:
Hi @salyh @LantaoJin 
I'm  picking this task up base on the existing SMA implementation (https://github.com/opensearch-project/opensearch-spark/pull/833), and the provided prototype (https://github.com/eliatra/opensearch-spark/pull/7)

Thanks!

salyh:
Consider https://github.com/opensearch-project/opensearch-spark/pull/784#discussion_r1824422004 when implementing WMA

andy-k-improving:
> The below is my proposed implementation on Spark SQL with `nth_value` aggregation and associated Logical plan.
> 
> ```
> 
> 
> CREATE TABLE employees (name STRING, dept STRING, salary INT, age INT, con STRING);
> 
> INSERT INTO employees VALUES ("Lisa", "Sales", 10000, 35, 'test');
> INSERT INTO employees VALUES ("Evan", "Sales", 32000, 38, 'test');
> INSERT INTO employees VALUES ("Fred", "Engineering", 21000, 28, 'test');
> INSERT INTO employees VALUES ("Alex", "Sales", 30000, 33, 'test');
> INSERT INTO employees VALUES ("Tom", "Engineering", 23000, 33, 'test');
> INSERT INTO employees VALUES ("Jane", "Marketing", 29000, 28, 'test');
> INSERT INTO employees VALUES ("Jeff", "Marketing", 35000, 38, 'test');
> INSERT INTO employees VALUES ("Paul", "Engineering", 29000, 23, 'test');
> INSERT INTO employees VALUES ("Chloe", "Engineering", 23000, 25, 'test');
> 
> 
> explain extended SELECT name, salary, 
> (nth_value(salary, 3) OVER w *3 + nth_value(salary, 2) OVER w *2 + nth_value(salary, 1) OVER w *1)/6 AS WMA
> 
> FROM employees 
[...]

Issue #655, opened 2024-09-12
[FEATURE]PPL new trendline command (SMA)


YANG-DB:
**Is your feature request related to a problem?**

Adding a new PPL `trendline` command to support computing a moving averages of fields.

We would like to support two flavours of moving average:

 ****SMA** : Simple moving average**

- f[i]: The value of field 'f' in the i-th data-point
- n: The number of data-points in the moving window (period)
- t: The current time index

`SMA(t) = (1/n) * Σ(f[i]), where i = t-n+1 to t`

---

****WMA** :  Weighted moving average**

`WMA(t) = Σ(w[i] * f[i]) / Σ(w[i]), where i = t-n+1 to t`
Where w[i] is the weight for the i-th data-point. 

In a typical WMA, the weights are linearly decreasing from the most recent to the oldest data-point:
`w[i] = n - (t - i), where i = t-n+1 to t`

The complete forumlation would be:
`WMA(t) = Σ((n - (t - i)) * f[i]) / Σ(n - (t - i)), where i = t-n+1 to t`

---

**Example**

The next command shows a trendline over a 5 month period events by month
```sql
source=t | stats count(date_month) | trendline sma(5, count) AS trend | fields  trend
```
The next command would compute a 5-point simple moving average of the 'cpu_usage' field and store it in a new field called 'smooth_cpu'.
```sql
source=t| trendline sma(5,cpu_usage) as smooth_cpu
```

Multiple trendlines could be calculated in a single command, such as 
```sql
| trendline sma(10,memory) as mem_trend wma(5,network_traffic) as net_trend.
```

salyh:
This issues and associated PRs covers Simple moving average (SMA) only.
For WMA (Weighted moving average) see issue https://github.com/opensearch-project/opensearch-spark/issues/831

PR #833, opened 2024-10-29
New trendline ppl command (SMA only)

salyh:
### Description
New trendline ppl command (SMA only). For WMA see https://github.com/opensearch-project/opensearch-spark/issues/831

### Related Issues
https://github.com/opensearch-project/opensearch-spark/issues/655

### Check List
- [ ] Implemented unit tests


ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -386,6 +386,19 @@ private java.util.Map<Alias, Field> buildLookupPair(List<OpenSearchPPLParser.Loo
...
+  public UnresolvedPlan visitTrendlineCommand(OpenSearchPPLParser.TrendlineCommandContext ctx) {
+    List<UnresolvedExpression> trendlineComputations = ctx.trendlineClause()
+            .stream()
+            .map(expressionBuilder::visit)
+            .collect(Collectors.toList());
+    return Optional.ofNullable(ctx.sortField())
+            .map(this::internalVisitExpression)
+            .map(Field.class::cast)
+            .map(sort -> new Trendline(sort, trendlineComputations))
+            .orElse(new Trendline(null, trendlineComputations));
```
YANG-DB, line 399: can you use optional instead of null ?

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/tree/Trendline.java
```diff
@@ -0,0 +1,68 @@
...
+
+@ToString
+@Getter
+@RequiredArgsConstructor
+@EqualsAndHashCode(callSuper = false)
+public class Trendline extends UnresolvedPlan {
+
+    private UnresolvedPlan child;
+    @Nullable
+    private final Field sortByField;
```
YANG-DB, line 24: this needs to be optional in case of no sort field given

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -82,14 +83,15 @@ commandName
    | RENAME
    | FILLNULL
    | FIELDSUMMARY
+   | TRENDLINE
```
YANG-DB, line 86: add to `TRENDLINE` to [keywordsCanBeId](https://github.com/opensearch-project/opensearch-spark/blob/cd730cb69b1b7a4f9e9fef6c090f9317a8fd5563/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L1011)

YANG-DB (changes requested):
Please also add command to the README, a dedicated page for this command with a link to the dedicated doc and examples in the examples page

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -640,6 +680,64 @@ public Expression visitSpan(Span node, CatalystPlanContext context) {
             return context.getNamedParseExpressions().push(window(field, value, node.getUnit()));
         }
 
+        @Override
+        public Expression visitTrendlineComputation(Trendline.TrendlineComputation node, CatalystPlanContext context) {
```
YANG-DB, line 684: @salyh can you please extract the content of this method and trendlineOrNullWhenThereAreTooFewDataPoints into a separate TrendlineStrategy class under this [utility package](https://github.com/opensearch-project/opensearch-spark/tree/main/ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils) ? it will simplify this visitor
thanks

YANG-DB (changes requested):

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
LantaoJin: @YANG-DB `commandName` set has already added to `keywordsCanBeId`. Not need to add it duplicately.

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/tree/Trendline.java
kt-eliatra: done

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
kt-eliatra: done

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
kt-eliatra: done

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -1116,4 +1130,5 @@ keywordsCanBeId
    | SEMI
    | ANTI
    | BETWEEN
+   | SMA
```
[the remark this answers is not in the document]
kt-eliatra: I'll remove this since it looks like a bug. It was already defined as `trendlineType` here https://github.com/eliatra/opensearch-spark/blob/606c40ef960e9dc4f3fb21aeef5ae4818951167a/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L260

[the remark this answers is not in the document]
kt-eliatra: done

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -247,6 +249,18 @@ fillnullCommand
    ;
 
 
+trendlineCommand
+   : TRENDLINE (SORT sortField)? trendlineClause (trendlineClause)*
+   ;
+
+trendlineClause
+   : trendlineType LT_PRTHS numberOfDataPoints = integerLiteral COMMA field = fieldExpression RT_PRTHS AS alias = fieldExpression
```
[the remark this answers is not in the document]
kt-eliatra: 1. It appeared in all the examples for this issue, that's why it's not optional atm.
2. changed to `qualifiedName`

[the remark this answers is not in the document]
LantaoJin: Alias should be defined to optional even all examples have aliases. In most case, the trendline command is the last command in pipe, user could ignore to name an alias

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
salyh: i agree, so nothing to do here

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
salyh: @LantaoJin @YANG-DB should we finally make `AS alias = fieldExpression` optional?

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -247,6 +249,18 @@ fillnullCommand
    ;
 
 
+trendlineCommand
+   : TRENDLINE (SORT sortField)? trendlineClause (trendlineClause)*
```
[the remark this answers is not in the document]
salyh: @YANG-DB @LantaoJin i think it adds (yet) unneccessary complexity because the use case can also be solved by simply chaining the trendline command

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
salyh: ok, will make it optional - what should be the name of the column if alias is omitted? just `trendline`?

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
YANG-DB: @LantaoJin - I like this idea but - I agree with @salyh here - lets release this as is for now and once we get any feedback we can revise this command...

YANG-DB (approved):

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
salyh: yes, good idea, @YANG-DB should i add this because you already approved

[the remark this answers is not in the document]
YANG-DB: yes plz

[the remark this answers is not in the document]
salyh: done

YANG-DB (changes requested):
@salyh plz add logical plan test in the [relevant folder](https://github.com/opensearch-project/opensearch-spark/tree/main/ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/ppl)

salyh:
> @salyh plz add logical plan test in the [relevant folder](https://github.com/opensearch-project/opensearch-spark/tree/main/ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/ppl) And also plz add a test with no alias example (for logical and actual queries)

done

YANG-DB (approved):

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/tree/Trendline.java
```diff
@@ -0,0 +1,62 @@
+package org.opensearch.sql.ast.tree;
```
jduo, line 1: Missing a license header here.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -386,6 +386,30 @@ private java.util.Map<Alias, Field> buildLookupPair(List<OpenSearchPPLParser.Loo
...
+    return Optional.ofNullable(ctx.sortField())
+            .map(this::internalVisitExpression)
+            .map(Field.class::cast)
+            .map(sort -> new Trendline(Optional.of(sort), trendlineComputations))
+            .orElse(new Trendline(Optional.empty(), trendlineComputations));
+  }
+
+  private Trendline.TrendlineComputation toTrendlineComputation(OpenSearchPPLParser.TrendlineClauseContext ctx) {
+    int numberOfDataPoints = Integer.parseInt(ctx.numberOfDataPoints.getText());
+    if (numberOfDataPoints < 0) {
```
jduo, line 404: Shouldn't this be < 1? Can't take the average with zero samples.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
```diff
@@ -386,6 +386,30 @@ private java.util.Map<Alias, Field> buildLookupPair(List<OpenSearchPPLParser.Loo
...
+      throw new SyntaxCheckException("Number of trendline data-points must be greater than or equal to 0");
+    }
+    Field dataField = (Field) expressionBuilder.visitFieldExpression(ctx.field);
+    String alias = ctx.alias == null?dataField.getField().toString()+"_trendline":ctx.alias.getText();
+    String computationType = ctx.trendlineType().getText();
```
jduo, line 409: Perhaps this should just evaluate the computation type enum now instead of later.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
```diff
@@ -0,0 +1,82 @@
+package org.opensearch.sql.ppl.utils;
```
jduo, line 1: License header needed.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+import org.opensearch.sql.ppl.CatalystQueryPlanVisitor;
+import scala.Option;
+import scala.Tuple2;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.opensearch.sql.ppl.utils.DataTypeTransformer.seq;
+
+public interface TrendlineStrategy {
```
jduo, line 19: Calling this a Strategy and making it an interface makes it easily confused with the Strategy pattern (where you delegate work to a Strategy interface and the Strategy is changeable at run time).

Suggest renaming to TrendlineCatalystUtils and make it a final @Utility class instead.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+
+    static List<NamedExpression> visitTrendlineComputations(CatalystQueryPlanVisitor.ExpressionAnalyzer expressionAnalyzer, List<Trendline.TrendlineComputation> computations, CatalystPlanContext context) {
```
jduo, line 21: The names and parameters of this and visitTrendlineComputation() are similar enough that it is hard to read (and also, this interface isn't a visitor so it might not be appropriate to prefix with "visit").

Suggest getting rid of visitTrendlineComputations, since it's trivial for the caller to do this.
Suggest rename visitTrendlineComputation() to be buildCatalystExpressionFromTrendlineComputation() to help clarity about what the method does.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+        expressionAnalyzer.visitLiteral(new Literal(Math.negateExact(node.getNumberOfDataPoints() - 1), DataType.INTEGER), context);
+        Expression windowLowerBoundary = context.popNamedParseExpressions().get();
+
+        //window definition
+        WindowSpecDefinition windowDefinition = new WindowSpecDefinition(
+                seq(),
+                seq(),
+                new SpecifiedWindowFrame(RowFrame$.MODULE$, windowLowerBoundary, CurrentRow$.MODULE$));
+
+        if (node.getComputationType() == Trendline.TrendlineType.SMA) {
```
jduo, line 38: Preconditions.checkArgument() here instead?

jduo (changes requested):

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
salyh: Not sure what you mean exactly but we will later have more trendline types than only SMA so we need an if or a case statement here.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
YANG-DB: agree - good point

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
YANG-DB: @jduo I don't understand what are you thinking here ? can u plz explain

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
YANG-DB: @jduo I think its an implementation issue - IMO we can keep it as is...

docs/ppl-lang/ppl-trendline-command.md
```diff
@@ -0,0 +1,60 @@
+## PPL trendline Command
+
+**Description**
+Using ``trendline`` command to calculate moving averages of fields.
+
+
+### Syntax
+`TRENDLINE [sort <[+|-] sort-field>] SMA(number-of-datapoints, field) [AS alias]`
```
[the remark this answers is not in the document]
salyh: fixed (sort is imho not a list here)

YANG-DB:
@jduo thanks for u'r feedback - really appreciated !

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
jduo: This builder extracts the trendline computation as a String then passes it as a String to TrendlineComputation's constructor. It might be better to just use TrendlineType.valueOf() at this point rather than pass around the String.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/TrendlineStrategy.java
salyh: I agree to rename it but other util classes in this packages are interfaces, so i vote to keep the interface type

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstBuilder.java
salyh: fixed

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
LantaoJin: > @YANG-DB @LantaoJin i think it adds (yet) unneccessary complexity because the use case can also be solved by simply chaining the trendline command

Sure, it brings me that could you add a test to chain multiple trendline commands in IT?

LantaoJin (changes requested):
https://github.com/opensearch-project/opensearch-spark/pull/833/files#r1823633566

LantaoJin:
Another question is whether the sort field can be different from the calculated field? If the answer is no, please throw out an exception for `| source = $testTable | trendline sort + name sma(3, age)`.
If the answer is yes, what is the special semantic for above query?

But this `sort` keyword in trendline still seems an one way door design, deep diving we could get another case:
`| source = $testTable | eval age_alias = age | trendline sort + age_alias sma(3, age)`.
Will this query fail either? There is no semantic problem IMO. That's why I comment and suggest [here](https://github.com/opensearch-project/opensearch-spark/pull/833#discussion_r1821772693).

However, I can give an approval to unblock releasing this as is for now as long as [above chaining case](https://github.com/opensearch-project/opensearch-spark/pull/833#pullrequestreview-2406704673) works.

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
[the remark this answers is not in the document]
salyh: added

salyh:
> https://github.com/opensearch-project/opensearch-spark/pull/833/files#r1823633566
> 
> To save time, you can ignore the plan assertion and just check the return data matching the expectation. Many thanks! Here is an example for chaining `trendline`: `| source = $testTable | eval age_1 = age, age_2 = age | trendline sort - age_1 sma(3, age_1) | trendline sort + age_2 sma(3, age_2)`

@LantaoJin @YANG-DB IT chaining test added

salyh:
> Another question is whether the sort field can be different from the calculated field? If the answer is no, please throw out an exception for `| source = $testTable | trendline sort + name sma(3, age)`. If the answer is yes, what is the special semantic for above query?
> 
> But this `sort` keyword in trendline still seems an one way door design, deep diving we could get another case: `| source = $testTable | eval age_alias = age | trendline sort + age_alias sma(3, age)`. Will this query fail either? There is no semantic problem IMO. That's why I comment and suggest [here](https://github.com/opensearch-project/opensearch-spark/pull/833#discussion_r1821772693).
> 
> However, I can give an approval to unblock releasing this as is for now as long as [above chaining case](https://github.com/opensearch-project/opensearch-spark/pull/833#pullrequestreview-2406704673) works.

I suggest to return to this when [WMA](https://github.com/opensearch-project/opensearch-spark/issues/831) will be implemented

YANG-DB (approved):

YANG-DB:
@salyh plz solve conflicts after recent changes...

salyh:
> @salyh plz solve conflicts after recent changes...

done

YANG-DB (approved):

YANG-DB (approved):

YANG-DB:
> Another question is whether the sort field can be different from the calculated field? If the answer is no, please throw out an exception for `| source = $testTable | trendline sort + name sma(3, age)`. If the answer is yes, what is the special semantic for above query?
> 
> But this `sort` keyword in trendline still seems an one way door design, deep diving we could get another case: `| source = $testTable | eval age_alias = age | trendline sort + age_alias sma(3, age)`. Will this query fail either? There is no semantic problem IMO. That's why I comment and suggest [here](https://github.com/opensearch-project/opensearch-spark/pull/833#discussion_r1821772693).
> 
> However, I can give an approval to unblock releasing this as is for now as long as [above chaining case](https://github.com/opensearch-project/opensearch-spark/pull/833#pullrequestreview-2406704673) works.

@LantaoJin I'm approving the PR for merge as this issue was addresses [here](https://github.com/opensearch-project/opensearch-spark/blob/ffcc9189a142616dd4a78e9928ecb966401e7c5c/integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTrendlineITSuite.scala#L217)

PR #837, opened 2024-10-30
Add metrics for successful/failed Spark index creation

Swiddis:
### Description
Success and failure metrics for spark index creation -- note that "failure" here means an execution failure (e.g. Spark couldn't execute the query), not validation failure (e.g. the index already exists).

E.g. success for creating a skipping index:
image

### Related Issues
N/A

### Check List
- [ ] Implemented unit tests
- [X] New added source code should include a copyright header
- [X] Commits are signed per the DCO using `--signoff`


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSparkIndexFactory.scala
```diff
@@ -45,9 +47,12 @@ object FlintSparkIndexFactory extends Logging {
    */
   def create(spark: SparkSession, metadata: FlintMetadata): Option[FlintSparkIndex] = {
     try {
-      Some(doCreate(spark, metadata))
+      val result = doCreate(spark, metadata)
+      emitIndexCreationStatusMetric(metadata, success = true)
+      Some(result)
     } catch {
       case e: Exception =>
+        emitIndexCreationStatusMetric(metadata, success = false)
```
dai-chen, line 55: Could you double check if this is the right place? I recall this factory is actually deserializer from Flint metadata to index class. I guess you may want to intercept API call in `FlintSpark`? Also for all metrics related logic, please consider AOP style rather than insert this logic in many places.

noCharger: +1, sample code https://github.com/opensearch-project/opensearch-spark/blob/main/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/refresh/util/RefreshMetricsAspect.scala

Swiddis: Moved the metric collection to the new location. I [experimented with trying something AOP-like but it was cumbersome](https://github.com/opensearch-project/opensearch-spark/blob/d77c723f43f181f59ff453228eb3cea35412f2ec/flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java#L134-L144) -- linking the existing logic in spark/refresh introduces a circular dependency, and the callable implicitly raising `Exception` makes error handling awkward if trying to make a general-purpose duplicate of what exists on the Scala side.

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
```diff
@@ -122,4 +136,19 @@ public IRestHighLevelClient createClient() {
   private String sanitizeIndexName(String indexName) {
     return OpenSearchClientUtils.sanitizeIndexName(indexName);
   }
+
+  private void emitIndexCreationMetric(String indexKind, boolean createSuccessful) {
+    String stateStr = createSuccessful ? "success" : "failed";
+    switch (indexKind) {
```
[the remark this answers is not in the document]
Swiddis: Default case is no-op, which linter says is redundant and should be removed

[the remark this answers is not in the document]
Swiddis: Default case is no-op, which linter says is redundant and should be removed

flint-core/src/main/scala/org/opensearch/flint/core/storage/FlintOpenSearchClient.java
```diff
@@ -122,4 +130,36 @@ public IRestHighLevelClient createClient() {
...
+      case "covering":
+        MetricsUtil.addHistoricGauge(String.format("%s.%s.count", MetricConstants.CREATE_COVERING_INDICES, "failed"), 1);
+        break;
+      case "mv":
+        MetricsUtil.addHistoricGauge(String.format("%s.%s.count", MetricConstants.CREATE_MV_INDICES, "failed"), 1);
+        break;
+      default:
+        break;
+    }
+  }
```
ykmr1224, line 164: Sorry if I confused you, but we want to have a method to reduce redundancy.

```
private void emitIndexCreationSuccessMetric(String indexKind) {
  emitIndexCreationMetric(indexKind, "success");
}

private void emitIndexCreationSuccessMetric(String indexKind) {
  emitIndexCreationMetric(indexKind, "failed");
}

private void emitIndexCreationMetric(String indexKind, String status) {
...
}
```

It might be simpler to directly call emitIndexCreationMetric with "success" or "failed" parameter.

ykmr1224 (approved):

Issue #746, opened 2024-10-04
[FEATURE] Write metadata to index mappings `_meta` as read cache for frontend user to access


seankao-az:
**Is your feature request related to a problem?**
Write the following to Flint index mappings `_meta` field so that frontend user can use it as a read cache for crucial metadata access.
* refresh interval
* source table names
* last refresh time

**What solution would you like?**
There’ll be a `_meta` field in the index mappings, within it will be a `properties` field, which contain the following fields:

* metadataCacheVersion
    * type: String
    * value will be constant "1.0"
* refreshInterval
    * type: Int
    * unit: seconds
    * this field is added only if index refresh type is auto refresh and refresh_interval option is set
* sourceTables
    * Array of String (table names)
* lastRefreshTime
    * type: Long
    * unit: milliseconds
    * this field is added only if index already gets refreshed at least once

Example:
```
GET myindex/_mappings

{
  "myindex": {
    "mappings": {
      "_meta": {
        "properties": {
          "metadataCacheVersion": "1.0",
          "refreshInterval": 900,
          "sourceTables": [
            "mys3.default.table1",
            "mys3.default.table2"
          ],
          "lastRefreshTime": 1727395328283
        }
      }
      "properties": { // existing OpenSearch index mapping
        "field1": ...
        "field2": ...
      }
    }
  }
}
```

**Incoming changes**
1. Writer class to handle the metadata cache writing; flow control
2. Add source tables to index metadata for materialized view
3. Track last refresh time for index
    1. Out of Scope: track each micro batch execution time for internal scheduler (spark streaming query)

PR #854, opened 2024-10-31
Extract source table names from mv query

seankao-az:
### Description
Extract source table names from mv query and add to `_meta.properties.sourceTables`

This is actually already introduced in #297, but we used it only for validation and didn't store the table names

#### Example

```
CREATE MATERIALIZED VIEW myglue_test.default.mv_test_source_tables
AS SELECT status FROM myglue_test.default.http_logs
WITH (
  auto_refresh = true,
  checkpoint_location = 's3://test/'
)

GET flint_myglue_test_default_mv_test_source_tables/_mappings
{
  "flint_myglue_test_default_mv_test_source_tables": {
    "mappings": {
      "_meta": {
        "latestId": "ZmxpbnRfbXlnbHVlX3Rlc3RfZGVmYXVsdF9tdl90ZXN0X3NvdXJjZV90YWJsZXM=",
        "kind": "mv",
        "indexedColumns": [
          {
            "columnType": "int",
            "columnName": "status"
          }
        ],
        "name": "myglue_test.default.mv_test_source_tables",
        "options": {
          "auto_refresh": "true",
          "refresh_interval": "5 minutes",
          "scheduler_mode": "external",
          "incremental_refresh": "false",
          "checkpoint_location": "s3://test/"
        },
        "source": "SELECT status FROM myglue_test.default.http_logs",
        "version": "0.6.0",
        "properties": {
          "sourceTables": [
            "myglue_test.default.http_logs"
          ],
          "env": {
            "SERVERLESS_EMR_VIRTUAL_CLUSTER_ID": "****",
            "SERVERLESS_EMR_JOB_ID": "****"
          }
        }
      },
      "properties": {
        "status": {
          "type": "integer"
        }
      }
    }
  }
}
```

### Related Issues
* #746 

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/metadatacache/FlintMetadataCache.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/metadatacache/FlintMetadataCache.scala
@@ -10,6 +10,7 @@ import scala.collection.JavaConverters.mapAsScalaMapConverter
 import org.opensearch.flint.common.metadata.FlintMetadata
 import org.opensearch.flint.common.metadata.log.FlintMetadataLogEntry
 import org.opensearch.flint.spark.FlintSparkIndexOptions
+import org.opensearch.flint.spark.mv.FlintSparkMaterializedView.MV_INDEX_TYPE
 import org.opensearch.flint.spark.scheduler.util.IntervalSchedulerParser
 
 /**
@@ -47,8 +48,6 @@ case class FlintMetadataCache(
 object FlintMetadataCache {
 
   val metadataCacheVersion = "1.0"
-  val mockTableName =
-    "dataSourceName.default.logGroups(logGroupIdentifier:['arn:aws:logs:us-east-1:123456:test-llt-xa', 'arn:aws:logs:us-east-1:123456:sample-lg-1'])"
 
   def apply(metadata: FlintMetadata): FlintMetadataCache = {
     val indexOptions = FlintSparkIndexOptions(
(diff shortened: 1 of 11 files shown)


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/mv/FlintSparkMaterializedView.scala
```diff
@@ -193,6 +196,13 @@ object FlintSparkMaterializedView {
      */
     def query(query: String): Builder = {
       this.query = query
+      // Extract source table names (possibly more than one)
+      this.sourceTables = flint.spark.sessionState.sqlParser
+        .parsePlan(query)
+        .collect { case relation: UnresolvedRelation =>
+          qualifyTableName(flint.spark, relation.tableName)
+        }
+        .toArray
```
ykmr1224, line 205: I think it is better extracting this as a method. (The method name should be like `extractSourceTableNames` and then comment won't be needed)

ykmr1224 (approved):

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/metadatacache/FlintMetadataCache.scala
```diff
@@ -61,14 +60,22 @@ object FlintMetadataCache {
     } else {
       None
     }
+    val sourceTables = metadata.kind match {
+      case MV_INDEX_TYPE =>
+        metadata.properties.get("sourceTables") match {
+          case list: java.util.ArrayList[_] =>
+            list.toArray.map(_.toString)
+          case _ => Array.empty[String]
```
dai-chen, line 68: Old MV doesn't have this property right? If so, shall we generate source tables on the fly in `FlintSparkMaterializedView` or elsewhere if empty? Because I saw we changed other code like `isTableProviderSupported` to depend on this field.

seankao-az: great point

seankao-az: Update to detect missing sourceTables and generate in `FlintSparkIndexFactory` for backward compatibility.
It won't necessarily update the index metadata storage, though, unless an `updateIndexMetadata` is explicitly called, which happens at RECOVER and ALTER. (also for CREATE but this always creates new index, and is irrelevant to old index)

dai-chen (approved):

Issue #869, opened 2024-11-05
[FEATURE] Add `TO_JSON_STRING` and `ARRAY_LENGHT` functions and refactor the `JSON` function


LantaoJin:
**Is your feature request related to a problem?**
A followup of https://github.com/opensearch-project/opensearch-spark/issues/667

1. Refactor `JSON`: Currently the `JSON` evaluates STRING/JSON_ARRAY/JSON_OBJECT. It brings many confusions. In Splunk, the data could be stored as JSON object, `JSON` function evaluates JSON object. But in Spark, there is no JSON object, but it contains STRING, StructType and ArrayType. So the `JSON` function can only evaluate STRING types.

 2. Add `ARRAY_LENGHT`: for the same reason above, we separate `JSON_ARRAY_LENGHT` to `JSON_ARRAY_LENGHT` and `ARRAY_LENGHT`. `JSON_ARRAY_LENGHT` only accepts STRING type, `ARRAY_LENGHT` only accepts ArrayType.

3. Add `TO_JSON_STRING`: After the refactor of (1), we still need a method to convert JSON_ARRAY/JSON_OBJECT to valid JSON STRING. `TO_JSON_STRING` accepts both StructType and ArrayType as input and returns JSON formatted string.

PR #780, opened 2024-10-15
Support PPL `JSON` functions: construction and extraction

LantaoJin:
### Description
This PR is the 1st part of addressing JSON functions requirement. It includes JSON construction and extraction related functions:

- json_object()
- json()
- json_array()
- json_array_lenght()
- json_extract()
- json_keys()
- json_valid()

The 2nd part (not this PR) of JSON functions are conversion and truncation functions:

- json_append
- json_delete
- json_extend
- json_set
- json_array_all_match
- json_array_any_match
- json_array_filter
- json_array_map
- json_array_reduce

### `JSON`

**Description**

`json(value)` Evaluates whether a value can be parsed as JSON. Returns the json string if valid, null otherwise.

**Argument type:** STRING/JSON_ARRAY/JSON_OBJECT

**Return type:** STRING

A STRING expression of a valid JSON object format.

Example:

    os> source=people | eval `valid_json()` = json('[1,2,3,{"f1":1,"f2":[5,6]},4]') | fields valid_json
    fetched rows / total rows = 1/1
    +---------------------------------+
    | valid_json                      |
    +---------------------------------+
    | [1,2,3,{"f1":1,"f2":[5,6]},4]   |
    +---------------------------------+

    os> source=people | eval `invalid_json()` = json('{"invalid": "json"') | fields invalid_json
    fetched rows / total rows = 1/1
    +----------------+
    | invalid_json   |
    +----------------+
    | null           |
    +----------------+

### `JSON_OBJECT`

**Description**

`json_object(<key>, <value>[, <key>, <value>]...)` returns a JSON object from members of key-value pairs.

**Argument type:**
- A \<key\> must be STRING.
- A \<value\> can be any data types.

**Return type:** JSON_OBJECT (Spark StructType)

A StructType expression of a valid JSON object.

Example:

    os> source=people | eval result = json(json_object('key', 123.45)) | fields result
    fetched rows / total rows = 1/1
    +------------------+
    | result           |
    +------------------+
    | {"key":123.45}   |
    +------------------+

    os> source=people | eval result = json(json_object('outer', json_object('inner', 123.45))) | fields result
    fetched rows / total rows = 1/1
    +------------------------------+
    | result                       |
    +------------------------------+
    | {"outer":{"inner":123.45}}   |
    +------------------------------+

### `JSON_ARRAY`

**Description**

`json_array(<value>...)` Creates a JSON ARRAY using a list of values.

**Argument type:**
- A \<value\> can be any kind of value such as string, number, or boolean.

[...]

Issue #675, opened 2024-09-17
[FEATURE] Add PPL support to access values in array


A-Gray-Cat:
**Is your feature request related to a problem?**
Currently, PPL doesn't support accessing values inside an array. For example, 

```
{
    "items": ["a", "b", "c"]
}
```

Users can't select value a, b, or c using PPL.

**What solution would you like?**
Users can access the first element of the array by using syntax like `eval first_item = items[0]`, or it can be retrieved directly using the `fields` command, e.g. `fields items[0]`.

YANG-DB:
@A-Gray-Cat thanks - we will review !

dblock:
[Catch All [Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta),  [3](https://github.com/krisfreedain),  [4](https://github.com/lukas-vlcek/)]

LantaoJin:
With #780 , we can access the first element in a array by
```
| eval first_elem = json_extract('{"items": ["a", "b", "c"]}', '$.items[0]') // json object string
```
or
```
| eval first_elem = json_extract('["a", "b", "c"]', '$.[0]') // json array string
```
But it requires the input is a json string. If the input is json array (ArrayType), I think we need a new function such as `array_get`:
```
| eval first_elem = array_get(json_array("a", "b", "c"), 0)
```
But as a workaround, we could rewrite by
```
| eval first_elem = json_extract(to_json_string(json_array("a", "b", "c")), '$.[0]')
```

From the issue's description, it seems the input is a json string and `json_extract` should work as the solution. @A-Gray-Cat

YANG-DB:
@A-Gray-Cat we are closing this since we have a workaround as described above

PR #870, opened 2024-11-05
[followup] Refactor JSON function and add TO_JSON_STRING, ARRAY_LENGHT functions

LantaoJin:
### Description

This is a followup of PR https://github.com/opensearch-project/opensearch-spark/pull/780

1. Refactor `JSON`: Currently the `JSON` can evaluate STRING/JSON_ARRAY/JSON_OBJECT. It brings many confusions and bugs. It because in Splunk or OpenSearch, the data could be stored as JSON object, `JSON` function evaluates JSON object and return its value or null if invalid. But in Spark, there is no JSON object, but it contains STRING, StructType and ArrayType. So to evaluate a json string, `JSON` function should accept STRING type only.

 2. Add `ARRAY_LENGHT`: for the same reason, the original `JSON_ARRAY_LENGHT` accepts both STRINGand JSON_ARRAY(ArrayType). We separate `JSON_ARRAY_LENGHT` to `JSON_ARRAY_LENGHT` and `ARRAY_LENGHT`. The `JSON_ARRAY_LENGHT` only accepts STRING type, `ARRAY_LENGHT` only accepts ArrayType.

3. Add `TO_JSON_STRING`: After the refactor of (1), we still need a method to convert JSON_ARRAY/JSON_OBJECT to valid JSON STRING. `TO_JSON_STRING` accepts both StructType and ArrayType as input and returns JSON formatted string.

### Related Issues
Resolves https://github.com/opensearch-project/opensearch-spark/issues/869

### Check List
- [x] Implemented unit tests

### DIFF
--- a/docs/ppl-lang/functions/ppl-json.md
+++ b/docs/ppl-lang/functions/ppl-json.md
@@ -4,11 +4,11 @@
 
 **Description**
 
-`json(value)` Evaluates whether a value can be parsed as JSON. Returns the json string if valid, null otherwise.
+`json(value)` Evaluates whether a string can be parsed as JSON format. Returns the string value if valid, null otherwise.
 
-**Argument type:** STRING/JSON_ARRAY/JSON_OBJECT
+**Argument type:** STRING
 
-**Return type:** STRING
+**Return type:** STRING/NULL
 
 A STRING expression of a valid JSON object format.
 
@@ -47,7 +47,7 @@ A StructType expression of a valid JSON object.
 
 Example:
 
-    os> source=people | eval result = json(json_object('key', 123.45)) | fields result
+    os> source=people | eval result = json_object('key', 123.45) | fields result
     fetched rows / total rows = 1/1
     +------------------+
     | result           |
@@ -55,7 +55,7 @@ Example:
     | {"key":123.45}   |
     +------------------+
 
-    os> source=people | eval result = json(json_object('outer', json_object('inner', 123.45))) | fields result
+    os> source=people | eval result = json_object('outer', json_object('inner', 123.45)) | fields result
     fetched rows / total rows = 1/1
     +------------------------------+
     | result                       |
@@ -81,13 +81,13 @@ Example:
 
     os> source=people | eval `json_array` = json_array(1, 2, 0, -1, 1.1, -0.11)
     fetched rows / total rows = 1/1
-    +----------------------------+
-    | json_array                 |
-    +----------------------------+
-    | 1.0,2.0,0.0,-1.0,1.1,-0.11 |
-    +----------------------------+
+    +------------------------------+
+    | json_array                   |
+    +------------------------------+
+    | [1.0,2.0,0.0,-1.0,1.1,-0.11] |
+    +------------------------------+
 
-    os> source=people | eval `json_array_object` = json(json_object("array", json_array(1, 2, 0, -1, 1.1, -0.11)))
+    os> source=people | eval `json_array_object` = json_object("array", json_array(1, 2, 0, -1, 1.1, -0.11))
     fetched rows / total rows = 1/1
     +----------------------------------------+
     | json_array_object                      |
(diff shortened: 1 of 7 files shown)


docs/ppl-lang/functions/ppl-json.md
```diff
@@ -81,29 +81,58 @@ Example:
...
-    os> source=people | eval `json_array_object` = json(json_object("array", json_array(1, 2, 0, -1, 1.1, -0.11)))
+    os> source=people | eval `json_array_object` = json_object("array", json_array(1, 2, 0, -1, 1.1, -0.11))
     fetched rows / total rows = 1/1
     +----------------------------------------+
     | json_array_object                      |
     +----------------------------------------+
     | {"array":[1.0,2.0,0.0,-1.0,1.1,-0.11]} |
     +----------------------------------------+
 
+### `TO_JSON_STRING`
```
LantaoJin, line 98: About the naming: `to_json` is good name for Spark since there is no json object data type. So `to_json` in Spark is converting StructType to json String. But `to_json` might not a good name for PPL since it's easy to cause misunderstanding as converting something to json object.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/BuiltinFunctionTransformer.java
```diff
@@ -102,7 +104,9 @@ public interface BuiltinFunctionTransformer {
             .put(COALESCE, "coalesce")
             .put(LENGTH, "length")
             .put(TRIM, "trim")
+            .put(ARRAY_LENGTH, "array_size")
```
LantaoJin, line 107: We choose `array_length` instead of `array_size` since we have had a function `json_array_length`. So it maps to the Spark builtin function `array_size`.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/BuiltinFunctionTransformer.java
```diff
@@ -126,26 +130,12 @@ public interface BuiltinFunctionTransformer {
         .put(
             JSON_ARRAY_LENGTH,
             args -> {
-                // Check if the input is an array (from json_array()) or a JSON string
-                if (args.get(0) instanceof UnresolvedFunction) {
```
LantaoJin, line 130: The current logic in this `if-else` could cause bugs. 
Both
`| eval b = json_array_length(json_array(1,2,3))` and 
`| eval b = json_array_length("[1,2,3]")` work fine.

 But imaging following two cases:
`| eval a = json_array(1,2,3), b = json_array_length(a)`
`| eval a = "[1,2,3]", b = json_array_length(a)`
The args of `json_array_length` are both `UnresolvedAttribute` which goes into `else` branch. But `a` in the first query will be resolved to `UnresolveFunction("array"..)` and `a` in the second case will be resolved to `StringLiteral`.

docs/ppl-lang/functions/ppl-json.md
YANG-DB: yes sounds like a better distinction ...

ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/ppl/PPLLogicalPlanJsonFunctionsTranslatorTestSuite.scala
YANG-DB, line 1: @LantaoJin did we cover the next [issue](https://github.com/opensearch-project/opensearch-spark/issues/675)?

YANG-DB (approved):
@LantaoJin I approved but plz make sure  we cover the next [issue](https://github.com/opensearch-project/opensearch-spark/issues/675)

ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/ppl/PPLLogicalPlanJsonFunctionsTranslatorTestSuite.scala
LantaoJin: With #780 , we can access the first element in a array by
```
| eval first_elem = json_extract('{"items": ["a", "b", "c"]}', '$.items[0]') // json object string
```
or
```
| eval first_elem = json_extract('["a", "b", "c"]}', '$.[0]') // json array string
```
But it requires the input is a json string. If the input is json array (ArrayType), I think we need a new function such as `array_get`:
```
| eval first_elem = array_get(json_array("a", "b", "c"), 0)
```
But as a workaround, we could rewrite by
```
| eval first_elem = json_extract(to_json_string(json_array("a", "b", "c")), , '$.[0]')
```
So keep https://github.com/opensearch-project/opensearch-spark/issues/675 open until we have `array_get`.

LantaoJin: Em, I think I can add `array_get` ASAP in this PR.

PR #879, opened 2024-11-07
Add `sample` parameter to `top` & `rare` command

YANG-DB:
### Description
Add a new sample command (`sample`) to reduce amount of scanned data points and allow approximation of a `top` or `rare` statements when faster sample based results if favour of exact long running results 

```sql
source = testTable  | rare address sample(50 percent)
source = testTable  | top 5 address by country sample(25 percent)
```

### Issues Resolved
- https://github.com/opensearch-project/opensearch-spark/issues/740

### Check List
- [x] Implemented unit tests

### DIFF
--- a/integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTopAndRareITSuite.scala
+++ b/integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTopAndRareITSuite.scala
@@ -83,6 +83,52 @@ class FlintSparkPPLTopAndRareITSuite
     val expectedPlan = Project(projectList, sortedPlan)
     comparePlans(expectedPlan, logicalPlan, checkAnalysis = false)
   }
+  
+  test("create ppl rare address field query test sample 75 %") {
+    val frame = sql(s"""
+         | source = $testTable| rare address sample(75 percent)
+         | """.stripMargin)
+
+    // Retrieve the results
+    val results: Array[Row] = frame.collect()
+    assert(results.length == 3)
+
+    val expectedRow = Row(1, "Vancouver")
+    assert(
+      results.head == expectedRow,
+      s"Expected least frequent result to be $expectedRow, but got ${results.head}")
+
+    // Retrieve the logical plan
+    val logicalPlan: LogicalPlan = frame.queryExecution.logical
+    // Define the expected logical plan
+    val addressField = UnresolvedAttribute("address")
+    val projectList: Seq[NamedExpression] = Seq(UnresolvedStar(None))
+
+    val aggregateExpressions = Seq(
+      Alias(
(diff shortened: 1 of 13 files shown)


LantaoJin:
One high level question:
How do we determine the relationship between percentage and precision? Or how much precision does it lose when sampling is decreased from 100% to 80% or from 80% to 50%?

ppl-spark-integration/src/main/antlr4/OpenSearchPPLLexer.g4
```diff
@@ -79,6 +80,7 @@ DESC:                               'DESC';
 DATASOURCES:                        'DATASOURCES';
 USING:                              'USING';
 WITH:                               'WITH';
+PERCENT:                            'PERCENT';
```
LantaoJin, line 83: SAMPLE and PERCENT should be added to `keywordsCanBeId`

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystPlanContext.java
```diff
@@ -117,6 +137,17 @@ public LogicalPlan withRelation(UnresolvedRelation relation) {
...
+    /**
+     * append sample-relation to relations list
+     *
+     * @param sampleRelation
+     * @return
+     */
+    public LogicalPlan withSampleRelation(Sample sampleRelation) {
+        this.relations.add(sampleRelation.child());
+        return with(sampleRelation);
+    }
```
LantaoJin, line 149: I really don't like this implementation. As I mentioned in another PR, `sample` IMO should be a common plan node instead of binding to `relation`. I am not sure how current behaviour works when a plan contains joins or correlated subqueries. For example, in query
```
source=tableA | join ON A.id = B.id tableB | top 1 A.id sample(50 per)
```
does it equals to
```
source=tableA sample(50 per) | join ON A.id = B.id [ source=tableB sample(50 per) ] | top 1 A.id
```
or
```
source=tableA | join ON A.id = B.id tableB | sample 50 per | top 1 A.id
```

integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLTopAndRareITSuite.scala
```diff
@@ -84,6 +84,52 @@ class FlintSparkPPLTopAndRareITSuite
     comparePlans(expectedPlan, logicalPlan, checkAnalysis = false)
   }
 
+  test("create ppl rare address field query test sample 75 %") {
```
LantaoJin, line 87: I strongly suggest to add some IT cases for top/rare with sample in complex join and subquery query.

YANG-DB:
closing since this has not yet shown to have a significant use case

PR #883, opened 2024-11-08
Fix bug for not able to get sourceTables from metadata

seankao-az:
### Description
Using custom implementation for `FlintIndexMetadataService` exposes a bug where we incorrectly get empty `sourceTables` from `FlintMetadata`.
Depending on how the `FlintMetadata` is generated, the `metadata.properties.get("sourceTables")` could either be of type java array or scala array.
This PR fixes the bug, now considering both cases.

Also added some logs for troubleshooting in production environment.

### Related Issues
* #746 
* #854 

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/mv/FlintSparkMaterializedView.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/mv/FlintSparkMaterializedView.scala
@@ -7,6 +7,7 @@ package org.opensearch.flint.spark.mv
 
 import java.util.Locale
 
+import scala.collection.JavaConverters.asScalaBufferConverter
 import scala.collection.JavaConverters.mapAsJavaMapConverter
 import scala.collection.convert.ImplicitConversions.`map AsScala`
 
@@ -18,6 +19,7 @@ import org.opensearch.flint.spark.FlintSparkIndexOptions.empty
 import org.opensearch.flint.spark.function.TumbleFunction
 import org.opensearch.flint.spark.mv.FlintSparkMaterializedView.{getFlintIndexName, MV_INDEX_TYPE}
 
+import org.apache.spark.internal.Logging
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.catalyst.FunctionIdentifier
 import org.apache.spark.sql.catalyst.analysis.{UnresolvedFunction, UnresolvedRelation}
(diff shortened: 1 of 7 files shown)


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/mv/FlintSparkMaterializedView.scala
```diff
@@ -179,13 +181,43 @@ object FlintSparkMaterializedView {
...
+   * Get source tables from Flint metadata properties field.
+   *
+   * @param metadata
+   *   Flint metadata
+   * @return
+   *   source table names
+   */
+  def getSourceTablesFromMetadata(metadata: FlintMetadata): Array[String] = {
+    logInfo(s"Getting source tables from metadata $metadata")
+    metadata.properties.get("sourceTables") match {
```
ykmr1224, line 206: Can `metadata.properties` to be null?

dai-chen (commented):
Is it possible to enforce a single and language-independent format in metadata, e.g. JSON array?

flint-spark-integration/src/main/scala/org/opensearch/flint/spark/mv/FlintSparkMaterializedView.scala
seankao-az: When FlintMetadata is built before passing in, we add a `properties` field even if it's empty..
We could still do a null check just to be safe.

seankao-az:
> Is it possible to enforce a single and language-independent format in metadata, e.g. JSON array?

Not quite sure..
Since when reading from OpenSearch index mappings, we use jackson library to deserialize into `FlintMetadata`, and jackson library will use `java.util.ArrayList`. I'll try using this type instead.

Or do you mean using json string? Like
```
{"sourceTables": "[\"a.b.c\"], [\"d.e.f\"]"}
```

dai-chen:
> > Is it possible to enforce a single and language-independent format in metadata, e.g. JSON array?
> 
> Not quite sure.. Since when reading from OpenSearch index mappings, we use jackson library to deserialize into `FlintMetadata`, and jackson library will use `java.util.ArrayList`. I'll try using this type instead.
> 
> Or do you mean using json string? Like
> 
> ```
> {"sourceTables": "[\"a.b.c\"], [\"d.e.f\"]"}
> ```

Yeah, just trying to understand where this problem comes from? Is Scala or Java array stored in metadata?

seankao-az:
mainly comes from this line of scala code. I was putting scala Array[String] here.
https://github.com/opensearch-project/opensearch-spark/pull/883/files#diff-1e6cbb4cde65f1142862f3b539a438d534fd70a229499efd671ef9aae67ff29fR75

Now I've changed it to use ArrayList. so for both cases (`mv.metadata()` and deserialized metadata from index mapping) will have ArrayList. Do you think this is better? @dai-chen

dai-chen (approved):

PR #889, opened 2024-11-11
Add jvmGCTime metrics

ykmr1224:
### Description
- Add jvmGCTime metriccs
- emit per task metrics and total for query
- Renamed ReadWriteBytesSparkListener to MetricsSparkListener to handle multiple metrics

### Related Issues
n/a

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
+++ b/flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
@@ -175,6 +175,16 @@ public final class MetricConstants {
      */
     public static final String INITIAL_CONDITION_CHECK_FAILED_PREFIX = "initialConditionCheck.failed.";
 
+    /**
+     * Metric for tracking the JVM GC time per task
+     */
+    public static final String TASK_JVM_GC_TIME_METRIC = "task.jvmGCTime.count";
+
+    /**
+     * Metric for tracking the total JVM GC time for query
+     */
+    public static final String TOTAL_JVM_GC_TIME_METRIC = "query.totalJvmGCTime.count";
+
     private MetricConstants() {
         // Private constructor to prevent instantiation
     }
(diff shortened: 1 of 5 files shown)


flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
```diff
@@ -175,6 +175,16 @@ public final class MetricConstants {
      */
     public static final String INITIAL_CONDITION_CHECK_FAILED_PREFIX = "initialConditionCheck.failed.";
 
+    /**
+     * Metric for tracking the JVM GC time per task
+     */
+    public static final String TASK_JVM_GC_TIME_METRIC = "task.jvmGCTime.count";
```
penghuo, line 181: why add `.count` as postfix?

flint-core/src/main/scala/org/opensearch/flint/core/metrics/MetricsSparkListener.scala
```diff
@@ -31,21 +32,28 @@ class ReadWriteBytesSparkListener extends SparkListener with Logging {
     recordsRead += inputMetrics.recordsRead
     bytesWritten += outputMetrics.bytesWritten
     recordsWritten += outputMetrics.recordsWritten
+    totalJvmGcTime += taskEnd.taskMetrics.jvmGCTime
```
penghuo, line 35: onTaskEnd is called on single thread?

flint-core/src/main/java/org/opensearch/flint/core/metrics/MetricConstants.java
ykmr1224: We currently put `.count` or `.processingTime` suffix for all the metrics since metrics are filtered based on suffix.

I chose `.count` for jvmGCTime since Spark also use `.count` suffix ([ref](https://spark.apache.org/docs/3.5.1/monitoring.html#component-instance--executor)).
(It is like the middle of count and processingTime, since it is summarizing all the GCTimes...)

flint-core/src/main/scala/org/opensearch/flint/core/metrics/MetricsSparkListener.scala
ykmr1224: That's good point.
I checked the spark implementation, and SparkListener is called from LiveListenerBus which process events in single thread. (It is designed to process events one by one to avoid concurrency issue) [ref](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/scheduler/LiveListenerBus.scala)

PR #930, opened 2024-11-19
Support alter from manual to auto without specifying scheduler mode

noCharger:
### Description
Support alter from manual to auto without specifying scheduler mode. In detail:

Step 1:
Create a manual refresh index (full or incremental)

Step 2:
Alter it by set auto_refreh=true, by default it will switch to auto refresh mode and use external scheduler when `spark.flint.job.externalScheduler.enabled = true` 

### Check List

### DIFF
--- a/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
+++ b/flint-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintSpark.scala
@@ -510,6 +510,10 @@ class FlintSpark(val spark: SparkSession) extends FlintSparkTransactionSupport w
   private def isSchedulerModeChanged(
       originalOptions: FlintSparkIndexOptions,
       updatedOptions: FlintSparkIndexOptions): Boolean = {
+    // Altering from manual to auto should not be interpreted as a scheduling mode change.
+    if (!originalOptions.options.contains(SCHEDULER_MODE.toString)) {
+      return false
+    }
     updatedOptions.isExternalSchedulerEnabled() != originalOptions.isExternalSchedulerEnabled()
   }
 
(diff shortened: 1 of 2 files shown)


seankao-az (commented):
What happens in these 2 scenarios?

Scenario 1
1. create index with "auto_refresh": true and "scheduler_mode": "external"
2. alter index with "auto_refresh": false
3. alter index with "auto_refresh": true

Scenario 2
1. create index with "auto_refresh": true and "scheduler_mode": "internal"
2. alter index with "auto_refresh": false
3. alter index with "auto_refresh": true

Is it intended that in both scenarios the resulting index should be auto refresh index using external scheduler?
Does the condition check in `isSchedulerModeChanged` cover this case? I believe originalOptions will still have "scheduler_mode": "internal"

noCharger:
> What happens in these 2 scenarios?
> 
> Scenario 1
> 
> 1. create index with "auto_refresh": true and "scheduler_mode": "external"
> 2. alter index with "auto_refresh": false
> 3. alter index with "auto_refresh": true
> 
> Scenario 2
> 
> 1. create index with "auto_refresh": true and "scheduler_mode": "internal"
> 2. alter index with "auto_refresh": false
> 3. alter index with "auto_refresh": true
> 
> Is it intended that in both scenarios the resulting index should be auto refresh index using external scheduler? Does the condition check in `isSchedulerModeChanged` cover this case? I believe originalOptions will still have "scheduler_mode": "internal"

If `scheduler_mode` is already exist in original index option, it should copied to the updated index and behave as of expacted, hense `Scenario 2` should be internal instead. The case to cover in this PR is when original index option doesn't have scheduler_mode field.

seankao-az (approved):

ykmr1224 (approved):

PR #979, opened 2024-12-09
Implement FlintJob to handle all query types in warmpool mode

saranrajnk:
### Description

This PR introduces support for FlintJob to handle all types of queries — interactive, streaming, and batch — with all data sources in warmpool mode. Additionally, FlintJob will also support non-warmpool mode for streaming and batch queries, configurable via a Spark configuration setting.

FlintJob invokes `Warmpool.scala`, which in turn calls the client to continuously fetch queries for execution. The client sets various Spark configurations, such as the datasource, resultIndex, and other parameters. It also controls when to terminate the loop and stop the job. When a valid query is received, the JobOperator flow is triggered to execute the query and write the results accordingly.

**Changes:**

- Introduces a new file, Warmpool.scala, which repeatedly calls `getNextStatement()` in a loop.
- Adds support in `JobOperator` to write the query results either to QueryResultWriter or an OpenSearch Index, depending on the job type.
- Implements the emission of success, failure, and latency metrics within `JobOperator`.

### Check List
- [x] Implemented unit tests

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
@@ -10,6 +10,8 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutor}
 
+import org.opensearch.flint.common.model.FlintStatement
+import org.opensearch.flint.common.scheduler.model.LangType
 import org.opensearch.flint.core.logging.CustomLogging
 import org.opensearch.flint.core.metrics.MetricConstants
 import org.opensearch.flint.core.metrics.MetricsUtil.registerGauge
@@ -58,30 +60,29 @@ object FlintJob extends Logging with FlintJobExecutor {
       if (resultIndexOption.isEmpty) {
         logAndThrow("resultIndex is not set")
       }
-      // https://github.com/opensearch-project/opensearch-spark/issues/138
-      /*
-       * To execute queries such as `CREATE SKIPPING INDEX ON my_glue1.default.http_logs_plain (`@timestamp` VALUE_SET) WITH (auto_refresh = true)`,
-       * it's necessary to set `spark.sql.defaultCatalog=my_glue1`. This is because AWS Glue uses a single database (default) and table (http_logs_plain),
-       * and we need to configure Spark to recognize `my_glue1` as a reference to AWS Glue's database and table.
-       * By doing this, we effectively map `my_glue1` to AWS Glue, allowing Spark to resolve the database and table names correctly.
-       * Without this setup, Spark would not recognize names in the format `my_glue1.default`.
-       */
-      conf.set("spark.sql.defaultCatalog", dataSource)
-      configDYNMaxExecutors(conf, jobType)
 
-      val jobOperator =
-        JobOperator(
-          applicationId,
-          jobId,
-          sparkSession,
+      configDYNMaxExecutors(conf, jobType)
+      val flintStatement =
+        new FlintStatement(
+          "running",
           query,
+          "",
           queryId,
-          dataSource,
-          resultIndexOption.get,
-          jobType,
-          streamingRunningCount,
-          statementRunningCount,
+          LangType.SQL,
+          currentTimeProvider.currentEpochMillis(),
+          Option.empty,
           Map.empty)
+
+      val jobOperator = createJobOperator(
+        sparkSession,
+        applicationId,
+        jobId,
+        flintStatement,
+        dataSource,
+        resultIndexOption.get,
+        jobType,
+        streamingRunningCount,
+        statementRunningCount)
       registerGauge(MetricConstants.STREAMING_RUNNING_METRIC, streamingRunningCount)
       jobOperator.start()
     } else {
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
@@ -8,12 +8,15 @@ package org.apache.spark.sql
 
 import java.util.concurrent.atomic.AtomicInteger
 
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor}
+
 import org.opensearch.flint.core.logging.CustomLogging
 import org.opensearch.flint.core.metrics.MetricConstants
 import org.opensearch.flint.core.metrics.MetricsUtil.registerGauge
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.flint.config.FlintSparkConf
+import org.apache.spark.util.ThreadUtils
 
 /**
  * Spark SQL Application entrypoint
@@ -26,52 +29,71 @@ import org.apache.spark.sql.flint.config.FlintSparkConf
  *   write sql query result to given opensearch index
  */
 object FlintJob extends Logging with FlintJobExecutor {
+  private val streamingRunningCount = new AtomicInteger(0)
+  private val statementRunningCount = new AtomicInteger(0)
+
   def main(args: Array[String]): Unit = {
     val (queryOption, resultIndexOption) = parseArgs(args)
 
     val conf = createSparkConf()
-    val jobType = conf.get("spark.flint.job.type", FlintJobType.BATCH)
-    CustomLogging.logInfo(s"""Job type is: ${jobType}""")
-    conf.set(FlintSparkConf.JOB_TYPE.key, jobType)
-
-    val dataSource = conf.get("spark.flint.datasource.name", "")
-    val query = queryOption.getOrElse(unescapeQuery(conf.get(FlintSparkConf.QUERY.key, "")))
-    if (query.isEmpty) {
-      logAndThrow(s"Query undefined for the ${jobType} job.")
-    }
-    val queryId = conf.get(FlintSparkConf.QUERY_ID.key, "")
-
-    if (resultIndexOption.isEmpty) {
-      logAndThrow("resultIndex is not set")
-    }
-    // https://github.com/opensearch-project/opensearch-spark/issues/138
-    /*
-     * To execute queries such as `CREATE SKIPPING INDEX ON my_glue1.default.http_logs_plain (`@timestamp` VALUE_SET) WITH (auto_refresh = true)`,
-     * it's necessary to set `spark.sql.defaultCatalog=my_glue1`. This is because AWS Glue uses a single database (default) and table (http_logs_plain),
-     * and we need to configure Spark to recognize `my_glue1` as a reference to AWS Glue's database and table.
-     * By doing this, we effectively map `my_glue1` to AWS Glue, allowing Spark to resolve the database and table names correctly.
-     * Without this setup, Spark would not recognize names in the format `my_glue1.default`.
-     */
-    conf.set("spark.sql.defaultCatalog", dataSource)
-    configDYNMaxExecutors(conf, jobType)
-
+    val sparkSession = createSparkSession(conf)
     val applicationId =
       environmentProvider.getEnvVar("SERVERLESS_EMR_VIRTUAL_CLUSTER_ID", "unknown")
     val jobId = environmentProvider.getEnvVar("SERVERLESS_EMR_JOB_ID", "unknown")
+    val isWarmpoolEnabled = conf.get(FlintSparkConf.WARMPOOL_ENABLED.key, "false").toBoolean
+    logInfo(s"isWarmpoolEnabled: ${isWarmpoolEnabled}")
 
-    val streamingRunningCount = new AtomicInteger(0)
-    val jobOperator =
-      JobOperator(
-        applicationId,
-        jobId,
-        createSparkSession(conf),
-        query,
-        queryId,
-        dataSource,
-        resultIndexOption.get,
-        jobType,
-        streamingRunningCount)
-    registerGauge(MetricConstants.STREAMING_RUNNING_METRIC, streamingRunningCount)
-    jobOperator.start()
+    if (!isWarmpoolEnabled) {
+      val jobType = sparkSession.conf.get("spark.flint.job.type", FlintJobType.BATCH)
+      CustomLogging.logInfo(s"""Job type is: ${jobType}""")
+      sparkSession.conf.set(FlintSparkConf.JOB_TYPE.key, jobType)
+
+      val dataSource = conf.get("spark.flint.datasource.name", "")
+      val query = queryOption.getOrElse(unescapeQuery(conf.get(FlintSparkConf.QUERY.key, "")))
+      if (query.isEmpty) {
+        logAndThrow(s"Query undefined for the ${jobType} job.")
+      }
+      val queryId = conf.get(FlintSparkConf.QUERY_ID.key, "")
+
+      if (resultIndexOption.isEmpty) {
+        logAndThrow("resultIndex is not set")
+      }
+      // https://github.com/opensearch-project/opensearch-spark/issues/138
+      /*
+       * To execute queries such as `CREATE SKIPPING INDEX ON my_glue1.default.http_logs_plain (`@timestamp` VALUE_SET) WITH (auto_refresh = true)`,
+       * it's necessary to set `spark.sql.defaultCatalog=my_glue1`. This is because AWS Glue uses a single database (default) and table (http_logs_plain),
+       * and we need to configure Spark to recognize `my_glue1` as a reference to AWS Glue's database and table.
+       * By doing this, we effectively map `my_glue1` to AWS Glue, allowing Spark to resolve the database and table names correctly.
+       * Without this setup, Spark would not recognize names in the format `my_glue1.default`.
+       */
+      conf.set("spark.sql.defaultCatalog", dataSource)
+      configDYNMaxExecutors(conf, jobType)
+
+      val jobOperator =
+        JobOperator(
+          applicationId,
+          jobId,
+          sparkSession,
+          query,
+          queryId,
+          dataSource,
+          resultIndexOption.get,
+          jobType,
+          streamingRunningCount,
+          statementRunningCount,
+          Map.empty)
+      registerGauge(MetricConstants.STREAMING_RUNNING_METRIC, streamingRunningCount)
+      jobOperator.start()
+    } else {
+      // Fetch and execute queries in warm pool mode
+      val warmpoolJob =
+        WarmpoolJob(
+          applicationId,
+          jobId,
+          sparkSession,
+          streamingRunningCount,
+          statementRunningCount)
+      warmpoolJob.start()
+    }
   }
 }
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
@@ -7,12 +7,14 @@ package org.apache.spark.sql
 
 import java.util.Locale
 import java.util.concurrent.ThreadPoolExecutor
+import java.util.concurrent.atomic.AtomicInteger
 
 import com.amazonaws.services.glue.model.{AccessDeniedException, AWSGlueException}
 import com.amazonaws.services.s3.model.AmazonS3Exception
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.apache.commons.text.StringEscapeUtils.unescapeJava
 import org.opensearch.common.Strings
+import org.opensearch.flint.common.model.FlintStatement
 import org.opensearch.flint.core.IRestHighLevelClient
 import org.opensearch.flint.core.logging.{CustomLogging, ExceptionMessages, OperationMessage}
 import org.opensearch.flint.core.metrics.MetricConstants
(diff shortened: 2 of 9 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -26,24 +36,176 @@ import org.apache.spark.sql.flint.config.FlintSparkConf
...
+    val sparkSession = createSparkSession(conf)
+    val applicationId =
+      environmentProvider.getEnvVar("SERVERLESS_EMR_VIRTUAL_CLUSTER_ID", "unknown")
+    val jobId = environmentProvider.getEnvVar("SERVERLESS_EMR_JOB_ID", "unknown")
+    val segmentName = getSegmentName(sparkSession)
+    val warmpoolEnabled = conf.get(FlintSparkConf.WARMPOOL_ENABLED.key, "false").toBoolean
+    logInfo(s"WarmpoolEnabled: ${warmpoolEnabled}")
+
+    if (!warmpoolEnabled) {
+      val jobType = sparkSession.conf.get("spark.flint.job.type", FlintJobType.BATCH)
```
andy-k-improving, line 56: Any particular reason to have the conf key hard-coded here?
We can probably do `FlintSparkConf.JOB_TYPE.key`, which similar to `FlintSparkConf.WARMPOOL_ENABLED.key,` on above.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -26,24 +36,176 @@ import org.apache.spark.sql.flint.config.FlintSparkConf
...
+      CustomLogging.logInfo(s"""Job type is: ${jobType}""")
+      sparkSession.conf.set(FlintSparkConf.JOB_TYPE.key, jobType)
+
+      val dataSource = conf.get("spark.flint.datasource.name", "")
```
andy-k-improving, line 60: Same for `DATA_SOURCE_NAME`

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
saranrajnk: The [existing code](https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala#L33) was doing the same; I just wrapped it in an `if` block. I can modify this if needed.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
saranrajnk: The [existing code](https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala#L37) was doing the same; I just wrapped it in an `if` block. I can modify this if needed.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -26,24 +36,176 @@ import org.apache.spark.sql.flint.config.FlintSparkConf
...
-      logAndThrow(s"Query undefined for the ${jobType} job.")
+    val sparkSession = createSparkSession(conf)
+    val applicationId =
+      environmentProvider.getEnvVar("SERVERLESS_EMR_VIRTUAL_CLUSTER_ID", "unknown")
+    val jobId = environmentProvider.getEnvVar("SERVERLESS_EMR_JOB_ID", "unknown")
+    val segmentName = getSegmentName(sparkSession)
+    val warmpoolEnabled = conf.get(FlintSparkConf.WARMPOOL_ENABLED.key, "false").toBoolean
+    logInfo(s"WarmpoolEnabled: ${warmpoolEnabled}")
+
+    if (!warmpoolEnabled) {
```
ykmr1224, line 55: This introduced huge if/else block, which reduce the readability/maintainability a lot. Can you split the class for warmpool and original interactive job?

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
```diff
@@ -26,24 +36,176 @@ import org.apache.spark.sql.flint.config.FlintSparkConf
...
+          // If exiting with non-zero status, emr-s job will fail.
+          // This is a part of the fault tolerance mechanism to handle such scenarios gracefully
+          System.exit(0)
+        }
+      }
     }
-    val queryId = conf.get(FlintSparkConf.QUERY_ID.key, "")
+  }
+
+  def queryLoop(commandContext: CommandContext, segmentName: String): Unit = {
```
ykmr1224, line 139: This looks similar to FlintREPL.queryLoop method, but modified. It would become very difficult to maintain since we need to very carefully maintain both to be consistent.
Can you add abstraction and avoid duplicates?
Can you add abstraction and avoid duplicates?

ykmr1224 (changes requested):
Can you clarify and document how WarmPool is abstracted and can be enabled/disabled?

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
```diff
@@ -547,6 +547,12 @@ trait FlintJobExecutor {
     }
   }
 
+  def getSegmentName(sparkSession: SparkSession): String = {
+    val maxExecutorsCount =
+      sparkSession.conf.get(FlintSparkConf.MAX_EXECUTORS_COUNT.key, "unknown")
+    String.format("%se", maxExecutorsCount)
+  }
```
noCharger, line 554: This `segmentName` is specific to warmpool logic; let us create abstractions on warmpool and record metrics via AOP.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
```diff
@@ -610,15 +610,15 @@ object FlintREPL extends Logging with FlintJobExecutor {
     }
   }
 
-  private def handleCommandTimeout(
+  def handleCommandTimeout(
```
noCharger, line 613: Both FlintJob and FlintREPL extends FlintJobExecutor, consider refactor common methods to FlintJobExecutor

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -32,7 +32,8 @@ case class JobOperator(
     dataSource: String,
     resultIndex: String,
     jobType: String,
-    streamingRunningCount: AtomicInteger)
+    streamingRunningCount: AtomicInteger,
+    statementContext: Map[String, Any] = Map.empty[String, Any])
```
noCharger, line 36: what's the purpose of adding statementContext which belongs to `FlintStatement` model?

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
noCharger: +1 on this, let's abstract the common interface and move from there.

saranrajnk: Ack

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJob.scala
saranrajnk: This redundancy is expected. After this PR is merged, FlintREPL will be deprecated. FlintJob will be the single point of entry for all types of queries.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala
saranrajnk: Ack

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
saranrajnk: The `getNextStatement` call in FlintJob retrieves all the query information, including the `statementContext`. However, when the query is a streaming or batch query, we need to invoke the `JobOperator`. Currently, the `JobOperator` class only accepts the query, queryId, and other related information, but it does not include the `statementContext`. As a result, when the `JobOperator` calls `executeStatement`, we may encounter issues. To resolve this, the `statementContext` should be passed to the `JobOperator`.

Additionally, before calling [`executeStatement`](https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala#L89), the `FlintStatement` is constructed. Currently, the `FlintStatement` does not include the `statementContext`. However, with the introduction of the warmpool, it becomes necessary to include the `statementContext` in the `FlintStatement`, as there is client-side logic that depends on it.

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -0,0 +1,410 @@
...
+
+  /**
+   * Finalize statement after processing
+   *
+   * @param dataToWrite
+   *   Data to write
+   * @param flintStatement
+   *   Flint statement
+   */
+  private def finalizeCommand(
```
ykmr1224, line 354: Is it duplicate with FlintREPL? Can you check other duplicate and avoid it whenever possible?

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -0,0 +1,410 @@
...
+import org.apache.spark.sql.flint.config.FlintSparkConf
+import org.apache.spark.util.ThreadUtils
+
+object WarmpoolJobConfConstants {
+  val MAPPING_CHECK_TIMEOUT = Duration(1, MINUTES)
+  val DEFAULT_QUERY_WAIT_TIMEOUT_MILLIS = 10 * 60 * 1000
+  val DEFAULT_QUERY_LOOP_EXECUTION_FREQUENCY = 100L
+}
+
+case class WarmpoolJob(
```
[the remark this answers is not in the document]
ykmr1224: It is not a good practice to separate PR for implementation and unit test. It would also become harder to review the unit test separately.

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
ykmr1224: Looks like the only difference is emit metrics and logging. If those are beneficial for Warmpool, should we generalize it and use for FlintREPL as well?

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -0,0 +1,410 @@
...
+        FlintSparkConf.REPL_INACTIVITY_TIMEOUT_MILLIS.key,
+        FlintOptions.DEFAULT_INACTIVITY_LIMIT_MILLIS)
+    val queryWaitTimeoutMillis: Long =
+      conf.getLong("spark.flint.job.queryWaitTimeoutMillis", DEFAULT_QUERY_WAIT_TIMEOUT_MILLIS)
+    val queryLoopExecutionFrequency: Long =
+      conf.getLong(
+        "spark.flint.job.queryLoopExecutionFrequency",
+        DEFAULT_QUERY_LOOP_EXECUTION_FREQUENCY)
+
+    val sessionManager = instantiateSessionManager(sparkSession, resultIndexOption)
```
noCharger, line 57: why do we need session manager for warmpool?

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -0,0 +1,410 @@
...
+        // Exit the JVM to prevent resource leaks and potential emr-s job hung.
+        // A zero status code is used for a graceful shutdown without indicating an error.
+        // If exiting with non-zero status, emr-s job will fail.
+        // This is a part of the fault tolerance mechanism to handle such scenarios gracefully
+        System.exit(0)
+      }
+    }
+  }
+
+  def queryLoop(commandContext: CommandContext): Unit = {
```
noCharger, line 96: why do we need the concept of query loop for warm pool?

noCharger (changes requested):
Can we remove the concept of interactive / batch / streaming job for warm pool?

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
saranrajnk: Warmpool requires multiple iterations as well before running the actual query.

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
saranrajnk: There are bunch of places where sessionManager needs to passed. The resultIndex for interactive query with custom datasource was being read from this sessionManagerImpl. So this has minimal dependency still.

saranrajnk: > Can we remove the concept of interactive / batch / streaming job for warm pool?

Why ? This classification is required for the WP logic. Also, It would be difficult to remove this at this point. Maybe we can pick it up later if required.

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
saranrajnk: As per the recent refactoring, the function was removed from `Warmpool.scala`

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
saranrajnk: In the latest revision, I updated `JobOperator.scala` to handle all types of queries. The branching has been removed from this file, but there are still some jobType checks used in `JobOperator.scala`

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -32,26 +31,40 @@ case class JobOperator(
...
+    } else {
+      statementRunningCount.incrementAndGet()
+    }
 
     // osClient needs spark session to be created first to get FlintOptions initialized.
     // Otherwise, we will have connection exception from EMR-S to OS.
     val osClient = new OSClient(FlintSparkConf().flintOptions())
 
+    // QueryResultWriter depends on sessionManager to fetch the sessionContext
+    val sessionManager = instantiateSessionManager(sparkSession, Some(resultIndex))
```
saranrajnk, line 66: Since `JobOperator` needs to support interactive queries, QueryResultWriter will be used. `QueryResultWriterImpl`, which handles the writing of query results, depends on `sessionManager`.

That's why `sessionManager` is being introduced here to satisfy this dependency (for interactive queries)

Reference: https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/QueryResultWriterImpl.scala#L20

ykmr1224 (approved):

YANG-DB (approved):

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
YANG-DB: @saranrajnk this information is critical 
IMO it would be best adding a markdown doc that explains the job operation sequence and inner details for clarity and future enhancements
thanks

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -32,26 +31,40 @@ case class JobOperator(
...
+
     // TODO: Update FlintJob to Support All Query Types. Track on https://github.com/opensearch-project/opensearch-spark/issues/633
```
noCharger, line 68: nit: remove the line, since this PR is to resolve this issue

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -157,30 +185,29 @@ case class JobOperator(
...
     }
   }
 
   def cleanUpResources(threadPool: ThreadPoolExecutor): Unit = {
-    val isStreaming = jobType.equalsIgnoreCase(FlintJobType.STREAMING)
     try {
-      // Wait for streaming job complete if no error
-      if (!throwableHandler.hasException && isStreaming) {
+      // Wait for job complete if no error
+      if (!throwableHandler.hasException && isStreamingOrBatch) {
```
noCharger, line 196: batch job doesn't have monitor, this logic change will cause job failure

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -32,26 +31,40 @@ case class JobOperator(
...
     with FlintJobExecutor {
 
   // JVM shutdown hook
   sys.addShutdownHook(stop())
+  val isStreamingOrBatch =
+    jobType.equalsIgnoreCase(FlintJobType.STREAMING) || jobType.equalsIgnoreCase(
+      FlintJobType.BATCH)
+  val isWarmpoolEnabled =
+    sparkSession.conf.get(FlintSparkConf.WARMPOOL_ENABLED.key, "false").toBoolean
+  val segmentName = getSegmentName(sparkSession)
```
noCharger, line 47: let's document these conf

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -0,0 +1,122 @@
...
+    var canProceed = true
+
+    try {
+      while (canProceed) {
+        statementExecutionManager.getNextStatement() match {
+          case Some(flintStatement) =>
+            flintStatement.running()
+            statementExecutionManager.updateStatement(flintStatement)
+
+            val jobType = spark.conf.get(FlintSparkConf.JOB_TYPE.key, FlintJobType.BATCH)
```
noCharger, line 70: can we introduce wp as a new job type?

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -0,0 +1,122 @@
...
+        flintStatement.queryId,
+        dataSource,
+        resultIndex,
+        jobType,
+        streamingRunningCount,
+        statementRunningCount,
+        flintStatement.context)
+    jobOperator
+  }
+}
```
noCharger, line 122: duplicate with https://github.com/opensearch-project/opensearch-spark/pull/979/files#diff-6376f87364ce454536e56511495af6644552474c8eaf7be24c3d31d1dbad22ddR72-R86, can we refactor?

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
noCharger: > Additionally, before calling [executeStatement](https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala#L89), the FlintStatement is constructed. Currently, the FlintStatement does not include the statementContext

I think a better idea is to refactor the JobOperator, to not taking fields from FlintStatement

noCharger (commented):
Let's document this WP job somewhere, also the new configurations introduced by WP

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
noCharger: can we leverage https://github.com/opensearch-project/opensearch-spark/blob/98579e11c8b9ec0b2c24eeaef8a05fc90d7162e8/spark-sql-application/src/main/scala/org/apache/spark/sql/SingleStatementExecutionManagerImpl.scala which already takes result index?

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
saranrajnk: Ack, will wrap the monitor logic only for streaming queries

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
saranrajnk: This requires bunch of changes on the Client and API. This will increase the scope of this PR. Will take this up in followup PR

spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
saranrajnk: Ack

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
saranrajnk: Adding this might increase the scope of this PR. I think we can go ahead with the current approach. We can pick this change later

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
saranrajnk: ack

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
saranrajnk: Ack

YANG-DB (approved):

noCharger (commented):
Everything looks good to me, besides adding session manager to flintjob. One design tenant for interfaces is to make the sink agnostic. I would prefer using SingleStatementExecutionManagerImpl.scala in this refactor to decouple result index and session manager. But since there's already 2 approvals, feel free to merge it.

saranrajnk:
> Everything looks good to me, besides adding session manager to flintjob. One design tenant for interfaces is to make the sink agnostic. I would prefer using SingleStatementExecutionManagerImpl.scala in this refactor to decouple result index and session manager. But since there's already 2 approvals, feel free to merge it.

I don't have permissions to merge. Can someone merge please?

noCharger:
> > Everything looks good to me, besides adding session manager to flintjob. One design tenant for interfaces is to make the sink agnostic. I would prefer using SingleStatementExecutionManagerImpl.scala in this refactor to decouple result index and session manager. But since there's already 2 approvals, feel free to merge it.
> 
> I don't have permissions to merge. Can someone merge please?

Let's create a issue for the fast follow up and linked here.

saranrajnk:
> > > Everything looks good to me, besides adding session manager to flintjob. One design tenant for interfaces is to make the sink agnostic. I would prefer using SingleStatementExecutionManagerImpl.scala in this refactor to decouple result index and session manager. But since there's already 2 approvals, feel free to merge it.
> > 
> > 
> > I don't have permissions to merge. Can someone merge please?
> 
> Let's create a issue for the fast follow up and linked here.

Created an issue - https://github.com/opensearch-project/opensearch-spark/issues/1050

Issue #956, opened 2024-11-28
[PPL-Lang]add `appendcol` ppl command 


YANG-DB:
**Is your feature request related to a problem?**
This feature request proposes the addition of an `appendcol`- command to OpenSearch’s Piped Processing Language (PPL). The `appendcol` command (modeled after Splunk’s command of the same name), allows users to append the result of one or more searches as additional columns to the existing search results, offering a powerful mechanism to enrich data by combining results horizontally.

**What solution would you like?**
The following query:
```sql
source=logs | stats count by status | appendcol [ stats avg(response_time) by status ]
```

In this example the `appendcol` command adds a new column containing the average response time for each status alongside the original column that shows the count by status. The original row structure is preserved.

_Implementation could probably use  rewriting into a join query._

**Add Proposal Document**

In many use cases, users need to compare or merge multiple result sets side-by-side without altering the existing row structure. The `appendcol` command will become a used feature for horizontally merging datasets, and bringing this feature to OpenSearch PPL would greatly enhance data processing flexibility.

By adding `appendcol` to PPL, users can:

* Combine different queries into a unified result set without duplicating rows.
* Enrich an existing dataset with additional metrics or fields from other searches.
* Improve the efficiency and readability of complex queries.

### **Technical Details**

**Syntax**
 The `appendcol` command would accept a query inside square brackets `[ ]`, representing the additional pipeline that produces the column(s) to append to the original result set.

**Behavior**
 The new column(s) would be aligned with the rows of the original dataset based on their order of appearance. Each appended column must produce the same number of rows as the base dataset to ensure proper alignment. Any discrepancies in row counts could result in null values for mismatched rows.

andy-k-improving:
@YANG-DB Since you mentioned the Splunk reference, I presume this is the the page:  https://docs.splunk.com/Documentation/SplunkCloud/latest/SearchReference/Appendcols
Can you confirm?

Also a few points on below would be great if you can clarify?
- Query that placed under the square brackets `[ ]` is a subQuery that based on the previous step of the pipe, so there will never be any additional dataSet || table be introduced by this `appendcol` command? 

Consider a table with structure:
```
Table name: logs
Columns:
 - IP: String 
 - status : Int
 - response_time : Int

```
Then:

```
source=logs | stats count by status | appendcol [ stats avg(response_time) by status ]
```
Is valid because `response_time` is part of the `source=logs`.


```
[...]

PR #990, opened 2024-12-11
PPL command implementation for `appendCol`

andy-k-improving:
### Description
Introduce the new PPL command `appendCol` which aim to aggregate result from multiple searches into a single comprehensive table for user to view.

This is accomplished by reading both main-search and sub-search in the form of `node` then transform it into the following of SQL with by adding `_row_number_` for the dataset's natural order, then join both main and sub search with the `_row_number_` column.

```
select t1.*, t2.* 

FROM (
     SELECT *, row_number() over (order by '1') as row_org 
     FROM employees) as t1

 FULL OUTER JOIN (
     SELECT *, row_number() over (order by '1') as row_app 
     FROM employees) as t2 

ON t1.row_org = t2.row_app;
```

### Related Issues
Resolves: https://github.com/opensearch-project/opensearch-spark/issues/956

### Check List
- [x] Implemented unit tests

## Test plan:

```
# Produce the artifact
sbt clean sparkPPLCosmetic/publishM2

# Start Spark with the plugin
bin/spark-sql --jars "/ABSOLUTE_PATH_TO_ARTIFACT/opensearch-spark-ppl_2.12-0.6.0-SNAPSHOT.jar" \
--conf "spark.sql.extensions=org.opensearch.flint.spark.FlintPPLSparkExtensions"  \
--conf "spark.sql.catalog.dev=org.apache.spark.opensearch.catalog.OpenSearchCatalog" \
--conf "spark.hadoop.hive.cli.print.header=true"

# Insert test table and data
CREATE TABLE employees (name STRING, dept STRING, salary INT, age INT, con STRING);

INSERT INTO employees VALUES ("Lisa", "Sales", 10000, 35, 'test');
INSERT INTO employees VALUES ("Evan", "Sales", 32000, 38, 'test');
INSERT INTO employees VALUES ("Fred", "Engineering", 21000, 28, 'test');
INSERT INTO employees VALUES ("Alex", "Sales", 30000, 33, 'test');
INSERT INTO employees VALUES ("Tom", "Engineering", 23000, 33, 'test');
INSERT INTO employees VALUES ("Jane", "Marketing", 29000, 28, 'test');
INSERT INTO employees VALUES ("Jeff", "Marketing", 35000, 38, 'test');
INSERT INTO employees VALUES ("Paul", "Engineering", 29000, 23, 'test');
INSERT INTO employees VALUES ("Chloe", "Engineering", 23000, 25, 'test');

# Append one sub-search:

source=employees | FIELDS name, dept, salary | APPENDCOL  [ stats count() as event_count];

name	dept	salary	event_count
Lisa	Sales------	10000	9
Fred	Engineering	21000	NULL
Paul	Engineering	29000	NULL
Evan	Sales------	32000	NULL
Chloe	Engineering	23000	NULL
Tom	Engineering	23000	NULL
Alex	Sales	30000	NULL
Jane	Marketing	29000	NULL
Jeff	Marketing	35000	NULL


# Append multiple sub-searches:

source=employees | FIELDS name, dept, salary | APPENDCOL  [ stats count() as event_count] | APPENDCOL [stats avg(age) as avg_age];

name	dept	salary	event_count	avg_age
Lisa	Sales------	10000	9	31.22222222222222
Fred	Engineering	21000	NULL	NULL
Paul	Engineering	29000	NULL	NULL
Evan	Sales------	32000	NULL	NULL
Chloe	Engineering	23000	NULL	NULL
Tom	Engineering	23000	NULL	NULL
Alex	Sales	30000	NULL	NULL
Jane	Marketing	29000	NULL	NULL
Jeff	Marketing	35000	NULL	NULL



# With override option (`salary` column from the main-search is being dropped and replaced by the `salary` column over the sub-search)

source=employees | FIELDS name, dept, salary | APPENDCOL OVERRIDE=true [stats avg(salary) as salary];

name	dept	salary
Lisa	Sales------	25777.777777777777
Fred	Engineering	NULL
Paul	Engineering	NULL
Evan	Sales------	NULL
Chloe	Engineering	NULL
Tom	Engineering	NULL
Alex	Sales	NULL
Jane	Marketing	NULL
Jeff	Marketing	NULL


```


docs/ppl-lang/ppl-appendcol-command.md
```diff
@@ -0,0 +1,79 @@
+## PPL `appendcol` command
+
+### Description
+Using `appendcol` command to append the result of a sub-search and attach it alongside with the input search results (The main search).
+
+### Syntax - APPENDCOL
+`APPENDCOL <override=?> [sub-search]...`
+
+* <override=?>: optional boolean field to specify should result from main-result be overwritten in the case of column name conflict.
+* sub-search: The PPL commands to execute for the secondary search.
```
YANG-DB, line 10: please state that the sub search uses the same data `source`

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
```diff
@@ -0,0 +1,121 @@
+package org.opensearch.sql.ppl.utils;
```
YANG-DB, line 1: plz add license header

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
```diff
@@ -0,0 +1,121 @@
...
+     * @param node The Node object that this util method search upon.
+     * @return The first occurrence of Relation object from the given Node.
+     */
+    static Relation retrieveRelationClause(Node node) {
+        while (node != null) {
+            if (node instanceof Relation) {
+                return (Relation) node;
+            } else {
+                try {
+                    node = node.getChild().get(0);
```
YANG-DB, line 66: use `visitFirstChild(node, context)` here as well

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+    /**
+     * Util method to perform analyzed() call against the given LogicalPlan to extract all fields
+     * that will be projected upon the execution in the form of Java List with user provided schema prefix.
+     * @param lp LogicalPlan instance to extract the projection fields from.
+     * @param tableName the table || schema name being appended as part of the returned fields.
+     * @return A list of Expression instances with alternated tableName || Schema information.
+     */
+    static List<Expression> getoverridedlist(LogicalPlan lp, String tableName) {
```
YANG-DB, line 86: rename `getoverridedlist` -> `getOverridedList`

YANG-DB (changes requested):

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
andy-k-improving: I have double checked the logic, but I don't think that is feasible to replace it with `visitFirstChild`, because at this point the `node` of sub-search has NO relation clause, which is not a valid query and exception will be thrown upon any `visit( )` method. 
With that said thou, do you foresee any alternative way to convert a Node into a LogicalPlan instance without any `source = xxx` clause?

LantaoJin:
Two high level questions:
1. appendCol command syntax is 
```APPENDCOL <override=?> [sub-search]...```

But the sub-search syntax is https://github.com/opensearch-project/opensearch-spark/blob/957de4e27d3a243ccb9beccd2ee72ab231e9dcf4/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L27-L29
Seems this PR doesn't follow the same syntax.

2. why the result of query `source=employees | FIELDS name, dept, salary | APPENDCOL  [ stats count() as event_count]`
is 
```
name	dept	salary	event_count
Lisa	Sales------	10000	9
Fred	Engineering	21000	NULL
Paul	Engineering	29000	NULL
Evan	Sales------	32000	NULL
Chloe	Engineering	23000	NULL
Tom	Engineering	23000	NULL
Alex	Sales	30000	NULL
Jane	Marketing	29000	NULL
Jeff	Marketing	35000	NULL 
```
instead of 
```
name	dept	salary	event_count
Lisa	Sales------	10000	9
Fred	Engineering	21000	9
Paul	Engineering	29000	9
Evan	Sales------	32000	9
Chloe	Engineering	23000	9
Tom	Engineering	23000	9
Alex	Sales	30000	9
Jane	Marketing	29000	9
Jeff	Marketing	35000	9
```

And how is the result of query `source=employees | stats sum(salary) as total_salary by dept | appendcol [  stats avg(age) as avg_age by dept ]`?

File: ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java [hunk 5]
```diff
@@ -253,6 +258,54 @@ public LogicalPlan visitTrendline(Trendline node, CatalystPlanContext context) {
...
+
+            context.retainAllNamedParseExpressions(p -> p);
+            context.retainAllPlans(p -> p);
+
+            // Join both Main and Sub search with _ROW_NUMBER_ column
+            LogicalPlan joinedQuery = join(
+                    mainSearchWithRowNumber, subSearchWithRowNumber,
+                    Join.JoinType.LEFT,
+                    Optional.of(new EqualTo(t1Attr, t2Attr)),
+                    new Join.JoinHint());
```
LantaoJin, line 295: Now I know why you got NULL in the example.
I think this is not what we expected. How about use `inner` and `cross` join together.
For example:
1. with same group-by key, convert it to join key in inner
```
source=employees | stats sum(salary) as total_salary by dept | appendcol [  stats avg(age) as avg_age by dept ]
```
2. without group-by key, use a cross join
```
source=employees | stats sum(salary) as total_salary by dept | appendcol [  stats avg(age) as avg_age ]
```
The implementation equals to
```
def appendCols(mainSearchDF: DataFrame, subSearchDF: DataFrame, joinKey: Option[String] = None): DataFrame = {
  joinKey match {
    case Some(key) =>
      // If a join key is provided, perform a join
      mainSearchDF.join(subSearchDF, Seq(key), "inner")
    case None =>
      // If no join key is provided, assume a global aggregation and use a cross join
      mainSearchDF.crossJoin(subSearchDF)
  }
}
```

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
```diff
@@ -0,0 +1,121 @@
...
+        // When override option present, extract fields to project from sub-search,
+        // then apply a dfDropColumns on main-search to avoid duplicate fields.
+        final SparkSession sparkSession = SparkSession.getActiveSession().get();
```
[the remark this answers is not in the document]
andy-k-improving: @LantaoJin Would you mind to point me the specific doc for this purpose, I have looked up the implementation and the doc for `eval` but seems all examples are rely on the user to supply the name of the columns to override.

However in this particular case (AppendCol), I need to first figure out the exact fields which sub-query will project (Including resolve `select *` to actual columns name), and only then I can take that list and apply a `dropDFColumn( )` on the main query.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: @YANG-DB , @LantaoJin the comment's seems to contradict what we have discussed over https://github.com/opensearch-project/opensearch-spark/issues/956 and the [Splunk doc](https://docs.splunk.com/Documentation/SplunkCloud/latest/SearchReference/Appendcols), hence would that be possible to clarify?

### Join condition
By referring to the prior discussion on Github issue and the Spunk doc, I don't think that is the case, as user won't be asked to enter the neither the join column or the join condition, hence it's up the code logic to generate the `natural_oder` for the dataFrame and use it to join, Can I confirm this?

I'm asking this because the above comment seems to indicate user will provide a the name of column for the join (Both inner || cross) join, but that is not the case the for the method signature || grammar of `appenCol( )`


### Expected result 

Accordingly the Splunk doc

```
Appends the fields of the [subsearch](https://docs.splunk.com/Splexicon:Subsearch) results with the input search results. 
All fields of the subsearch are combined into the current results, 
with the exception of [internal fields](https://docs.splunk.com/Splexicon:Internalfield). 
For example, the first subsearch result is merged with the first main result, 
the second subsearch result is merged with the second main result, and so on.
```

Both `main` and the `sub` are simply being joined row by row from top the bottom, 1:1 without explicit joining condition, and in the case of row.length( ) different between two dataFrame, `null` will be used to fill the gap, and this seems to align with described on Github issue.  

From Github issue:
```
Behavior
The new column(s) would be aligned with the rows of the original dataset based on their order of appearance.
Each appended column must produce the same number of rows as the base dataset to ensure proper alignment. 
Any discrepancies in row counts could result in null values for mismatched rows.
```

Hence I don't think neither `cross` and `inner` join will do the same, as rows will be truncated in both scenario.
And I reckon this is what make `appendCol` as standalone implementation, or else user can simply achieved by existing `subQuery` or `LookUp` command?

andy-k-improving:
> Two high level questions:
> 
> 1. appendCol command syntax is
>    `APPENDCOL <override=?> [sub-search]...`
> 
> And the sub-search syntax is
> 
> https://github.com/opensearch-project/opensearch-spark/blob/957de4e27d3a243ccb9beccd2ee72ab231e9dcf4/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L27-L29
> 
> 
> Seems this PR doesn't follow the sub-search syntax.
> I prefer to follow the current sub-search syntax, in case we could combine columns from different tables. for examples:
> `source=employees | FIELDS name, dept, salary | APPENDCOL [ search source = company | stats count() as event_count ]`
> But if this is intentional (appendcol only works for one table), I am okey for current syntax.
> 2. why the result of query `source=employees | FIELDS name, dept, salary | APPENDCOL  [ stats count() as event_count]`
>    is
> 
> ```
> name	dept	salary	event_count
> Lisa	Sales------	10000	9
> Fred	Engineering	21000	NULL
> Paul	Engineering	29000	NULL
> Evan	Sales------	32000	NULL
> Chloe	Engineering	23000	NULL
> Tom	Engineering	23000	NULL
> Alex	Sales	30000	NULL
> Jane	Marketing	29000	NULL
> Jeff	Marketing	35000	NULL 
> ```
> 
> instead of
> 
> ```
> name	dept	salary	event_count
> Lisa	Sales------	10000	9
> Fred	Engineering	21000	9
> Paul	Engineering	29000	9
> Evan	Sales------	32000	9
> Chloe	Engineering	23000	9
> Tom	Engineering	23000	9
> Alex	Sales	30000	9
> Jane	Marketing	29000	9
> Jeff	Marketing	35000	9
> ```
> 
> PS, what is the expected result of query `source=employees | stats sum(salary) as total_salary by dept | appendcol [ stats avg(age) as avg_age by dept ]`?

Yep, the sub-search under `appendcol` is restricted to use the same dataSource as the main PPL command, for the expected result, we can probably discuss it on the other thread, in order to centralise the convo.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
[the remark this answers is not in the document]
LantaoJin: > However in this particular case (AppendCol), I need to first figure out the exact fields which sub-query will project (Including resolve select * to actual columns name).

I have an idea. We could add a restriction to the sub-search clause for `AppendCol`
1. For case with fields or aliases, use the fields directly, for example
```
source=employees | FIELDS name, dept, age | APPENDCOL OVERRIDE=true [ ... | FIELDS foo, bar, age ];
```
Adding `foo`, `bar`, `age` to `fieldsToRemove`.
```
source=employees | FIELDS name, dept, age | APPENDCOL OVERRIDE=true [ stats avg(age) as age, count(name) as cnt ];
```
Adding `age`, `cnt` to `fieldsToRemove`.
2. For case with `allFields`. throw an exception `APPENDCOL should specify the output fields`:
```
source=employees | FIELDS name, dept, age | APPENDCOL OVERRIDE=true [ where age > 0 ];
```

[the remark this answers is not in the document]
andy-k-improving: That makes sense; it also eliminates the need for a live Spark session during query analysis time.
Will documentation the limitation (Exception being thrown when select * present && override==true) on doc and update the code logic accordingly.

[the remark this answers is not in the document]
YANG-DB: yes its much simpler - we need to document these assumptions in the command spec

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: @YANG-DB  @LantaoJin Thanks for the clarification, that make much more sense now!
If I follow your thought on above logic, that means the code logic need to somehow know how many rows will be returned from both main and sub search.

As it will need such number, in order to determine whether cross join || inner being applied, however I don't think such info is available during the visitor steps, when compositing the Spark's logical plan.

Is there any way I could predict such number or some reasonable assumption can be made? 
 - Ex: if both main and sub search has identical group by field, then the output rows should match.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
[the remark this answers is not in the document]
andy-k-improving: Done.

docs/ppl-lang/README.md
```diff
@@ -74,6 +74,8 @@ For additional examples see the next [documentation](PPL-Example-Commands.md).
   
     - [`expand commands`](ppl-expand-command.md)
 
+    - [`appendcol commands`](ppl-appendcol-command.md)
```
currantw, line 77: Nit. Should this be singular (i.e. "appencol command")? Same for join, lookup, subquery, correlation, trendline, and expand above as well.

docs/ppl-lang/ppl-appendcol-command.md
```diff
@@ -0,0 +1,83 @@
...
+Using `appendcol` command to append the result of a sub-search and attach it alongside with the input search results (The main search).
+
+### Syntax - APPENDCOL
+`APPENDCOL <override=?> [sub-search]...`
+
+* <override=?>: optional boolean field to specify should result from main-result be overwritten in the case of column name conflict.
+* sub-search: Executes PPL commands as a secondary search. The sub-search uses the same data specified in the source clause of the main search results as its input.
+
+
+#### Example 1: To append the result of `stats ave(age) as AVG_AGE` into existing search result
```
currantw, line 13:
```suggestion
#### Example 1: To append the result of `stats avg(age) as AVG_AGE` into existing search result   
```

docs/ppl-lang/ppl-appendcol-command.md
```diff
@@ -0,0 +1,83 @@
...
+### Syntax - APPENDCOL
+`APPENDCOL <override=?> [sub-search]...`
+
+* <override=?>: optional boolean field to specify should result from main-result be overwritten in the case of column name conflict.
+* sub-search: Executes PPL commands as a secondary search. The sub-search uses the same data specified in the source clause of the main search results as its input.
+
+
+#### Example 1: To append the result of `stats ave(age) as AVG_AGE` into existing search result   
+
+The example append the result of sub-search `stats ave(age) as AVG_AGE` alongside with the main-search.
```
currantw, line 15:
```suggestion
The example append the result of sub-search `stats avg(age) as AVG_AGE` alongside with the main-search.
```

docs/ppl-lang/ppl-appendcol-command.md
```diff
@@ -0,0 +1,83 @@
...
+    | Lisa | Sales------ |  35 | 31.2222222222222|  
+    | Fred | Engineering |  28 | NULL            |  
+    | Paul | Engineering |  23 | NULL            |  
+    | Evan | Sales------ |  38 | NULL            |   
+    | Chloe| Engineering |  25 | NULL            |   
+    | Tom  | Engineering |  33 | NULL            |   
+    | Alex | Sales       |  33 | NULL            |  
+    | Jane | Marketing   |  28 | NULL            |  
+    | Jeff | Marketing   |  38 | NULL            |  
+    +------+-------------+-----+------------------+
```
currantw, line 33: Are these extra dashes supposed to be here...? Same for other examples below.
Nit. Might as well align the dividers if you're gonna be updating this anyway 🤷 

```suggestion
    +------+-------------+-----+------------------+  
    | name | dept        | age | AVG_AGE          |  
    +------+-------------+-----+------------------+  
    | Lisa | Sales       |  35 | 31.2222222222222 |  
    | Fred | Engineering |  28 | NULL             |  
    | Paul | Engineering |  23 | NULL             |  
    | Evan | Sales       |  38 | NULL             |   
    | Chloe| Engineering |  25 | NULL             |   
    | Tom  | Engineering |  33 | NULL             |   
    | Alex | Sales       |  33 | NULL             |  
    | Jane | Marketing   |  28 | NULL             |  
    | Jeff | Marketing   |  38 | NULL             |  
    +------+-------------+-----+------------------+  
```

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -275,6 +277,10 @@ trendlineType
    | WMA
    ;
 
+appendcolCommand
+   : APPENDCOL (OVERRIDE EQUAL override = booleanLiteral)? LT_SQR_PRTHS commands (PIPE commands)* RT_SQR_PRTHS
```
currantw, line 281: Is there some extra text here?

```suggestion
   : APPENDCOL (OVERRIDE EQUAL booleanLiteral)? LT_SQR_PRTHS commands (PIPE commands)* RT_SQR_PRTHS
```

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/tree/AppendCol.java
```diff
@@ -0,0 +1,55 @@
...
+@ToString
+@EqualsAndHashCode(callSuper = false)
+@AllArgsConstructor
+public class AppendCol extends UnresolvedPlan {
+
+    public boolean override;
+
+    private UnresolvedPlan subSearch;
+
+    private UnresolvedPlan child;
```
currantw, line 33: Can these two attributes be `final`?

```suggestion
    public final boolean override;

    private final UnresolvedPlan subSearch;

    private UnresolvedPlan child;
```

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -253,6 +259,61 @@ public LogicalPlan visitTrendline(Trendline node, CatalystPlanContext context) {
...
+        final String TABLE_RHS = "APPENDCOL_T2";
+        final UnresolvedAttribute t1Attr = new UnresolvedAttribute(seq(TABLE_LHS, APPENDCOL_ID));
+        final UnresolvedAttribute t2Attr = new UnresolvedAttribute(seq(TABLE_RHS, APPENDCOL_ID));
+        final List<Expression> fieldsToRemove = new ArrayList<>(List.of(t1Attr, t2Attr));
+        final Node subSearchNode = node.getSubSearch();
+
+        // Apply an additional projection layer on main-search to provide natural order.
+        LogicalPlan mainSearch = visitFirstChild(node, context);
+        var mainSearchWithRowNumber = AppendColCatalystUtils.getRowNumStarProjection(context, mainSearch, TABLE_LHS);
+        context.withSubqueryAlias(mainSearchWithRowNumber);
```
currantw, line 276: I don't have much experience in Java. Is there any guidance/conventions around when to use `var` vs. an explicit type?

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: @LantaoJin Would you mind to have a look on above for the join condition and advise?

Thanks :)

docs/ppl-lang/README.md
andy-k-improving: I reckon we will need to review this on case by case basis, in `appendCol`, that make sense to be singular. 
But in the case of Join, Trendline...... which each command has some variety, that make sense to be plural.
Ex: Trendline has `SMA` and `WMA`, 
Subquery has `InSubquery`, `ExistsSubquery` and `ScalarSubquery`....

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: `var` usually being used in order to shorten the length of the line to make the code more readable, in the case of the complex type.

Ex: List<SomeComplexLengthlyType<SomeMoreInnerType>> = xxxx

I reckon the convention is to keep each line below 150 characters max, and will be checked via JaCoCo.

However in this particular case, I don't think we have this rule enforced during the compilation, hence no compilation will be thrown with || without `var` in this case.

ppl-spark-integration/src/main/java/org/opensearch/sql/ast/tree/AppendCol.java
andy-k-improving: Agree, as both variables should be effective final, I have updated the visiter class to use setter, instead of direct access.

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
andy-k-improving: Would you mind to elaborate more here, indeed I need to read the override option from the user.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
currantw: Cool, thanks

docs/ppl-lang/README.md
currantw: Alright, happy to leave as is

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
currantw: Perhaps I'm not reading the parser syntax correctly...but you currently have override (`OVERRIDE` and `override`) and equals (`EQUAL` and `=`) twice here, right? Can it not just be `OVERRIDE EQUAL booleanLiteral`?

andy-k-improving: Not really, in this case `OVERRIDE EQUAL` is used to match the wording ` override =` from user input only, but no variable || data will be stored, it only enforce the grammar || syntax check.

With `override = booleanLiteral` on the other side is an instruction set for ANLTR to: 
 * Read a `booleanLiteral` and then store it with the name `override` for later usage. 


So by combining both, `OVERRIDE EQUAL override = booleanLiteral` will first check the syntax and then store the boolean option for the AppenCol command use.

currantw: > Perhaps I'm not reading the parser syntax correctly

Clearly this was correct 😆 

Thanks for the explanation!

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: @YANG-DB @LantaoJin By following up on this point, I executed the same appendCols query on Splunk to confirm the behaviour (with the same dataset) and this is what I got:

Splunk query:
```
index=test 
| table name, dept, age
| rename age as age_original 
| appendcols [ search index=test | where age > 30 | rename age as age_above_30 | table age_above_30 ]
```

Result:
image

Hence indeed left join is being used on the reference implementation, which both tables will be joined with its natural order, and null || empty will be placed in the case of mis-alignment.


Hence can I have the confirmation on this, which should we keep the `left-join` approach, or we should switch to `inner || cross` instead? 
As by switching to `inner || cross` will result in different output behaviour compare to the reference impl.

YANG-DB: > @YANG-DB @LantaoJin By following up on this point, I executed the same appendCols query on Splunk to confirm the behaviour (with the same dataset) and this is what I got:
> 
> Splunk query:
> ```
> index=test 
> | table name, dept, age
> | rename age as age_original 
> | appendcols [ search index=test | where age > 30 | rename age as age_above_30 | table age_above_30 ]
> ```
> 
> Result:
> image
> 
> Hence indeed left join is being used on the reference implementation, which both tables will be joined with its natural order, and null || empty will be placed in the case of mis-alignment.
> 
> 
> Hence can I have the confirmation on this, which should we keep the `left-join` approach, or we should switch to `inner || cross` instead? 
> As by switching to `inner || cross` will result in different output behaviour compare to the reference impl.
> 
> 
> 

Thanks for the deep dive into the feature comparison 
I don't have a strong opinion on this so we can go either way - @LantaoJin - up to you I guess ;)

andy-k-improving: @LantaoJin Hi, would you mind to advise on this?
Thanks,

YANG-DB:
> > Two high level questions:
> > 
> > 1. appendCol command syntax is
> >    `APPENDCOL <override=?> [sub-search]...`
> > 
> > And the sub-search syntax is
> > https://github.com/opensearch-project/opensearch-spark/blob/957de4e27d3a243ccb9beccd2ee72ab231e9dcf4/ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4#L27-L29
> > 
> > Seems this PR doesn't follow the sub-search syntax.
> > I prefer to follow the current sub-search syntax, in case we could combine columns from different tables. for examples:
> > `source=employees | FIELDS name, dept, salary | APPENDCOL [ search source = company | stats count() as event_count ]`
> > But if this is intentional (appendcol only works for one table), I am okey for current syntax.
> > 2. why the result of query `source=employees | FIELDS name, dept, salary | APPENDCOL  [ stats count() as event_count]`
> > is
> > ```
> > name	dept	salary	event_count
> > Lisa	Sales------	10000	9
> > Fred	Engineering	21000	NULL
> > Paul	Engineering	29000	NULL
> > Evan	Sales------	32000	NULL
> > Chloe	Engineering	23000	NULL
> > Tom	Engineering	23000	NULL
> > Alex	Sales	30000	NULL
> > Jane	Marketing	29000	NULL
> > Jeff	Marketing	35000	NULL 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of
> > ```
> > name	dept	salary	event_count
> > Lisa	Sales------	10000	9
> > Fred	Engineering	21000	9
> > Paul	Engineering	29000	9
> > Evan	Sales------	32000	9
> > Chloe	Engineering	23000	9
> > Tom	Engineering	23000	9
> > Alex	Sales	30000	9
> > Jane	Marketing	29000	9
> > Jeff	Marketing	35000	9
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > PS, what is the expected result of query `source=employees | stats sum(salary) as total_salary by dept | appendcol [ stats avg(age) as avg_age by dept ]`?
> 
> Yep, the sub-search under `appendcol` is restricted to use the same dataSource as the main PPL command, for the expected result, we can probably discuss it on the other thread, in order to centralise the convo.

@andy-k-improving I would go with the same behaviour to match with  Splunk ...

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
LantaoJin: @andy-k-improving I think you are right. Seems `appendcols` in Spunk doesn't perform any conditional join. It doesn't make sure the corresponding rows could be joined together with their equivalent values. Your original design (left join by row_number) is the correct approach!

LantaoJin: For example, there are 3 tables:
table1 (col1 INT, col2 STRING)
1, a
2, b
3, c

table2 (col3 INT, col4 STRING)
1, aa
2, bb
3, cc

table3 (col5 INT, col6 STRING)
1, aaa
4, bbb
5, ccc
6, ddd

`search table1 | appendcol [search table2 | fields col4]` returns
1, a, aa
2, b, bb
3, c, cc

`search table1 | appendcol [search table3 | fields col6]` returns
1, a, aaa
2, b, bbb
3, c, ccc
NULL, NULL, ddd <--- please check this behaviours in splunk, LEFT JOIN will remove this line, but FULL OUTER will contains this line.

YANG-DB (approved):

YANG-DB:
@andy-k-improving LGTM - please validate @LantaoJin question and respond here
thanks

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: @LantaoJin I can confirm this should be a FULL OUTER join instead. 
Will update the logic and test-case to reflect this. 
image

andy-k-improving: @LantaoJin I have updated the logic and added additional test-cases to assert on integ-test.
Can you have another look and advise?

docs/ppl-lang/ppl-appendcol-command.md
```diff
@@ -0,0 +1,83 @@
...
+Using `appendcol` command to append the result of a sub-search and attach it alongside with the input search results (The main search).
+
+### Syntax - APPENDCOL
+`APPENDCOL <override=?> [sub-search]...`
+
+* <override=?>: optional boolean field to specify should result from main-result be overwritten in the case of column name conflict.
+* sub-search: Executes PPL commands as a secondary search. The sub-search uses the same data specified in the source clause of the main search results as its input.
+
+
+#### Example 1: To append the result of `stats avg(age) as AVG_AGE` into existing search result
```
LantaoJin, line 13: We need to add an additional example here and a test in IT, which is the most commonly used scenario for the AppenCol command:
`source=employees | stats avg(age) as avg_age1 by dept | APPENDCOL [ stats avg(age) as avg_age2 by dept ]`
The result should look like 
```
dept             avg_age1 avg_age2
Sales            35.3333  35.3333
Engineering      27.25    27.25
Marketing        33       33
```

LantaoJin: LGTM if above comment addressed.

LantaoJin (changes requested):

docs/ppl-lang/ppl-appendcol-command.md
LantaoJin: Besides, please update https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/PPL-Example-Commands.md as well.

docs/ppl-lang/ppl-appendcol-command.md
```diff
@@ -0,0 +1,100 @@
...
+    +------+-------------+-----+------------------+  
+
+
+#### Example 2: To compare multiple stats commands with side by side with appendCol.
+
+This example demonstrates a common use case: performing multiple statistical calculations and displaying the results side by side in a horizontal layout.
+
+PPL query:
+
+    os> source=employees | stats avg(age) as avg_age1 by dept | fields dept, avg_age1 | APPENDCOL  [ stats avg(age) as avg_age2 by dept | fields avg_age2 ];
```
LantaoJin, line 42: What happens without `| fields avg_age2`?

In Splunk, command `source=employees | stats avg(age) as avg_age1 by dept | APPENDCOLS [ stats avg(age) as avg_age2 by dept ]` could display the same outputs. Does current implementation have to add those `fields` to get the same outputs?

docs/ppl-lang/ppl-appendcol-command.md
andy-k-improving: Done.

docs/ppl-lang/ppl-appendcol-command.md
andy-k-improving: The current implementation in this PR don't mandate an projection clause like `fields`, and I'm adding this due to discrepancy on the default behaviour of option `override`.

In Splunk, columns override happens even `override=false` or absent, when this occur, fields that present main search will override the column with same name on sub-search (Main > Sub), and it's negate when `override=true` (Sub > Main), duplicated fields will be discarded and only the precedence one will be displayed. 


However this is not the current behaviour of this PR, as we discussed earlier, the code logic don't proactively analyse the prospect logical plan for neither `main` nor `sub`, to figure out the exact columns being projected on the result query, and hence when the `override` option is absent on this PR, the query will result in 4 columns, rather than 3, which are:
Ex: `source=employees | stats avg(age) as avg_age1 by dept | APPENDCOL  [ stats avg(age) as avg_age2 by dept ];` 

1.  avg_age1 (Main)
2. dept (Main)
3. avg_age2 (Sub)
4. dept (Sub)

```
+--------------------+-------------+--------------------+-------------+
|     avg_age1       |    dept     |     avg_age2       |    dept     |
+--------------------+-------------+--------------------+-------------+
|       27.25        | Engineering |       27.25        | Engineering |
| 35.333333333333336 |    Sales    | 35.333333333333336 |    Sales    |
|        33.0        |  Marketing  |        33.0        |  Marketing  |
+--------------------+-------------+--------------------+-------------+
```


Compare to what on Splunk, which the `dept` from main search is being used and the `dept` from sub-search be discarded. 
image

To get back to the original question, the statistic will always be accurate, but the number of columns will be differ, in the case of duplicated columns with `override=false`.

LantaoJin: I see. Could you illustrate in docs/ppl-lang/ppl-appendcol-command.md as well? I think it's helpful for migration.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
```diff
@@ -0,0 +1,107 @@
...
+                .collect(Collectors.toList());
+        final Relation table = new Relation(unresolvedExpressionList);
+        while (subSearch != null) {
+            try {
+                subSearch = subSearch.getChild().get(0);
+            } catch (NullPointerException ex) {
+                ((UnresolvedPlan) subSearch).attach(table);
+                break;
+            }
+        }
```
LantaoJin, line 65: Can you rewrite this `try-catch` to
```java
while (subSearch != null) {
    List<? extends Node> children = subSearch.getChild();
    if (children == null || children.isEmpty()) {
        if (subSearch instanceof UnresolvedPlan) {
            ((UnresolvedPlan) subSearch).attach(table);
        }
        break;
    }
    subSearch = children.get(0);
}
```

docs/ppl-lang/ppl-appendcol-command.md
LantaoJin: What is the result of query `source=employees | stats avg(age) as avg_age1 by dept | APPENDCOL OVERRIDE=true [ stats avg(age) as avg_age2 by dept ]`?

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
YANG-DB: @LantaoJin yes I agree - I would even introduce a utility method for this specific functionality:

```
public LogicalPlan visitSubSearchChild(Node node, BiConsumer<Node, UnresolvedPlan> op) {
    Node subSearch = node;
    while (subSearch != null) {
        List<? extends Node> children = subSearch.getChild();
        if (children == null || children.isEmpty()) {
            op.accept(subSearch, table); 
            break;
        }
        subSearch = children.get(0); 
    }
    return (LogicalPlan) subSearch;
}

---
...
visitSubSearchChild(node, (subNode, unresolvedTable) -> {
    ((UnresolvedPlan) subNode).attach(unresolvedTable); 
});
...
```
LMK if this make sense ...

docs/ppl-lang/ppl-appendcol-command.md
andy-k-improving: Yes, that is correct, in the case of `source=employees | stats avg(age) as avg_age1 by dept | APPENDCOL OVERRIDE=true [ stats avg(age) as avg_age2 by dept ]`, the `dept` column from the main search will be discarded, and only 3 columns will be displated.
`dept (Sub), avg_age1 (Main), avg_age2 (Sub)`

Regarding the documentation, I will update the `ppl-appendcol-command.md` accordingly

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -253,6 +259,61 @@ public LogicalPlan visitTrendline(Trendline node, CatalystPlanContext context) {
...
+
+        // Duplicate the relation clause from main-search to sub-search.
+        AppendColCatalystUtils.appendRelationClause(node.getSubSearch(), context.getRelations());
```
YANG-DB, line 279: Can these parts be internal (implementation details) inside the `AppendColCatalystUtils` ?

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/AppendColCatalystUtils.java
andy-k-improving: Hi @LantaoJin @YANG-DB 
I understand this may not be good to have catch NPE for flow control, however I would argue the `try ... catch` is still required in this case after adding the `if (children == null || children.isEmpty()) `

Let's consider the original code snapshot:

```
subSearch = subSearch.getChild().get(0);
``` 

There two places in total will throw the NPE, which are: 

* `subSearch.getChild()`
* `.get(0)`

For the first point `subSearch.getChild()`, this is an abstract method and the unit-test I have implemented which indicated NPE will be thrown on certain type of Node implementation, such as `Aggregation.java`, while class attribute `child` or equivalent is null.

Implementation from Aggregation.java:
```
public List<UnresolvedPlan> getChild() {
    return ImmutableList.of(this.child);
  }
```
Associated NPE being thrown during `ImmutableList` construction time:
```
          <testcase classname="org.opensearch.flint.spark.ppl.PPLLogicalPlanAppendColCommandTranslatorTestSuite" name="test AppendCol with NO transformation on main" time="0.006">
                      <failure type="java.lang.NullPointerException">java.lang.NullPointerException
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:213)
	at com.google.common.collect.SingletonImmutableList.&lt;init&gt;(SingletonImmutableList.java:40)
	at com.google.common.collect.ImmutableList.of(ImmutableList.java:89)
	at org.opensearch.sql.ast.tree.Aggregation.getChild(Aggregation.java:67)
```

Which that said thou, I agree `children == null || children.isEmpty()` is sufficient to catch the second NPE `.get(0)`.

Should I proceed to add `children == null || children.isEmpty()` to prevent the NPE `.get(0)` while keeping the original code logic to catch NPE from `getChild()` or keeping the current implement to have `try...catch` to handle both scenarios?

YANG-DB (commented):
@andy-k-improving I would like to simplify the method `visitAppendCol(..)` so its simpler / smaller and more the complexity away into `AppendColCatalystUtils` 

does it make sense ?

andy-k-improving:
> @andy-k-improving I would like to simplify the method `visitAppendCol(..)` so its simpler / smaller and more the complexity away into `AppendColCatalystUtils`
> 
> does it make sense ?

On high level, that make sense, however how you foresee the structure will be like on `visitAppendCol( )` after the refactor?

Do you foresee t he logical plan composition flow happen within `visitAppendCol( )` or it will be moved entirely to `AppendColCatalystUtils`.

Ex:
```
visitAppendCol( ) {

      # Composite main query
      
      # Extract relation 
      
      # Composite sub-query
      
      # Some more operations.....
      
      # Return the result 
 }


v.s 


visitAppendCol( ) {
      
      # One util call to pass all argument into `AppendColCatalystUtils` for processing, expecting a logical plan with main and sub as return.
      
      # Some minor logic, which couldn't be done Util.
      
      # Return the result.

}
```

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
andy-k-improving: This depends on the approach of the refactor, would you mind to have a look on my below comments and advise.
In the meanwhile I have done some refactor move unecessary detail back to the Util class.

LantaoJin (approved):
Really appreciate your code revisions and clear explanations, which is improving the code quality. LGTM.

PR #994, opened 2024-12-16
#963 Unimplemented PPL Sort Syntax

### LINKED ISSUE #963: [BUG] Outdated PPL Sorting Syntax
**What is the bug?**

PPL supports syntax for sorting numerically, lexicographically, or by IP address (e.g. sort num(field_name), sort str(field_name), and sort ip(field_name), respectively) -- but this syntax has no effect on the resulting sort order. It should be removed.

This syntax appears to have been replaced by the addition of data types, and sorting is _always_ done according to the data type (i.e. numerical data types are sorted numerically, strings are sorted lexicographically), rather than as specified by `str`, `num`, or `ip` (i.e. these keywords _have no effect on the result_). Moreover, this syntax is not mentioned in the supporting user documentation for either OpenSearch SQL or Spark, and there is no actual implementation of the functionality in either code bases - the specified "sort type" is simply ignored.

**How can one reproduce the bug?**

Steps to reproduce the behaviour:
1. Create data set with numerical field
2. Sort using the `str` keyword (i.e. `sort str(field_name)`)
3. Observed: syntax is valid, but sorting is still numerical

**What is the expected behaviour?**

As a user, you would likely _expect_ the result to be sorted as specified (i.e. numerically, lexicographically, or by IP address), but the actual behaviour is to always sort by the field's data type.

If a user still wants to sort a numerical field lexicographically (for example) once this syntax is removed, they can still do so by first casting the field to a numerical data type before sorting it.

**What is your host/environment?**

N/A

**Do you have any screenshots?**

None

Related to #3145 (Add IP data type) and https://github.com/opensearch-project/sql/issues/3180 (same issue is SQL plugin).

currantw:
### Description

Updates documentation to mark PPL sort syntax (e.g. `sort num`) as _unimplemented._
Temporary measure related to #963, where the consensus seems to be that this syntax should eventually be supported.

### Related Issues

OpenSearch SQL issue: https://github.com/opensearch-project/sql/issues/3180

### Check List
- [x] ~Implemented unit tests~
- [x] ~Implemented tests for combination with other commands~
- [x] ~New added source code should include a copyright header~


ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
```diff
@@ -515,15 +515,7 @@ wcFieldList
    ;
 
 sortField
-   : (PLUS | MINUS)? sortFieldExpression
-   ;
-
-sortFieldExpression
```
YANG-DB, line 521: @currantw if i understand correctly - we remove the cast capability in the sort statement in favour of using the `cast` operator before ...
- can u plz add a few additional test examples for these use cases ?

currantw:
> I don't think the sort syntax is `outdated`. But I agree we should remove this syntax since it is unimplemented. How about change the PR title to `Remove unimplemented PPL Sort Syntax`?

Thanks @LantaoJin. Agree that "unimplemented" is more accurate. Please also see the discussion on the [equivalent SQL issue](https://github.com/opensearch-project/sql/issues/3180). The consensus there seems to be to keep this syntax for the time being, update all the documentation to indicate that this method is deprecated, and remove it completely in the next major release.

As such, I will be updating this PR to merely mark the syntax as deprecated in the documentation.

ppl-spark-integration/src/main/antlr4/OpenSearchPPLParser.g4
currantw: Based on the discussion on the [corresponding SQL issue](https://github.com/opensearch-project/sql/issues/3180), it seems like the consensus is to just mark as deprecated right now. I've updated this PR as suggested, so I think this comment is resolved for the time being. I've added a comment on the issue above with some example of equivalent `sort` commands using `cast`.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -263,7 +262,6 @@ public UnresolvedExpression visitCaseExpr(OpenSearchPPLParser.CaseExprContext ct
     public UnresolvedExpression visitIsEmptyExpression(OpenSearchPPLParser.IsEmptyExpressionContext ctx) {
         Function trimFunction = new Function(TRIM.getName().getFunctionName(), Collections.singletonList(this.visitFunctionArg(ctx.functionArg())));
         Function lengthFunction = new Function(LENGTH.getName().getFunctionName(), Collections.singletonList(trimFunction));
-        Compare lengthEqualsZero = new Compare(EQUAL.getName().getFunctionName(), lengthFunction, new Literal(0, DataType.INTEGER));
```
YANG-DB, line 266: why was this removed ?

YANG-DB (commented):
mostly good - just one question about this part:
```
Compare lengthEqualsZero = new Compare(EQUAL.getName().getFunctionName(), lengthFunction, new Literal(0, DataType.INTEGER));
```

acarbonetto:
This doesn't resolve #963 - but just deprecates the unimplemented language.  
I would link the issues together, but I don't think we should close that issue just because we're deprecating the language.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
```diff
@@ -80,7 +77,7 @@ public class AstExpressionBuilder extends OpenSearchPPLParserBaseVisitor<Unresol
                     .put("isnotnull", IS_NOT_NULL.getName().getFunctionName())
                     .put("ispresent", IS_NOT_NULL.getName().getFunctionName())
                     .build();
-    private AstBuilder astBuilder;
+    private final AstBuilder astBuilder;
```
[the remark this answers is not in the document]
YANG-DB: @currantw I agree with @acarbonetto - can we only keep changes relevant for this PR ?
thanks

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/parser/AstExpressionBuilder.java
currantw: Unused variable.

currantw:
Based on the conversation [here](https://github.com/opensearch-project/sql/issues/3180), it seems like consensus is that we should implement the PPL sort syntax, rather than deprecate and remove it. As such, I have changed this PR to simply mark it as "unimplemented" so that users don't unwittingly use it.

ppl-spark-integration/src/test/scala/org/opensearch/flint/spark/ppl/PPLLogicalPlanParseTranslatorTestSuite.scala
```diff
@@ -122,41 +123,47 @@ class PPLLogicalPlanParseTranslatorTestSuite
 
   test("test parse email & host expressions including cast and sort commands") {
```
YANG-DB, line 124: @currantw plz add the corresponding IT for a concrete index:
```
    val query =
      "source=t" +
        " | parse address '(?<streetNumber>\\d+) (?<street>.+)'" +
        " | eval streetNumberInt = cast(streetNumber as integer)" +
        " | where streetNumberInt > 500" +
        " | sort streetNumberInt" +
        " | fields streetNumber, street"
```

YANG-DB (approved):

YANG-DB:
@currantw LGTM ...
- @LantaoJin, @acarbonetto - any additional comments ?

docs/ppl-lang/PPL-Example-Commands.md
```diff
@@ -274,7 +274,8 @@ source = table |  where ispresent(a) |
 - `source=accounts | parse email '.+@(?<host>.+)' | stats count() by host`
 - `source=accounts | parse email '.+@(?<host>.+)' | eval eval_result=1 | fields host, eval_result`
 - `source=accounts | parse email '.+@(?<host>.+)' | where age > 45 | sort - age | fields age, email, host`
-- `source=accounts | parse address '(?<streetNumber>\d+) (?<street>.+)' | where streetNumber > 500 | sort num(streetNumber) | fields streetNumber, street`
+- `source=accounts | parse address '(?<streetNumber>\d+) (?<street>.+)' | eval streetNumberInt = cast(streetNumber as integer) | where streetNumberInt > 500 | sort streetNumberInt | fields streetNumber, street`
+  - **Note**: The `sort num` syntax is not implemented. To sort numerically, cast to a numerical data type - e.g. `sort cast(streetNumber as integer)`. See [#963](https://github.com/opensearch-project/opensearch-spark/issues/963) for more details.
```
acarbonetto, line 278:
```suggestion
  - **Note**: The `sort num` syntax is not implemented. See [#963](https://github.com/opensearch-project/opensearch-spark/issues/963) for more details.
```

docs/ppl-lang/ppl-parse-command.md
```diff
@@ -68,6 +68,8 @@ PPL query:
     | 880            | Holmes Lane    |
     +----------------+----------------+
 
+**Note**: The `sort num` syntax is not implemented. To sort numerically, cast to a numerical data type - e.g. `sort cast(streetNumber as integer)`. See [#963](https://github.com/opensearch-project/opensearch-spark/issues/963) for more details.
```
acarbonetto, line 71: Someone in the future may get confused by this comment since the `sort num` syntax is no longer mentioned here. 
I'd suggest removing the node completely.

acarbonetto (approved):

YANG-DB:
@currantw plz check why build is failing

YANG-DB (approved):

LantaoJin (approved):

acarbonetto (approved):

PR #1002, opened 2024-12-24
Add stalled PRs workflow

Swiddis:
### Description
Copies the stalled PR workflow from OS core: https://github.com/opensearch-project/OpenSearch/blob/main/.github/workflows/stalled.yml

### Related Issues
N/A

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/.github/workflows/stalled.yml
+++ b/.github/workflows/stalled.yml
@@ -1,0 +1,28 @@
+name: Label Stalled PRs
+on:
+  schedule:
+    - cron: '15 15 * * *' # Run every day at 15:15 UTC / 7:15 PST / 8:15 PDT
+permissions:
+  pull-requests: write
+jobs:
+  stale:
+    if: github.repository == 'opensearch-project/OpenSearch'
+    runs-on: ubuntu-latest
+    steps:
+      - name: GitHub App token
+        id: github_app_token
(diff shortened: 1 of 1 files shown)


.github/workflows/stalled.yml
```diff
@@ -0,0 +1,28 @@
+name: Label Stalled PRs
+on:
+  schedule:
+    - cron: '15 15 * * *' # Run every day at 15:15 UTC / 7:15 PST / 8:15 PDT
+permissions:
+  pull-requests: write
+jobs:
+  stale:
+    if: github.repository == 'opensearch-project/OpenSearch'
```
dai-chen, line 9: This is supposed to be our repo name?

Swiddis: Fixed

dai-chen (approved):

dai-chen:
Please double check if we need to create stalled label manually.

Swiddis:
I suspect we don't, but I'm not sure. In either case it wouldn't error if we create it ourselves, but may error if we don't.

acarbonetto (approved):

RyanL1997 (approved):

penghuo (commented):
Thx!

Issue #992, opened 2024-12-13
(RFC) Integration Tests on Apache Spark and Spark EMR


normanj-bitquill:
## Problem Overview

Work is underway to create the files needed for starting Docker clusters that can be used for integration testing. There will initially be two clusters, one for testing with Apache Spark and one for testing with Spark EMR.

The integration tests should be able switch between the two clusters (and any future clusters) without any changes.

The integration tests will run from either SBT or a standalone script. This to allow setting up CI steps for running the integration tests as well as running them locally.

Need an execution model for the integration tests that will work with both Apache Spark and Spark EMR.

## Proposed Solution

Structure the tests as a set of queries. Each query will have an expected query plan, and expected results (if the query succeeds). These tests can be made available to the Spark container in a bound directory. There will be another bound directory for holding the test results and query plans.

The bound directories are:
* `/tests/queries` - Each query to run is in a separate file
* `/tests/actual_query_plans`
* `/tests/actual_results`

A Spark application is created that runs the integration tests. The application will look in the directory `/tests/queries`. For each query file that it finds, it will:
* Run the query and save the results to `/tests/actual_results`
* Call explain on the query and save the results to `/tests/actual_query_plans`

The SBT build is updated for the integration test phase to do the following:
1. Build the Spark application
2. Start a Docker cluster (chosen based on a defined property or default to Apache Spark). The Spark container will run the Spark application and then shut down.
3. The SBT build will wait for the Spark container to shutdown.
4. Shutdown the Docker cluster
5. Use a local Scala integration test framework to verify the results of of the tests. The framework could generate HTML reports.

This solution does not involve connecting remotely to the Spark container. Since Spark is only running a Spark application, the solution will work for both Apache Spark and Spark EMR.

## Docker Clusters

Each Docker cluster will contain the following:
* Spark container
* OpenSearch server container
* OpenSearch Dashboards container
* Minio container

The Spark container is configured with both the Flint and PPL extensions, enabling it to both execute PPL queries and query indices on the OpenSearch server.

The OpenSearch Dashboards container is configured to connect to the OpenSearch server container.

The Spark container is started up as a driver and runs the Spark application.

normanj-bitquill:
I did some more testing and found another approach that could work.

Create a simple Spark app.
1. Look in a bound directory for one or more queries to run
2. Run each query and save the the result dataframes to files in the bound directory

[...]

PR #1001, opened 2024-12-22
update spark-docker example with jupyter tutorial  & notebook 

YANG-DB:
### Description
As part of the need to educate the community and users of how to use flint, ppl and its functionality we would like to introduce a mechanism (framework) that will allow setting up a simple tutorial based experience that will assist users to explore and experiment with Flint , Flint API, PPL, Queries and more.

### Related Issues
#992 
#1010

### Check List
- [x] Updated documentation (docs/ppl-lang/README.md)
- [x] Implemented unit tests
- [ ] Implemented tests for combination with other commands
- [ ] New added source code should include a copyright header
- [x] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/sql/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1003, opened 2024-12-31
Docker integ test with async API

normanj-bitquill:
### Description
Update the integration test docker stack to support the OpenSearch Async API and using Minio as an S3 storage engine. Also includes having everything configured on startup.

### Related Issues
#992 

### Check List
- [ ] Implemented unit tests


YANG-DB (approved):

YANG-DB:
@normanj-bitquill I've also started working on some similar [PR with Iceberg based docker-compose](https://github.com/opensearch-project/opensearch-spark/pull/1001) - 
lets see if we can share knowledge ...

normanj-bitquill:
### Current Status

In my testing, I have added a second OpenSearch container. This resolves issues with the cluster and indices going into yellow state.

Changes in OpenSearch node:
* Bind the docker socket
* Replace the `aws-java-sdk-emrserverless-*.jar` file with on that uses a different EMRClient. The new EMRClient writes a file to a directory (`/tmp/docker`) with the arguments for running docker.
* A script is run on the container that reads files from `/tmp/docker` and runs docker with the arguments.
* Disable system indices and system indices permissions
* Connect with second OpenSearch container

What works:
* Initiate query with aysnc-api on OpenSearch container
* Starts a new Spark docker container to call `spark-submit` using information from the start job request that the EMRClient received
* New container connects to spark cluster to execute the job
* Creates the request and result indices
* Saves the result to the result index

What is missing:
* Detecting that the job has finished (this is on the spark-submit container)
* The query is for a table on an S3 data source, query fails with table not found

normanj-bitquill:
@YANG-DB I have partially working async API in the latest commit. These are my testing steps:

1. Start the cluster using `docker compose up`
2. Run `spark-shell` on the master spark container.
   1. Create an external table with a location under `s3a://integ-test/`
   2. Put some data in the table
3. Submit a query:
   ```
   curl -u '...' -X POST -H 'Content-Type: application/json' -d '{"datasource": "mys3", "lang": 
   "sql", "query": "SELECT * FROM mys3.default.foo"}
   ```
4. Retrieve the result (will need to wait until it is ready, maybe about 1 minute)
   ```
   curl -u '...' -X POST -H 'Content-Type: application/json' -d '{}' 'http://localhost:9200/query_execution_result_mys3/_search?pretty'
   ```

The OpenSearch container will need to bind the docker socket `/var/run/docker.sock`. By default this isn't available on Mac, but can be enabled.

The OpenSearch container will start another container to process the async query. This is the place where we could slip in the EMR spark container (if there is any value from it).

normanj-bitquill:
I have not tested retrieving results using the Async API. This is likely broken, since cannot check on the EMR job status. I also haven't tested a streaming query (also likely broken).

docker/integ-test/docker-compose.yml
```diff
@@ -8,29 +23,35 @@ services:
       - "${UI_PORT:-4040}:4040"
       - "${SPARK_CONNECT_PORT}:15002"
     entrypoint: /opt/bitnami/scripts/spark/master-entrypoint.sh
+    user: root
```
YANG-DB, line 26: @normanj-bitquill why is this mandatory ?

docker/integ-test/docker-compose.yml
```diff
@@ -8,29 +23,35 @@ services:
...
         target: /opt/bitnami/spark/conf/log4j2.properties
       - type: bind
-        source: $PPL_JAR
+        source: ../../$PPL_JAR
         target: /opt/bitnami/spark/jars/ppl-spark-integration.jar
       - type: bind
-        source: $FLINT_JAR
+        source: ../../$FLINT_JAR
         target: /opt/bitnami/spark/jars/flint-spark-integration.jar
+      - type: bind
```
YANG-DB, line 52: can you plz explain what is the issue here we need to solve ?

docker/integ-test/opensearch/docker-command-runner.sh
YANG-DB, line 1: can you plz explain what is the issue here we need to solve ?

docker/integ-test/opensearch/emr-src/com/amazonaws/services/emrserverless/AWSEMRServerlessClientBuilder.java
```diff
@@ -0,0 +1,29 @@
+package com.amazonaws.services.emrserverless;
```
YANG-DB, line 1: missing license header

docker/integ-test/metastore/Dockerfile
YANG-DB, line 1: missing license header

docker/integ-test/opensearch/Dockerfile
YANG-DB, line 1: missing license header

docker/integ-test/opensearch/docker-command-runner.sh
YANG-DB: missing license header

docker/integ-test/opensearch/emr-src/org/opensearch/spark/emrserverless/DockerEMRServerlessClient.java
YANG-DB, line 1: missing license header

docker/integ-test/opensearch/opensearch-docker-it-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+
+./docker-command-runner.sh &
+echo $! > /var/run/docker-command-runner.pid
+
+su opensearch ./opensearch-docker-entrypoint.sh "$@"
+
+kill -TERM `cat /var/run/docker-command-runner.pid`
```
YANG-DB, line 11: why is this called ?
can you plz explain the entire flow for the docker compose ?

docker/integ-test/spark/spark-master-entrypoint.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
YANG-DB, line 1: this is nice - can you add some documentation of what is done here ?

docker/integ-test/opensearch/docker-command-runner.sh
normanj-bitquill: I have explained it here.
https://github.com/opensearch-project/opensearch-spark/pull/1003#discussion_r1911887918

docker/integ-test/docker-compose.yml
normanj-bitquill: Updated, the Spark master and worker nodes now use an image that is built from a `Dockerfile`. No need to start the containers with the `root` user.

normanj-bitquill:
@YANG-DB Using the async API to get the results works once the results are available in the result index.

Some interesting behaviour though:
* When you submit a query, the container starts up and runs the query. It then has a loop (looking for new queries) and continues until it finally times out after 3 minutes.
* If you submit a query and then submit a second query (in a different session), the second query isn't actually run until the container for the first query has exited (after the 3 minute timeout)
* When the container starts up, it downloads the Flint and PPL extensions. This adds latency and I am trying to have it use the locally built Flint and PPL jar files.

I am likely missing some code for the new EMRClient for job management. This may explain some of the behaviour above. If the Spark app is going to run for 3 minutes, we should skip calling docker when the same session is used and the container is still running.

normanj-bitquill:
> @YANG-DB Using the async API to get the results works once the results are available in the result index.
> 
> Some interesting behaviour though:
> 
> * When you submit a query, the container starts up and runs the query. It then has a loop (looking for new queries) and continues until it finally times out after 3 minutes.
> * If you submit a query and then submit a second query (in a different session), the second query isn't actually run until the container for the first query has exited (after the 3 minute timeout)
> * When the container starts up, it downloads the Flint and PPL extensions. This adds latency and I am trying to have it use the locally built Flint and PPL jar files.
> 
> I am likely missing some code for the new EMRClient for job management. This may explain some of the behaviour above. If the Spark app is going to run for 3 minutes, we should skip calling docker when the same session is used and the container is still running.

I have fixed up this behaviour. The Spark cluster as is, only supports running one application at a time. I have changed the spark submit container to run the query locally. This allows multiple sessions to run at the same time, but each session could have its own container running.

I have also disabled downloading the dependencies on startup, to speed up query execution.

normanj-bitquill:
@YANG-DB I have added some documentation in this PR. Can you take a look and help with getting a second reviewer?

docker/integ-test/opensearch/opensearch-docker-it-entrypoint.sh
YANG-DB: ok nice !!

dai-chen (commented):
Is this IT with SQL async query API? Just wondering if this should be in SQL repo?

normanj-bitquill:
> Is this IT for Flint Spark with SQL async query API? Just wondering if this should be in SQL repo?

@dai-chen This is for three use cases:
* Submitting queries directly to Spark in order to test the PPL extension for Spark.
* Submitting queries directly to Spark that use the OpenSearch datasource. Useful for testing the Flint extension for Spark.
* Using the Async API to submit queries to the OpenSearch server. Useful for testing the EMR workflow and querying S3/Glue datasources.

Something very similar would make sense for the SQL repo. The SQL repo will not need the spark master and spark containers.

YANG-DB:
> > Is this IT for Flint Spark with SQL async query API? Just wondering if this should be in SQL repo?
> 
> @dai-chen This is for three use cases:
> 
> * Submitting queries directly to Spark in order to test the PPL extension for Spark.
> * Submitting queries directly to Spark that use the OpenSearch datasource. Useful for testing the Flint extension for Spark.
> * Using the Async API to submit queries to the OpenSearch server. Useful for testing the EMR workflow and querying S3/Glue datasources.
> 
> Something very similar would make sense for the SQL repo. The SQL repo will not need the spark master and spark containers.

Actually this make sense as a wide test harness for such use cases where we want to test connectivity between different opensearch datasources ...

docker/integ-test/.env
```diff
@@ -5,9 +5,13 @@ MASTER_UI_PORT=8080
...
-FLINT_JAR=../../flint-spark-integration/target/scala-2.12/flint-spark-integration-assembly-0.7.0-SNAPSHOT.jar
+PPL_JAR=./ppl-spark-integration/target/scala-2.12/ppl-spark-integration-assembly-0.7.0-SNAPSHOT.jar
+FLINT_JAR=./flint-spark-integration/target/scala-2.12/flint-spark-integration-assembly-0.7.0-SNAPSHOT.jar
+SQL_APP_JAR=./spark-sql-application/target/scala-2.12/sql-job-assembly-0.7.0-SNAPSHOT.jar
 OPENSEARCH_NODE_MEMORY=512m
 OPENSEARCH_ADMIN_PASSWORD=C0rrecthorsebatterystaple.
 OPENSEARCH_PORT=9200
+OPENSEARCH_PA_PORT=9600
 OPENSEARCH_DASHBOARDS_PORT=5601
+S3_ACCESS_KEY=Vt7jnvi5BICr1rkfsheT
```
penghuo, line 16: is it for minio? it could be anything string?

docs/docker/integ-test/README.md
```diff
@@ -0,0 +1,166 @@
...
+This container also has a docker volume used to persist the metastore.
+
+### Minio (S3)
+
+A Minio server that acts as an S3 server. Is used as a part of the workflow of executing an S3/Glue query.
+It will contain the S3 tables data.
+
+This container also has a docker volume used to persist the S3 data.
+
+### Configuration-Updated
```
penghuo, line 71: typo. Configuration-updater?

docs/docker/integ-test/README.md
```diff
@@ -0,0 +1,166 @@
...
+```shell
+docker compose up -d
+```
+
+Stopping the cluster:
+```shell
+docker compose down -d
+```
+
+## Creating Tables in S3
```
penghuo, line 98: could add example of how to submit query using async API?

docs/docker/integ-test/README.md
```diff
@@ -0,0 +1,166 @@
+# Docker Cluster for Integration Testing
+
+## Introduction
+
+The docker cluster in `docker/integ-test` is designed to be used for integration testing. It supports the following
+use cases:
+1. Submitting queries directly to Spark in order to test the PPL extension for Spark.
+2. Submitting queries directly to Spark that use the OpenSearch datasource. Useful for testing the Flint extension
+   for Spark.
+3. Using the Async API to submit queries to the OpenSearch server. Useful for testing the EMR workflow and querying
```
penghuo, line 10: what is the difference of 2 and 3?

docker/integ-test/.env
normanj-bitquill: Yes, this is for minio container and is only valid for the minio container.

docs/docker/integ-test/README.md
normanj-bitquill: Added an example of submitting a query and getting the results.

docs/docker/integ-test/README.md
normanj-bitquill: For 2, a query is sent directly to Spark and Spark needs to query the index on the OpenSearch server. An example of where this might be useful is testing PPL on Spark, querying OpenSearch indices.

For 3, a query is sent directly to the OpenSearch server for the S3/Glue datasource. A new Spark container is started to process the request. The Spark container needs to query the table using Hive and minio. The Spark container is also making use of the `spark-sql-application` in this repository to process the query.

penghuo (approved):

YANG-DB:
thanks @penghuo !

docker/integ-test/.env
LantaoJin: @penghuo the current end2end test always failed with
```
2025-05-12T01:57:02.7209031Z com.amazonaws.services.s3.model.AmazonS3Exception: The Access Key Id you provided does not exist in our records. (Service: Amazon S3; Status Code: 403; Error Code: InvalidAccessKeyId; Request ID: ...
```
What is the correct key is?

penghuo: I do not know. it is not actually s3 key.

PR #1011, opened 2025-01-08
Adaptive rate limiting for OpenSearch bulk requests

seankao-az:
### Description
Add (client side) adaptive rate limiting for OpenSearch bulk requests. For each successful bulk request, rate limit will increase linearly; for each failure, rate limit will decrease exponentially.

Screenshot 2025-01-02 at 5 20 57 PM
Each worker node adaptively adjust their own rate limit based on responses they get from the OpenSearch cluster. They update their rate limit each time they receive a response.

Retry requests will not be rate limited. They will go through regardless of the rate limit. Only new request will be blocked by rate limiter. An alternative is to have retries also be rate limited. However, this will result in slowing down retries and increasing overall response time, since we do not ensure fairness or priority. Retries could be delayed due to new requests consuming the available rate.

Several new spark properties are added:
- `spark.datasource.flint.write.bulk.rate_limit_per_node.enabled`: to enable the feature
- `spark.datasource.flint.write.bulk.rate_limit_per_node.min`: min rate limit (documents/sec)
- `spark.datasource.flint.write.bulk.rate_limit_per_node.max`: max rate limit (documents/sec)
- `spark.datasource.flint.write.bulk.rate_limit_per_node.increase_step`: rate limit increase step for successful bulk request
- `spark.datasource.flint.write.bulk.rate_limit_per_node.decrease_ratio`: rate limit decrease ratio for failed bulk request

This PR also introduces a new (shaded) dependency:
```
"com.google.guava" % "guava" % "33.3.1-jre"
```

### Test

Screenshot 2025-01-09 at 5 28 45 PM

```
25/01/10 00:34:35 INFO BulkRequestRateLimiter: Setting rate limit for bulk request to 17396 documents/sec
25/01/10 00:34:35 INFO BulkRequestRateLimiter: Acquired 5068 permits
25/01/10 00:34:35 INFO HttpStatusCodeResultPredicate: Checking if status code is retryable: 200
25/01/10 00:34:35 INFO HttpStatusCodeResultPredicate: Status code 200 check result: false
25/01/10 00:34:35 INFO HttpAOSSResultPredicate: Checking if response is retryable
25/01/10 00:34:35 INFO HttpAOSSResultPredicate: Status code 200 is not 400. Check result: false
25/01/10 00:34:35 INFO OpenSearchBulkWrapper: Found retryable failure in the bulk response
25/01/10 00:34:35 INFO BulkRequestRateLimiter: Setting rate limit for bulk request to 13916 documents/sec
[...]

Issue #640, opened 2024-09-10
[FEATURE] Avoid throttling when writing data to index


ykmr1224:
**Is your feature request related to a problem?**
When flint execute indexing query, each worker node will call OpenSearch bulk API to write data.
Currently it could end up causing throttling and execution failure due to too high request rate.
Although already implemented rate limiter (https://github.com/opensearch-project/opensearch-spark/pull/567) and bulk request retry (https://github.com/opensearch-project/opensearch-spark/pull/572), it still have high chance to cause failure especially when:
- There are many worker node
- OpenSearch Serverless is used as index store (does not support WAIT_FOR refresh policy)
- Other job or other process send requests to the destination

**What solution would you like?**
To fully address this issue, we want to introduce adaptive rate limit to `bulk` request. One example of adaptive control algorithm is AIMD(additive-increase/multiplicative-decrease), which is used for TCP congestion control ([ref](https://en.wikipedia.org/wiki/Additive_increase/multiplicative_decrease)). By utilizing this kind of algorithm, each node will try to adjust their traffic by gradually increase the call rate and quickly decrease when the request is throttled. (Each node will ideally end up similar call rate)

[Netflix/concurrency-limits](https://github.com/Netflix/concurrency-limits) provides AIMD based concurrency limit and we might be able to utilize it for our implementation.

**What alternatives have you considered?**
- Centralized rate limiter using DynamoDB, Redis, etc.
- Driver node to decide rate limit for each worker node

**Do you have any additional context?**
n/a

dai-chen:
Will create separate issue to track follow-up work.

PR #1027, opened 2025-02-01
Stabilize adaptive rate limit by considering current rate

ykmr1224:
### Description
- Based on testing adaptive rate limit implementation with data, I identified following issues:
 - Rate limit was increase even when the current limit is not fully utilized, and this lead to too high rate limit and limit decrease took long time to become effective.
 - As the rate limit is adjusted when a request finished, adjustment frequency depends on the current request rate. (means rate limit is increased slowly when rate limit is low, and quickly when rate limit is high)
- This PR address the first issue. (Second issue is not critical as far as we don't set minimum rate to lower value. current value is 5000)
- BulkRequestRateMeter collect data points for current rate using recent 3 seconds window, and calculate the estimated current rate. If the current rate is lower than 80% of rate limit, it won't increase the rate.

### Related Issues
- https://github.com/opensearch-project/opensearch-spark/issues/640
- Previous PR: https://github.com/opensearch-project/opensearch-spark/pull/1011

### Check List
- [x] Implemented unit tests

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -543,6 +543,8 @@ In the index mapping, the `_meta` and `properties`field stores meta and schema i
 - `spark.datasource.flint.read.scroll_size`: default value is 100.
 - `spark.datasource.flint.read.scroll_duration`: default value is 5 minutes. scroll context keep alive duration.
 - `spark.datasource.flint.retry.max_retries`: max retries on failed HTTP request. default value is 3. Use 0 to disable retry.
+- `spark.datasource.flint.retry.bulk.max_retries`: max retries on failed bulk request. default value is 10. Use 0 to disable retry.
+- `spark.datasource.flint.retry.bulk.initial_backoff`: initial backoff in seconds for bulk request retry, default is 4.
 - `spark.datasource.flint.retry.http_status_codes`: retryable HTTP response status code list. default value is "429,502" (429 Too Many Request and 502 Bad Gateway).
 - `spark.datasource.flint.retry.exception_class_names`: retryable exception class name list. by default no retry on any exception thrown.
 - `spark.datasource.flint.read.support_shard`: default is true. set to false if index does not support shard (AWS OpenSearch Serverless collection). Do not use in production, this setting will be removed in later version. 
--- a/flint-core/src/main/scala/org/opensearch/flint/core/storage/BulkRequestRateLimiterImpl.java
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/storage/BulkRequestRateLimiterImpl.java
@@ -19,12 +19,14 @@ public class BulkRequestRateLimiterImpl implements BulkRequestRateLimiter {
   private final long maxRate;
   private final long increaseStep;
   private final double decreaseRatio;
+  private final RequestRateMeter requestRateMeter;
 
   public BulkRequestRateLimiterImpl(FlintOptions flintOptions) {
     minRate = flintOptions.getBulkRequestMinRateLimitPerNode();
     maxRate = flintOptions.getBulkRequestMaxRateLimitPerNode();
     increaseStep = flintOptions.getBulkRequestRateLimitPerNodeIncreaseStep();
     decreaseRatio = flintOptions.getBulkRequestRateLimitPerNodeDecreaseRatio();
+    requestRateMeter = new RequestRateMeter();
 
     LOG.info("Setting rate limit for bulk request to " + minRate + " documents/sec");
     this.rateLimiter = RateLimiter.create(minRate);
@@ -42,6 +44,7 @@ public class BulkRequestRateLimiterImpl implements BulkRequestRateLimiter {
   public void acquirePermit(int permits) {
     this.rateLimiter.acquire(permits);
     LOG.info("Acquired " + permits + " permits");
+    requestRateMeter.addDataPoint(System.currentTimeMillis(), permits);
   }
 
   /**
@@ -49,7 +52,18 @@ public class BulkRequestRateLimiterImpl implements BulkRequestRateLimiter {
    */
   @Override
   public void increaseRate() {
-    setRate(getRate() + increaseStep);
+    if (isEstimatedCurrentRateCloseToLimit()) {
+      setRate(getRate() + increaseStep);
+    } else {
+      LOG.info("Rate increase was blocked.");
+    }
+    LOG.info("Current rate limit for bulk request is " + getRate() + " documents/sec");
+  }
+
+  private boolean isEstimatedCurrentRateCloseToLimit() {
+    long currentEstimatedRate = requestRateMeter.getCurrentEstimatedRate();
+    LOG.info("Current estimated rate is " + currentEstimatedRate + " documents/sec");
+    return getRate() * 0.8 < currentEstimatedRate;
   }
 
   /**
--- a/flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkWrapper.java
+++ b/flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkWrapper.java
@@ -50,7 +50,6 @@ public class OpenSearchBulkWrapper {
    * @return Last result
    */
   public BulkResponse bulk(RestHighLevelClient client, BulkRequest bulkRequest, RequestOptions options) {
-    rateLimiter.acquirePermit(bulkRequest.requests().size());
     return bulkWithPartialRetry(client, bulkRequest, options);
   }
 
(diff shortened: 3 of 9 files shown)


ykmr1224:
I'll address integ test failure.

flint-core/src/main/scala/org/opensearch/flint/core/storage/BulkRequestRateLimiterImpl.java
```diff
@@ -42,14 +44,26 @@ public void acquirePermit() {
...
+    } else {
+      LOG.info("Rate increase was blocked.");
+    }
+    LOG.info("Current rate limit for bulk request is " + getRate() + " documents/sec");
+  }
+
+  private boolean isEstimatedCurrentRateCloseToLimit() {
+    long currentEstimatedRate = requestRateMeter.getCurrentEstimatedRate();
+    LOG.info("Current estimated rate is " + currentEstimatedRate + " documents/sec");
+    return getRate() * 0.8 < currentEstimatedRate;
```
dai-chen, line 66: Is 0.8 optimal value in our test? Is it necessary to be configurable?

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkWrapper.java
```diff
@@ -69,11 +71,13 @@ private BulkResponse bulkWithPartialRetry(RestHighLevelClient client, BulkReques
           })
           .get(() -> {
             requestCount.incrementAndGet();
+            rateLimiter.acquirePermit(nextRequest.get().requests().size());
```
dai-chen, line 74: So this changes the initial design earlier that retry request now will also go through the rate limiter, right?

flint-core/src/main/scala/org/opensearch/flint/core/storage/RequestRateMeter.java
```diff
@@ -0,0 +1,47 @@
...
+  private static class DataPoint {
+    long timestamp;
+    long requestCount;
+    public DataPoint(long timestamp, long requestCount) {
+      this.timestamp = timestamp;
+      this.requestCount = requestCount;
+    }
+  }
+
+  private List<DataPoint> dataPoints = new LinkedList<>();
```
dai-chen, line 27: np: Is Queue more clear for this sliding window impl? Because `getCurrentEstimatedRate` is `synchronized` anyway, there is no need to maintain `currentSum` and synchronize others if we use concurrent queue?

flint-core/src/main/scala/org/opensearch/flint/core/storage/OpenSearchBulkWrapper.java
ykmr1224: Yes, that's right.

flint-core/src/main/scala/org/opensearch/flint/core/storage/BulkRequestRateLimiterImpl.java
ykmr1224: It should not matter a lot as far as it is close to 1. (As we calculate the current rate based on last 3 seconds, it requires some buffer) I didn't see much benefit to make it configurable.

flint-core/src/main/scala/org/opensearch/flint/core/storage/RequestRateMeter.java
ykmr1224: `currentSum` is just a minor optimization to avoid calculating sum again and again. (should not have much impact, though)
As we delete the item within while loop, we need to synchronize. Otherwise, the queue could be manipulated in another thread. (`removeOldDataPoints` could be called by multiple thread at the same time.)

Using Queue interface make sense. Let me fix.

docs/index.md
```diff
@@ -543,6 +543,8 @@ In the index mapping, the `_meta` and `properties`field stores meta and schema i
 - `spark.datasource.flint.read.scroll_size`: default value is 100.
 - `spark.datasource.flint.read.scroll_duration`: default value is 5 minutes. scroll context keep alive duration.
 - `spark.datasource.flint.retry.max_retries`: max retries on failed HTTP request. default value is 3. Use 0 to disable retry.
+- `spark.datasource.flint.retry.bulk.max_retries`: max retries on failed bulk request. default value is 10. Use 0 to disable retry.
+- `spark.datasource.flint.retry.bulk.initial_backoff`: initial backoff in seconds for bulk request retry, default is 4.
```
penghuo, line 547: any reason choose 4s as default value?

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
```diff
@@ -136,6 +136,18 @@ object FlintSparkConf {
...
+  val BULK_MAX_RETRIES =
+    FlintConfig(s"spark.datasource.flint.${FlintRetryOptions.BULK_MAX_RETRIES}")
+      .datasourceOption()
+      .doc("max retries on failed HTTP request, 0 means retry is disabled, default is 10")
+      .createWithDefault(String.valueOf(FlintRetryOptions.DEFAULT_BULK_MAX_RETRIES))
+
+  val BULK_INITIAL_BACKOFF =
+    FlintConfig(s"spark.datasource.flint.${FlintRetryOptions.BULK_INITIAL_BACKOFF}")
+      .datasourceOption()
+      .doc("initial backoff for bulk request retry, default is 4")
```
penghuo, line 148: polilsh doc, `initial backoff in seconds for bulk request retry, default is 4s`

flint-core/src/main/scala/org/opensearch/flint/core/storage/RequestRateMeter.java
```diff
@@ -0,0 +1,47 @@
...
+    public DataPoint(long timestamp, long requestCount) {
+      this.timestamp = timestamp;
+      this.requestCount = requestCount;
+    }
+  }
+
+  private Queue<DataPoint> dataPoints = new LinkedList<>();
+  private long currentSum = 0;
+
+  synchronized void addDataPoint(long timestamp, long requestCount) {
```
penghuo, line 31: nit, could we consider reduce the datapoints maintained by using bucket-based aggregation which have bounde memory usage.

flint-core/src/main/scala/org/opensearch/flint/core/storage/RequestRateMeter.java
```diff
@@ -0,0 +1,47 @@
...
+    dataPoints.add(new DataPoint(timestamp, requestCount));
+    currentSum += requestCount;
```
penghuo, line 33: we could removeOldDataPoint during add to reduce memory usage.

ykmr1224: Number of data points tend to be small like around 1 ~ 20 since each batch contain around 5k requests (we put data point for each batch). And memory usage can be almost ignorable.

flint-core/src/main/scala/org/opensearch/flint/core/storage/RequestRateMeter.java
ykmr1224: Number of data points tend to be small like around 1 ~ 20 since each batch contain around 5k requests (we put data point for each batch). And memory usage is almost ignorable.

docs/index.md
ykmr1224: It was fixed value and I made it a configuration. The original intention is having higher initial backoff to quickly reduce the rate.

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
ykmr1224: Thanks, fixed.

dai-chen (approved):
Thanks for the changes!

YANG-DB (commented):
@ykmr1224 
Are you familiar with [resilience4j](https://github.com/resilience4j/resilience4j) ?
I think this can be an interesting option for better resilient with managing fault tolerance for remote communications.
We already use it in [SQL repo](https://github.com/opensearch-project/sql/blob/11a644050c9c8030e82321e41917e8efce38ac16/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchResourceMonitor.java#L8)

ykmr1224:
> @ykmr1224 Are you familiar with [resilience4j](https://github.com/resilience4j/resilience4j) ? I think this can be an interesting option for better resilient with managing fault tolerance for remote communications. We already use it in [SQL repo](https://github.com/opensearch-project/sql/blob/11a644050c9c8030e82321e41917e8efce38ac16/opensearch/src/main/java/org/opensearch/sql/opensearch/monitor/OpenSearchResourceMonitor.java#L8)

I checked resilience4j earlier, do we see any additional benefit than using Failsafe library which we are currently using?

flint-core/src/main/scala/org/opensearch/flint/core/storage/RequestRateMeter.java
ykmr1224: Added removeOldDataPoints in addDataPoint.

YANG-DB:
IMO there is a benefit of using the same library in both repos (SQL/spark)...

penghuo (approved):

PR #1032, opened 2025-02-04
Support OpenSearch alias field type

penghuo:
#### Summary
This PR introduces support for alias fields in `FlintDataType`, allowing fields to be referenced by alternative names. The change enables schema parsing and JSON processing to recognize alias fields and map them to their corresponding primary fields.

#### Changes
1. **FlintDataType.scala**
   - Updated `deserializeJValue` to identify and process alias fields.
   - Modified `deserializeField` to support alias metadata assignment.

2. **FlintJacksonParser.scala**
   - Implemented `fieldMapping` to associate JSON keys with schema field indices, allowing alias fields to be correctly resolved during parsing.
   - Updated JSON parsing logic to handle multiple schema fields mapped to a single JSON key.

3. **Add OpenSearch Table datatype in docs**
   - https://github.com/penghuo/penghuo-opensearch-spark/blob/dataTypeAlias/docs/opensearch-table.md#data-types

#### Testing
- Added test cases for alias field deserialization to validate schema correctness.
- Integration tests confirm alias field support in OpenSearch queries.

### Related Issues
https://github.com/opensearch-project/opensearch-spark/issues/1033

### Check List
- [x] Implemented unit tests

### DIFF
--- a/docs/opensearch-table.md
+++ b/docs/opensearch-table.md
@@ -59,8 +59,8 @@ The following table defines the data type mapping between OpenSearch index field
 | byte                     | ByteType                          |
 | double                   | DoubleType                        |
 | float                    | FloatType                         |
-| date(Timestamp)          | DateType                          |
-| date(Date)               | TimestampType                     |
+| date(Timestamp)          | TimestampType                     |
+| date(Date)               | DateType                          |
 | keyword                  | StringType, VarcharType, CharType |
 | text                     | StringType(meta(osType)=text)     |
 | object                   | StructType                        |
--- a/docs/opensearch-table.md
+++ b/docs/opensearch-table.md
@@ -47,6 +47,37 @@ Using a wildcard index name:
 val df = spark.sql("SELECT * FROM dev.default.`my_index*`")
 df.show()
 ```
+## Data Types
+The following table defines the data type mapping between OpenSearch index field type and Spark data type.
+
+| **OpenSearch FieldType** | **SparkDataType**                 |
+|--------------------------|-----------------------------------|
+| boolean                  | BooleanType                       |
+| long                     | LongType                          |
+| integer                  | IntegerType                       |
+| short                    | ShortType                         |
+| byte                     | ByteType                          |
+| double                   | DoubleType                        |
+| float                    | FloatType                         |
+| date(Timestamp)          | DateType                          |
+| date(Date)               | TimestampType                     |
+| keyword                  | StringType, VarcharType, CharType |
+| text                     | StringType(meta(osType)=text)     |
+| object                   | StructType                        |
+| alias                    | Inherits referenced field type    |
+
+* OpenSearch data type date is mapped to Spark data type based on the format:
+    * Map to DateType if format = strict_date, (we also support format = date, may change in future)
+    * Map to TimestampType if format = strict_date_optional_time_nanos, (we also support format =
+      strict_date_optional_time | epoch_millis, may change in future)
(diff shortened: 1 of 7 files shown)


penghuo:
@ykmr1224 @dai-chen please help review.

docs/opensearch-table.md
```diff
@@ -47,6 +47,37 @@ Using a wildcard index name:
...
+|--------------------------|-----------------------------------|
+| boolean                  | BooleanType                       |
+| long                     | LongType                          |
+| integer                  | IntegerType                       |
+| short                    | ShortType                         |
+| byte                     | ByteType                          |
+| double                   | DoubleType                        |
+| float                    | FloatType                         |
+| date(Timestamp)          | DateType                          |
+| date(Date)               | TimestampType                     |
```
ykmr1224, line 63: Why do we map Timestamp to DateType and Date to TimestampType?

penghuo: it is typo, fix in doc. https://github.com/opensearch-project/opensearch-spark/pull/1032/commits/6066ee13bb14099890b4005009caedea872d691f

ykmr1224 (approved):

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/json/FlintJacksonParser.scala
```diff
@@ -448,13 +449,33 @@ class FlintJacksonParser(
     var badRecordException: Option[Throwable] = None
     var skipRow = false
 
+    // Build mapping from JSON key to sequence of schema field indices.
```
dai-chen, line 452: np: Is it possible to add UT for this class in future?

dai-chen (approved):
Thanks for the changes!

qianheng-aws:
LGTM!

One minor concern from me is that the implementation of this PR is more like the Option3 we mentioned in: https://github.com/opensearch-project/sql/pull/3246#issue-2788757345

So there are some minor behavioral differences between OpenSearch index PPL and Flint PPL. Since we populate the InternalRow with alias columns filled with real data during scanning, the alias columns won't reflect updates made to the original columns in the following execution.

Nevertheless, since Spark doesn't support alias types, implementing like Option1 is challenging in Flint. Besides, it's a very edge case where users would query using alias columns while updating the original columns.

YANG-DB (approved):

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/json/FlintJacksonParser.scala
penghuo: Add a task to track it. https://github.com/opensearch-project/opensearch-spark/issues/1037

penghuo:
> LGTM!
> 
> One minor concern from me is that the implementation of this PR is more like the Option3 we mentioned in: [opensearch-project/sql#3246 (comment)](https://github.com/opensearch-project/sql/pull/3246#issue-2788757345)
> 
> So there are some minor behavioral differences between OpenSearch index PPL and Flint PPL. Since we populate the InternalRow with alias columns filled with real data during scanning, the alias columns won't reflect updates made to the original columns in the following execution.
> 
> Nevertheless, since Spark doesn't support alias types, implementing like Option1 is challenging in Flint. Besides, it's a very edge case where users would query using alias columns while updating the original columns in this query.

Yes, it is option3. 
This is not an issue for the OS since it operates in near-real-time. Consequently, the latest document updates may not be immediately reflected during a search.

Issue #1026, opened 2025-01-31
[BUG] The default behavior of PPL Lookup does not work as expected


penghuo:
**What is the bug?**
* issue-1, expect `source = $sourceTable | LOOKUP $lookupTbl id` should replace id column , but the result include 2 id columns. IMO, it should work as same as default behaviour `source = $sourceTable | LOOKUP $lookupTbl id REPLACE id, department`.  
```
+----+-----+----------+-------+------+----+----------+
|  id| name|occupation|country|salary|  id|department|
+----+-----+----------+-------+------+----+----------+
|1000| Jake|  Engineer|England|100000|1000|        IT|
|1001|Hello|    Artist|    USA| 70000|NULL|      NULL|
|1002| John|    Doctor| Canada|120000|1002|      DATA|
|1003|David|    Doctor|   NULL|120000|1003|        HR|
|1004|David|      NULL| Canada|     0|NULL|      NULL|
|1005| Jane| Scientist| Canada| 90000|1005|      DATA|
+----+-----+----------+-------+------+----+----------+
```
* issue-2,  got exception when execute `source = $sourceTable | LOOKUP $lookupTbl id REPLACE id, department`
```
[AMBIGUOUS_REFERENCE] Reference `id` is ambiguous, could be: [`__auto_generated_subquery_name_l`.`id`, `__auto_generated_subquery_name_s`.`id`].
org.apache.spark.sql.AnalysisException: [AMBIGUOUS_REFERENCE] Reference `id` is ambiguous, could be: [`__auto_generated_subquery_name_l`.`id`, `__auto_generated_subquery_name_s`.`id`].
```

**How can one reproduce the bug?**
* Add following code to FlintSparkPPLLookupITSuite
```
  protected def sourceTable(testTable: String): Unit = {
    sql(s"""
           | CREATE TABLE $testTable
           | (
           |   id INT,
           |   name STRING,
           |   occupation STRING,
           |   country STRING,
           |   salary INT
           | )
           | USING $tableType $tableOptions
           |""".stripMargin)

    // Insert data into the new table
    sql(s"""
           | INSERT INTO $testTable
           | VALUES (1000, 'Jake', 'Engineer', 'England' , 100000),
           |        (1001, 'Hello', 'Artist', 'USA', 70000),
           |        (1002, 'John', 'Doctor', 'Canada', 120000),
           |        (1003, 'David', 'Doctor', null, 120000),
[...]

PR #1035, opened 2025-02-05
[BUG] Fix the PPL Lookup command behavior when inputField is missing and REPLACE with existing fields

LantaoJin:
### Description
**issue-1:**

The current behaviour of [inputField](https://github.com/opensearch-project/opensearch-spark/blob/main/docs/ppl-lang/ppl-lookup-command.md) is "If you don't specify any `<inputField>`, **all fields of `<lookupIndex>`** where matched values are applied to result output." 

The correct behaviour should be "If you don't specify any `<inputField>`, **all fields of `<lookupIndex>` that are not the mapping fields** where matched values are applied to result output." 

**issue-2:**

A known bug. The existing IT "test LOOKUP lookupTable name REPLACE occupation" provided a workaround. The PR fixed the issue and provide a new IT "test LOOKUP lookupTable name REPLACE occupation - 2"

### Related Issues
Resolves https://github.com/opensearch-project/opensearch-spark/issues/1026

### Check List
- [x] Implemented unit tests

### DIFF
--- a/integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLLookupITSuite.scala
+++ b/integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLLookupITSuite.scala
@@ -51,9 +51,6 @@ class FlintSparkPPLLookupITSuite
 
   test("test LOOKUP lookupTable uid AS id REPLACE department") {
     val frame = sql(s"source = $sourceTable| LOOKUP $lookupTable uid AS id REPLACE department")
-    // frame.show()
-    // frame.explain(true)
-    val results: Array[Row] = frame.collect()
     val expectedResults: Array[Row] = Array(
       Row(1000, "Jake", "Engineer", "England", 100000, "IT"),
       Row(1001, "Hello", "Artist", "USA", 70000, null),
@@ -61,15 +58,17 @@ class FlintSparkPPLLookupITSuite
       Row(1003, "David", "Doctor", null, 120000, "HR"),
       Row(1004, "David", null, "Canada", 0, null),
       Row(1005, "Jane", "Scientist", "Canada", 90000, "DATA"))
+    assertSameRows(expectedResults, frame)
 
-    implicit val rowOrdering: Ordering[Row] = Ordering.by[Row, Integer](_.getAs[Integer](0))
-    assert(results.sorted.sameElements(expectedResults.sorted))
     val lookupProject =
       Project(Seq(UnresolvedAttribute("department"), UnresolvedAttribute("uid")), lookupAlias)
     val joinCondition = EqualTo(UnresolvedAttribute("uid"), UnresolvedAttribute("id"))
     val joinPlan = Join(sourceAlias, lookupProject, LeftOuter, Some(joinCondition), JoinHint.NONE)
     val coalesceForSafeExpr =
-      Coalesce(Seq(UnresolvedAttribute("department"), UnresolvedAttribute("department")))
+      Coalesce(
+        Seq(
+          UnresolvedAttribute("__auto_generated_subquery_name_l.department"),
+          UnresolvedAttribute("department")))
     val projectAfterJoin = Project(
       Seq(
         UnresolvedStar(Some(Seq("__auto_generated_subquery_name_s"))),
@@ -88,9 +87,6 @@ class FlintSparkPPLLookupITSuite
 
   test("test LOOKUP lookupTable uid AS id APPEND department") {
     val frame = sql(s"source = $sourceTable| LOOKUP $lookupTable uid AS id APPEND department")
-    // frame.show()
-    // frame.explain(true)
-    val results: Array[Row] = frame.collect()
     val expectedResults: Array[Row] = Array(
       Row(1000, "Jake", "Engineer", "England", 100000, "IT"),
       Row(1001, "Hello", "Artist", "USA", 70000, null),
@@ -98,16 +94,17 @@ class FlintSparkPPLLookupITSuite
       Row(1003, "David", "Doctor", null, 120000, "HR"),
       Row(1004, "David", null, "Canada", 0, null),
       Row(1005, "Jane", "Scientist", "Canada", 90000, "DATA"))
-
-    implicit val rowOrdering: Ordering[Row] = Ordering.by[Row, Integer](_.getAs[Integer](0))
-    assert(results.sorted.sameElements(expectedResults.sorted))
+    assertSameRows(expectedResults, frame)
 
     val lookupProject =
       Project(Seq(UnresolvedAttribute("department"), UnresolvedAttribute("uid")), lookupAlias)
     val joinCondition = EqualTo(UnresolvedAttribute("uid"), UnresolvedAttribute("id"))
     val joinPlan = Join(sourceAlias, lookupProject, LeftOuter, Some(joinCondition), JoinHint.NONE)
     val coalesceExpr =
-      Coalesce(Seq(UnresolvedAttribute("department"), UnresolvedAttribute("department")))
+      Coalesce(
+        Seq(
+          UnresolvedAttribute("department"),
+          UnresolvedAttribute("__auto_generated_subquery_name_l.department")))
     val coalesceForSafeExpr = Coalesce(Seq(coalesceExpr, UnresolvedAttribute("department")))
     val projectAfterJoin = Project(
       Seq(
@@ -128,9 +125,6 @@ class FlintSparkPPLLookupITSuite
   test("test LOOKUP lookupTable uid AS id REPLACE department AS country") {
     val frame =
       sql(s"source = $sourceTable| LOOKUP $lookupTable uid AS id REPLACE department AS country")
-    // frame.show()
-    // frame.explain(true)
-    val results: Array[Row] = frame.collect()
     val expectedResults: Array[Row] = Array(
       Row(1000, "Jake", "Engineer", 100000, "IT"),
       Row(1001, "Hello", "Artist", 70000, "USA"),
@@ -138,16 +132,17 @@ class FlintSparkPPLLookupITSuite
       Row(1003, "David", "Doctor", 120000, "HR"),
       Row(1004, "David", null, 0, "Canada"),
       Row(1005, "Jane", "Scientist", 90000, "DATA"))
-
-    implicit val rowOrdering: Ordering[Row] = Ordering.by[Row, Integer](_.getAs[Integer](0))
-    assert(results.sorted.sameElements(expectedResults.sorted))
+    assertSameRows(expectedResults, frame)
 
     val lookupProject =
       Project(Seq(UnresolvedAttribute("department"), UnresolvedAttribute("uid")), lookupAlias)
     val joinCondition = EqualTo(UnresolvedAttribute("uid"), UnresolvedAttribute("id"))
     val joinPlan = Join(sourceAlias, lookupProject, LeftOuter, Some(joinCondition), JoinHint.NONE)
     val coalesceForSafeExpr =
-      Coalesce(Seq(UnresolvedAttribute("department"), UnresolvedAttribute("country")))
+      Coalesce(
+        Seq(
+          UnresolvedAttribute("__auto_generated_subquery_name_l.department"),
+          UnresolvedAttribute("__auto_generated_subquery_name_s.country")))
     val projectAfterJoin = Project(
       Seq(
         UnresolvedStar(Some(Seq("__auto_generated_subquery_name_s"))),
@@ -167,9 +162,6 @@ class FlintSparkPPLLookupITSuite
   test("test LOOKUP lookupTable uid AS id APPEND department AS country") {
     val frame =
       sql(s"source = $sourceTable| LOOKUP $lookupTable uid AS id APPEND department AS country")
-    // frame.show()
-    // frame.explain(true)
-    val results: Array[Row] = frame.collect()
     val expectedResults: Array[Row] = Array(
       Row(1000, "Jake", "Engineer", 100000, "England"),
       Row(1001, "Hello", "Artist", 70000, "USA"),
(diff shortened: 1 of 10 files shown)


ppl-spark-integration/src/main/scala/org/opensearch/flint/spark/FlintPPLSparkExtensions.scala
```diff
@@ -16,7 +16,7 @@ class FlintPPLSparkExtensions extends (SparkSessionExtensions => Unit) {
 
   override def apply(extensions: SparkSessionExtensions): Unit = {
     extensions.injectParser { (spark, parser) =>
-      new FlintSparkPPLParser(parser)
+      new FlintSparkPPLParser(parser, spark)
```
LantaoJin, line 19: The SparkSession in using is passed in from here.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystPlanContext.java
```diff
@@ -283,4 +286,8 @@ public Expression resolveJoinCondition(
         isResolvingJoinCondition = false;
         return result;
     }
+
```
YANG-DB, line 289: @LantaoJin can u explain why this change is needed ?

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/RelationUtils.java
```diff
@@ -65,4 +75,23 @@ static TableIdentifier getTableIdentifier(QualifiedName qualifiedName) {
...
     }
+
+    static boolean columnExistsInCatalogTable(SparkSession spark, Field field, LogicalPlan plan) {
+        UnresolvedRelation relation = PPLSparkUtils.findLogicalRelations(plan).head();
+        QualifiedName tableQualifiedName = QualifiedName.of(Arrays.asList(relation.tableName().split("\\.")));
+        TableIdentifier sourceTableIdentifier = getTableIdentifier(tableQualifiedName);
+        boolean sourceTableExists = spark.sessionState().catalog().tableExists(sourceTableIdentifier);
+        if (sourceTableExists) {
+            try {
+                CatalogTable table = spark.sessionState().catalog().getTableMetadata(getTableIdentifier(tableQualifiedName));
```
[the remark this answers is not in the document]
YANG-DB: would make sense doing this metadata discovery more generic so that all relations have this info ?

YANG-DB (approved):
Apart from a minor question - LGTM

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/RelationUtils.java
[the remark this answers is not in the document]
LantaoJin: Yes if we see more requirements later. Now I am still willing to leave the metadata binding to Spark internal if not necessary.

integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLLookupITSuite.scala
```diff
@@ -429,4 +415,32 @@ class FlintSparkPPLLookupITSuite
 
     comparePlans(expectedPlan, logicalPlan, checkAnalysis = false)
   }
+
+  test("test LOOKUP lookupTable name") {
```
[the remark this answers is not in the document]
LantaoJin: Let me double-check the behaviour of SPL. My understand is keeping 2 `occupation` columns because we didn't know what strategies for command `LOOKUP lookupTable name`.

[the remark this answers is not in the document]
penghuo: OUTPUT is default option?
`| LOOKUP lookupTable name` equivalent to `| LOOKUP lookupTable name REPLACE dimensionTable.*`.  columns conflict-resolution should be applied. 

btw, if the output include 2 occupation columns, are these naming conflict?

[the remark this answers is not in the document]
LantaoJin: You are right, the `occupation` field from Source should be replaced with `occupation` field from Lookup if there are duplicated columns. Double-confirmed via SPL.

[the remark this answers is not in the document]
LantaoJin: But achieving this goal may bring more complexity than expected, because what is being duplicated is not the field names of the lookup table and the source table, but the field names of the lookup table and the output of the **source plan**. But our current implementation is to pass an unresolved logical plan to Spark, and at this time we have not obtained the output of the source plan.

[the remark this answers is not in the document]
LantaoJin: @penghuo , I have fixed to address your comment in https://github.com/opensearch-project/opensearch-spark/pull/1035/commits/2e954b230a22527e1e697f46c5d4a60e0ed868bf. Here is new behaivour:

source 
| id | col1 | col2 |
|----------|----------|----------|
| 1 | a | b |
| 2 | aa | bb |

lookup 
| id | col1 | col3 |
|----------|----------|----------|
| 1 | x | y |
| 3 | xx | yy |

```
index=source | lookup id
```
will output
| id | col1 | col2 | col3 |
|----------|----------|----------|----------|
| 1 | x | b | y |
| 2 | null | bb | null |

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -193,41 +199,58 @@ public LogicalPlan visitFilter(Filter node, CatalystPlanContext context) {
...
+                // If some output fields from source side duplicate to fields of lookup table, these fields will
+                // be replaced by fields from lookup table in output.
+                // For example, the lookup table contains fields [id, col1, col3] and source side fields are [id, col1, col2].
+                // For query "index = sourceTable | fields id, col1, col2 | LOOKUP lookupTable id",
+                // the col1 is duplicated field and id is mapping key (and duplicated).
+                // The query outputs 4 fields: [id, col1, col2, col3]. Among them, `col2` is the original field from source,
+                // the matched values of `col1` from lookup table will replace to the values of `col1` from source.
+                Set<Field> intersection = new HashSet<>(RelationUtils.getFieldsFromCatalogTable(context.getSparkSession(), lookupTable));
+                LOG.debug("The fields list in lookup table are {}", intersection);
+                List<Field> sourceOutput = RelationUtils.getFieldsFromCatalogTable(context.getSparkSession(), searchSide);
```
qianheng-aws, line 218: What if `searchSide` is not a table and its output has diffed from the original fields of catalog relation?

LantaoJin: Yes. I am thinking how to resolve the output of plan without submitting spark job. Any ideas?

qianheng-aws:
Found another new issue on lookup command.  Below PPL got exception:
```
[UNRESOLVED_COLUMN.WITH_SUGGESTION] A column or function parameter with name `new_col` cannot be resolved.
```

`source = $sourceTable | LOOKUP $lookupTable name REPLACE occupation as new_col`.

But it should supported as said in the doc:
>  If you specify \<outputField\> with an existing field name in source query, its values will be replaced or appended by matched values from \<inputField\>. If the field specified in \<outputField\> is a new field, an extended new field will be applied to the results.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -193,41 +199,58 @@ public LogicalPlan visitFilter(Filter node, CatalystPlanContext context) {
...
+                LOG.debug("The fields list in source output are {}", sourceOutput);
+                Set<Field> mappingFieldsOfLookup = node.getLookupMappingMap().keySet();
+                // lookup mapping keys are not concerned to drop here, it will be checked later.
+                intersection.removeAll(mappingFieldsOfLookup);
+                intersection.retainAll(sourceOutput);
+                List<Expression> duplicated = buildProjectListFromFields(new ArrayList<>(intersection), expressionAnalyzer, context)
```
qianheng-aws, line 224: If we drop the duplicated columns directly, then its behavior is different with what we do with explicit replacement as here: https://github.com/opensearch-project/opensearch-spark/blob/785d02b7252b6ee0b5acbc98b6d228503dbbaa24/ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/utils/LookupTransformer.java#L101-L102

LantaoJin: That's I said it is strange, see https://github.com/opensearch-project/opensearch-spark/pull/1035#discussion_r1944259017

LantaoJin: Not sure is it a bug of SPL or just the designed behaviour.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
qianheng-aws: If we don't need to fallback to source value when mismatching, a simple idea is ignoring the output fields of `searchSide`, it won't throw exception although we want to drop the columns not exist in the output of a plan.

Otherwise, it seems we always need an analyzed plan for handling the logic of fallback.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
qianheng-aws: If we follow that behavior, I think it will make the logic simpler. Then we need to change the logic in `buildOutputProjectList` as well to align.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
LantaoJin: Yes, why did I initially get stuck on trying to obtain the attributes of the source table?

LantaoJin:
> > Found another new issue on lookup command. Below PPL got exception:
> > ```
> > [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column or function parameter with name `new_col` cannot be resolved.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `source = $sourceTable | LOOKUP $lookupTable name REPLACE occupation as new_col`.
> > But it should be supported as said in the doc:
> > > If you specify  with an existing field name in source query, its values will be replaced or appended by matched values from . If the field specified in  is a new field, an extended new field will be applied to the results.
> 
> Good catching! It's because the current implementation `Coalesce` expression cannot be resolved with a not existed attribute. Maybe we can just add some special handling when we have a way to extract the output of source plan: only extract expressions from last `Project` and `Aggregation`

One option (the latest commit) is to keep the current query failure behaviour to simply the logic. I have updated the doc description.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
LantaoJin: > If we follow that behavior, I think it will make the logic simpler. Then we need to change the logic in `buildOutputProjectList` as well to align.

No, this strange behavior only happens for the duplicated fields when REPLACE/APPEND clause is missing. The `fall back to source output` case still existing in REPLACE/APPEND clause.

ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
YANG-DB: +1 for @qianheng-aws suggestion on `ignoring the output fields of searchSide`

integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLLookupITSuite.scala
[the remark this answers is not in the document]
penghuo: agree, the behavior is strange. but let's align with it.

Update my undestanding with lookup with [SPL test](https://github.com/penghuo/BusyBox/blob/main/splunk/lookup.md). 

The lookup command enriches your fact table by performing a left join with a lookup (dimension) table. Specifically, it:
1. Joins the fact table with the lookup table based on a join key (lookupMappingField). By default, the command appends all columns from the lookup table (except the join key) to the fact table. Alternatively, you can explicitly specify the desired output fields.
2. Resolves column name conflicts between the two tables using one of two strategies:
    * REPLACE: The lookup table’s value replaces the fact table’s value.
    * APPEND: The lookup table’s value is used only if the corresponding fact table column is missing or null.

One highligh is,
* fact table

id | col1 | col2
-- | -- | --
1 | a | b
2 | aa | bb
3 | null | ccc

* dimension table

id | col1 | col3
-- | -- | --
1 | x | y
3 | xx | yy

 if `source="fact.csv" | lookup "dim.csv" id OUTPUT id, col1, col3`, fact table id column is replaced with dimension id column, the result is, 

id | col1 | col2 | col3
-- | -- | -- | --
1 | x | b | y
null | null | bb | null
3 | xx | ccc | yy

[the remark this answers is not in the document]
LantaoJin: The output of value **null** for field `id` and `col1` of row 2 really surprised me. In our implementation, only the matched rows (explicitly joined row) will be applied the output strategies (whatever replace or append). So the output of `source="fact.csv" | lookup "dim.csv" id OUTPUT id, col1, col3` is, 

id | col1 | col2 | col3
-- | -- | -- | --
1 | x | b | y
2 | aa | bb | null
3 | xx | ccc | yy

But seems in Splunk, whatever a row is matched or not, the conflicted column from source will be replaced to new value from lookup. @penghuo will we refactor to apply Splunk's logic?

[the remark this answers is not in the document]
LantaoJin: Aligned all behaviours in https://github.com/opensearch-project/opensearch-spark/pull/1035/commits/42eb9e0f79235a8a893f587bea1891243fd1c94c

LantaoJin:
**Summary:**
Aligned with all behaviours of Splunk `lookup` command (all ITs have crossing verified), except "Specify `outputField` to a new field in APPEDN strategy". 

The latest description for `outputField` became to:
> If you specify \<outputField\> with an existing field name in source query, its values will be replaced or appended by matched values from \<inputField\>. If the field specified in \<outputField\> is a new field, in REPLACE strategy, an extended new field will be applied to the results, but fail in APPEND strategy.

@penghuo @qianheng-aws please take another look.

qianheng-aws:
LGTM

qianheng-aws (approved):

integ-test/src/integration/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLLookupITSuite.scala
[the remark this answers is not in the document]
penghuo: yes, it is not intuitive, but let's align with SPL.

LantaoJin:
No, only if AS a new field in APPEND. check the IT file, I have added the unsupported case. @penghuo

LantaoJin:
Open https://github.com/opensearch-project/opensearch-spark/issues/1040 as a followup. (Since this failure case is a corner case and it could be rewritten to workaround). IMO, It's reasonable to throw exception, so we could treat it nice to have.

penghuo (approved):
Thx!

Issue #1034, opened 2025-02-05
[FEATURE] Support query OpenSearch index with ip field


penghuo:
**Is your feature request related to a problem?**
```
{
    "clientip": {
      "type": "ip"
    }
}
```

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

PR #1044, opened 2025-02-11
Support OpenSearch ip field type

ykmr1224:
### Description
- Support OpenSearch `ip` field type
- This PR implements `ip` field to be stored as StringType.
- Following section summarize the possible approach and their pros/cons. Considering the filtering can be pushed down to OpenSearch, We don't have much benefit to use UDT or Binary type.

### Comparison of the possible approaches
1. **String Type (`StringType`)**
   - ✅ Simple and fully compatible with Spark
   - ❌ Storage inefficient and slower for filtering

2. **User Defined Type (UDT)**
   - ✅ Enables custom IP-specific operations  (e.g., subnet calculations)
   - ❌ Requires additional maintenance and potential compatibility issues
   - ❌ Serialization / Deserialization overhead

3. **Binary Type (`BinaryType`)**
   - ✅ More storage-efficient and performant
   - ❌ Less readable and requires explicit type conversion

### Related Issues
https://github.com/opensearch-project/opensearch-spark/issues/1034

### Check List
- [x] Implemented unit tests

### DIFF
--- a/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
+++ b/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
@@ -115,6 +115,11 @@ object FlintDataType {
       // binary types
       case JString("binary") => BinaryType
 
+      // ip type
+      case JString("ip") =>
+        metadataBuilder.putString("osType", "ip")
+        StringType
+
       // not supported
       case unknown => throw new IllegalStateException(s"unsupported data type: $unknown")
     }
--- a/integ-test/src/integration/scala/org/apache/spark/opensearch/table/OpenSearchTableQueryITSuite.scala
+++ b/integ-test/src/integration/scala/org/apache/spark/opensearch/table/OpenSearchTableQueryITSuite.scala
@@ -121,4 +121,27 @@ class OpenSearchTableQueryITSuite
       checkKeywordsExistsInExplain(df, expectedPlanFragment: _*)
     }
   }
+
+  test("Query index with ip data type") {
+    val index1 = "t0001"
+    withIndexName(index1) {
+      indexWithIp(index1)
+
+      var df = spark.sql(s"""SELECT client, server FROM ${catalogName}.default.$index1""")
+      val ip0 = "192.168.0.10"
+      val ip1 = "192.168.0.11"
+      val ip2 = "100.10.12.123"
+      val ip3 = "2001:db8:3333:4444:5555:6666:7777:8888"
+      val ip4 = "2001:db8::1234:5678"
+      checkAnswer(df, Seq(Row(ip0, ip2), Row(ip1, ip2), Row(ip3, ip4)))
+
+      df = spark.sql(
+        s"""SELECT client, server FROM ${catalogName}.default.$index1 WHERE client = '192.168.0.10'""")
+      checkAnswer(df, Seq(Row(ip0, ip2)))
+
+      df = spark.sql(
+        s"""SELECT client, server FROM ${catalogName}.default.$index1 WHERE server = '2001:db8::1234:5678'""")
+      checkAnswer(df, Seq(Row(ip3, ip4)))
+    }
+  }
 }
--- a/docs/opensearch-table.md
+++ b/docs/opensearch-table.md
@@ -69,6 +69,7 @@ The following table defines the data type mapping between OpenSearch index field
 | date(Date)               | DateType                          |
 | keyword                  | StringType, VarcharType, CharType |
 | text                     | StringType(meta(osType)=text)     |
+| ip                       | StringType(meta(osType)=ip)       |
 | object                   | StructType                        |
 | alias                    | Inherits referenced field type    |
 
--- a/integ-test/src/integration/scala/org/opensearch/flint/OpenSearchSuite.scala
+++ b/integ-test/src/integration/scala/org/opensearch/flint/OpenSearchSuite.scala
@@ -181,6 +181,33 @@ trait OpenSearchSuite extends BeforeAndAfterAll {
     index(indexName, oneNodeSetting, mappings, docs)
   }
 
+  def indexWithIp(indexName: String): Unit = {
+    val mappings = """{
+                     |  "properties": {
+                     |    "client": {
+                     |      "type": "ip"
+                     |    },
+                     |    "server": {
+                     |      "type": "ip"
+                     |    }
+                     |  }
+                     |}""".stripMargin
+    val docs = Seq(
+      """{
+                     |  "client": "192.168.0.10",
+                     |  "server": "100.10.12.123"
+                     |}""".stripMargin,
+      """{
+                    |  "client": "192.168.0.11",
+                    |  "server": "100.10.12.123"
+                    |}""".stripMargin,
+      """{
+                    |  "client": "2001:db8:3333:4444:5555:6666:7777:8888",
+                    |  "server": "2001:db8::1234:5678"
+                    |}""".stripMargin)
+    index(indexName, oneNodeSetting, mappings, docs)
+  }
+
   def index(index: String, settings: String, mappings: String, docs: Seq[String]): Unit = {
     openSearchClient.indices.create(
       new CreateIndexRequest(index)


flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
```diff
@@ -106,6 +106,11 @@ object FlintDataType {
       // binary types
       case JString("binary") => BinaryType
 
+      // ip type
+      case JString("ip") =>
```
dai-chen, line 110: This maps IP field type to Spark SQL type, right? Any changes required for the other direction?

dai-chen (commented):
Just wonder any benefits for UDT approach if we build functions on IP type in future?

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/datatype/FlintDataType.scala
ykmr1224: If we use UDT, we need to define a class for UDT.
For UDT and BinaryType, we also need to implement automatic conversion or function for when comparing it with string literal.

ykmr1224:
> Just wonder any benefits for UDT approach if we build functions on IP type in future?

If we expect very intensive comparison on ip address field, UDT could perform better since it can store the IP address data in more optimized way (like byte array), and don't need to parse text every time.
One possible use case is join between ip address and ip address range (or CIDR), but considering Spark doesn't have functions to support ip address related calculation/comparison, I think this is not a popular use case.

integ-test/src/integration/scala/org/apache/spark/opensearch/table/OpenSearchTableQueryITSuite.scala
```diff
@@ -65,4 +65,27 @@ class OpenSearchTableQueryITSuite extends OpenSearchCatalogSuite with FlintPPLSu
...
+
+      var df = spark.sql(s"""SELECT client, server FROM ${catalogName}.default.$index1""")
+      val ip0 = "192.168.0.10"
+      val ip1 = "192.168.0.11"
+      val ip2 = "100.10.12.123"
+      val ip3 = "2001:db8:3333:4444:5555:6666:7777:8888"
+      val ip4 = "2001:db8::1234:5678"
+      checkAnswer(df, Seq(Row(ip0, ip2), Row(ip1, ip2), Row(ip3, ip4)))
+
+      df = spark.sql(
```
penghuo, line 82: does = push down to DSL?

ykmr1224: Yes. It is pushed down.

currantw:
> > Just wonder any benefits for UDT approach if we build functions on IP type in future?
> 
> If we expect very intensive comparison on ip address field, UDT could perform better since it can store the IP address data in more optimized way (like byte array), and don't need to parse text every time. One possible use case is join between ip address and ip address range (or CIDR), but considering Spark doesn't have functions to support ip address related calculation/comparison, I think this is not a popular use case.

As mentioned in my comment [above](https://github.com/opensearch-project/opensearch-spark/pull/1044#issuecomment-2654499790), there is an existing `cidrmatch` function in PPL that may be relevant.

ykmr1224:
> Further to this conversation, [#3145](https://github.com/opensearch-project/sql/issues/3145) added support for an IP address data type in OpenSearch SQL (corresponding [pull request](https://github.com/opensearch-project/sql/pull/3175)).
> 
> In OpenSearch SQL, the IP address data type:
> 
> * Uses the [inet.ipaddr](https://seancfoley.github.io/IPAddress/IPAddress/apidocs/inet/ipaddr/package-summary.html) library.
> * Supports equality and comparison operators. Two IP address are equal if they represent the same address, even if they have different string representations.
> * Supports conversion to/from strings and comparison between IP addresses and strings.
> * Supports sorting IP addresses.
> 
> For consistency between OpenSearch SQL and Spark, it probably makes sense to have the same behaviour in OpenSearch Spark - or at least to work towards it? I'm not particularly with how UDT's work in Spark, but I assume this would be necessary to implement the above functionality?
> 
> Note that there is also an existing `cidrmatch` function in OpenSearch Spark (see [issue](https://github.com/opensearch-project/opensearch-spark/issues/671) and [pull request](https://github.com/opensearch-project/opensearch-spark/pull/706/files)). I assume that any new IP data should work with that existing function, and that the associated documentation/tests/etc. should be updated if needed?

Thank you very much for the extra context.
I didn't notice ip address support was added to OpenSearch SQL. It would make more sense to keep consistency.
Seems `cidrmatch` works against STRING field, and we need to consider it if we use UDT to store IP field.
Let me explore options to fulfill the compatibility.

ykmr1224:
I did some more investigation regarding UDT and tried if we can utilize it to realize similar functionality as in SQL plugin.
I an not quite familiar with the internal of Spark, please let me know if you have further insights or objection to my findings.

Refer [this commit](https://github.com/ykmr1224/opensearch-spark/commit/75876dc00ebdfc02868e5e41df975b3022cb9236) for the implementation.
I identified several gaps:
- Even if we use UDT, Spark calculates the equality by serialized data. In the case of above implementation, the equality is calculated by UTF8String. This means we cannot realize equality between different notation. (If we serialize to normalized IP address, it can realize equality check but that cannot recover the original notation)
- We cannot automatically cast string to UDT. In case of `WHERE ip = '192.168.0.10'`, we need explicit type conversion like `WHERE ip = text_to_ip('192.168.0.10')` (We might be able to implement custom analyzer rule to apply implicit type coercion, I haven't tried it yet.)
- If we need explicit type conversion, we cannot push down the predicate since the type conversion is done in Spark.

I'll try if implicit type coercion is possible, but let me know your thoughts.

LantaoJin:
> Refer [this commit](https://github.com/ykmr1224/opensearch-spark/commit/75876dc00ebdfc02868e5e41df975b3022cb9236) for the implementation.

I love this UDT implementation. We will try UDT in Calcite too.

ykmr1224:
I found the implementation of equality check, and we cannot override the behavior unless we make change to Spark repository.

Equality is checked by generated code, and executed in executor. The code generation is defined in [CodeGenerator.genEqual](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala#L621)

```
  def genEqual(dataType: DataType, c1: String, c2: String): String = dataType match {
    case BinaryType => s"java.util.Arrays.equals($c1, $c2)"
    case FloatType =>
      s"((java.lang.Float.isNaN($c1) && java.lang.Float.isNaN($c2)) || $c1 == $c2)"
    case DoubleType =>
      s"((java.lang.Double.isNaN($c1) && java.lang.Double.isNaN($c2)) || $c1 == $c2)"
    case st: StringType if st.supportsBinaryOrdering => s"$c1.binaryEquals($c2)"
    case st: StringType => s"$c1.semanticEquals($c2, ${st.collationId})"
    case dt: DataType if isPrimitiveType(dt) => s"$c1 == $c2"
    case dt: DataType if dt.isInstanceOf[AtomicType] => s"$c1.equals($c2)"
    case array: ArrayType => genComp(array, c1, c2) + " == 0"
    case struct: StructType => genComp(struct, c1, c2) + " == 0"
    case udt: UserDefinedType[_] => genEqual(udt.sqlType, c1, c2)
    case CalendarIntervalType => s"$c1.equals($c2)"
    case NullType => "false"
    case _ =>
      throw QueryExecutionErrors.cannotGenerateCodeForIncomparableTypeError(
        "equality", dataType)
  }
```

If we look at `case udt: ...` it falls down to `udt.sqlType`, which means the equality checking code is generated based on the underlying type used for serialization. And it does not have any extension point to override this behavior.
One possible workaround is to normalize the ip address before serializing it, but it would change the notation from original form and affect the output (even if original data is `::ffff:192.168.0.120`, output would be `192.168.0.120` and we cannot recover original form).

I think it is better storing it as String and prepare matching functions for ip address if users need.

> @currantw @penghuo @dai-chen 
Could you give your opinion?

penghuo:
> I found the implementation of equality check, and we cannot override the behavior unless we make change to Spark repository.
> 
> Equality is checked by generated code, and executed in executor. The code generation is defined in [CodeGenerator.genEqual](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/codegen/CodeGenerator.scala#L621)
> 
> ```
>   def genEqual(dataType: DataType, c1: String, c2: String): String = dataType match {
>     case BinaryType => s"java.util.Arrays.equals($c1, $c2)"
>     case FloatType =>
>       s"((java.lang.Float.isNaN($c1) && java.lang.Float.isNaN($c2)) || $c1 == $c2)"
>     case DoubleType =>
>       s"((java.lang.Double.isNaN($c1) && java.lang.Double.isNaN($c2)) || $c1 == $c2)"
>     case st: StringType if st.supportsBinaryOrdering => s"$c1.binaryEquals($c2)"
>     case st: StringType => s"$c1.semanticEquals($c2, ${st.collationId})"
>     case dt: DataType if isPrimitiveType(dt) => s"$c1 == $c2"
>     case dt: DataType if dt.isInstanceOf[AtomicType] => s"$c1.equals($c2)"
>     case array: ArrayType => genComp(array, c1, c2) + " == 0"
>     case struct: StructType => genComp(struct, c1, c2) + " == 0"
>     case udt: UserDefinedType[_] => genEqual(udt.sqlType, c1, c2)
>     case CalendarIntervalType => s"$c1.equals($c2)"
>     case NullType => "false"
>     case _ =>
>       throw QueryExecutionErrors.cannotGenerateCodeForIncomparableTypeError(
>         "equality", dataType)
>   }
> ```
> 
> If we look at `case udt: ...` it falls down to `udt.sqlType`, which means the equality checking code is generated based on the underlying type used for serialization. And it does not have any extension point to override this behavior. One possible workaround is to normalize the ip address before serializing it, but it would change the notation from original form and affect the output (even if original data is `::ffff:192.168.0.120`, output would be `192.168.0.120` and we cannot recover original form).
> 
> I think it is better storing it as String and prepare matching functions for ip address if users need.
> 
> > @currantw @penghuo @dai-chen
> 
> Could you give your opinion?

1. Mapping to Spark SQL Data Type
Since Spark SQL lacks a native IP data type, it supports implicit conversion. Mapping the OpenSearch index IP field as a string in Spark SQL is sensible because comparing string values works naturally (i.e., stringValue = stringValue). IMO, Spark-SQL and OpenSearch-SQL could maintain their own dialects. @dai-chen @LantaoJin what is your thoughts?

2. Mapping to Spark PPL Data Type
PPL supports an IP data type along with implicit conversion. Therefore, mapping the IP field as an IP data type in Spark PPL is appropriate, enabling implicit conversion for expressions such as ipValue = stringValue. In my opinion, it is acceptable not to support implicit conversion on extended UDTs, and this limitation can be noted in the documentation.

ykmr1224:
> @ykmr1224 I see some UDT example like this: https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/test/ExamplePointUDT.scala.
> 
> Can the custom `equals` in the example work for us?

I initially refer that example and thought we can override the equal method, but as I mentioned above, it didn't work and I was able to identify the actual implementation where override is not considered.

ykmr1224:
Recap of the comparison between storing ip field as string and UDT.

1. **String Type (`StringType`)**
   - ✅ Simple and fully compatible with Spark
   - ✅ Predicate can be pushed down
   - ❌ Equality check work differently when predicate is pushed down (it works as string match in spark, but work as ip address match in OpenSearch)
   - ❌ Slower for filtering 

Query Example:
```
... WHERE ip_field = '192.168.0.10';
... WHERE cidrmatch(ip_field, '192.168.0.0/24'); -- takes (string, string) as params
```

2. **User Defined Type (UDT)**
(Suppose we store normalized form in UDT for consistent equality check)

   - ✅ Provide type safe implementation for UDFs
   - ❌ Explicit type conversion will be needed when writing conditions, and it prevents predicate push down
   - ❌ Output would be different from original notation (due to normalization)
   - ❌ Serialization / Deserialization overhead

Query Example:
```
... WHERE ip_field = text_to_ip('192.168.0.10');
... WHERE cidrmatch(ip_field, '192.168.0.0/24'); -- takes (IPAddress, string) as params
```

penghuo:
> ❌ Explicit type conversion will be needed when writing conditions, and it prevents predicate push down
In `WHERE cidrmatch(ip_field, '192.168.0.0/24')` cidrmatch can be pushed down, right?. 

Couple questions.
* What is expected result If literal value is not valid?
* How to support ip type data in `create table statement`?

ykmr1224:
> In WHERE cidrmatch(ip_field, '192.168.0.0/24') cidrmatch can be pushed down, right?
I suppose it is possible (haven't been able to realize it), but we would need special logic to push this down.

> What is expected result If literal value is not valid?
Literal value would be handled in the function. And the function can ignore invalid IP address or raise error depending on the requirements. I think this is not specific to UDT.

> How to support ip type data in create table statement?
I found Spark doesn't support UDT in create table statement.
To create table with UDT field, we need to use DataFrame API.
With this approach, we can make a field with metadata and it won't differentiate UDT and type mapping with metadata.

PR #1074, opened 2025-02-26
Adding fast refresh setting during index creation

toepkerd:
### Description
This change explicitly sets the `refresh_interval` setting of an OpenSearch index to 1 second during query results index creation. This makes query results searchable sooner, reducing query latency. os-spark's ability to create the index with the correct setting in AOSS has been manually verified in preprod. Query latency improvement has also been verified in both preprod and prod environments:

PREPROD:
without fast refresh (99 trials)
avg: 31.327941333405
p50: 30.407076120377
p90: 39.990168094635

with fast refresh (94 trials)
avg: 24.779297154
p50: 22.494572401047
p90: 30.810921907425

PROD:
without fast refresh (100 trials)
avg: 24.429269089699
p50: 24.293976902962
p90: 30.645400214195

with fast refresh (99 trials)
avg: 18.175485560388
p50: 17.660987138748
p90: 20.322132110596

### Related Issues

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
@@ -108,6 +108,15 @@ trait FlintJobExecutor {
       }
     }""".stripMargin
 
+  // Fast refresh index setting for OpenSearch index. Eliminates index refresh as a source
+  // of latency for ad-hoc queries.
+  val resultIndexSettings =
+    """{
+      "index": {
+        "refresh_interval": "1s"
+      }
+    }""".stripMargin
+
   // Define the data schema
   val schema = StructType(
     Seq(
@@ -199,7 +208,7 @@ trait FlintJobExecutor {
     if (osClient.doesIndexExist(resultIndex)) {
       writeData(resultData, resultIndex, refreshPolicy)
     } else {
-      createResultIndex(osClient, resultIndex, resultIndexMapping)
+      createResultIndex(osClient, resultIndex, resultIndexMapping, resultIndexSettings)
       writeData(resultData, resultIndex, refreshPolicy)
     }
   }
@@ -375,7 +384,7 @@ trait FlintJobExecutor {
       case e: IllegalStateException
           if e.getCause != null &&
             e.getCause.getMessage.contains("index_not_found_exception") =>
-        createResultIndex(osClient, resultIndex, resultIndexMapping)
+        createResultIndex(osClient, resultIndex, resultIndexMapping, resultIndexSettings)
       case e: InterruptedException =>
         val error = s"Interrupted by the main thread: ${e.getMessage}"
         Thread.currentThread().interrupt() // Preserve the interrupt status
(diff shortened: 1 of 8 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
```diff
@@ -394,7 +394,14 @@ trait FlintJobExecutor {
       mapping: String): Either[String, Unit] = {
     try {
       logInfo(s"create $resultIndex")
-      osClient.createIndex(resultIndex, mapping)
+      val fastRefreshSettings =
```
seankao-az, line 397: 1. move constant string outside of this method, similar to `resultIndexMapping`
2. add settings argument to createResultIndex
3. nitpick: rename `fastRefreshSettings` to `resultIndexSettings`

spark-sql-application/src/main/scala/org/apache/spark/sql/OSClient.scala
```diff
@@ -61,16 +61,18 @@ class OSClient(val flintOptions: FlintOptions) extends Logging {
...
    *   Left value indicates an error.
    */
-  def createIndex(osIndexName: String, mapping: String): Unit = {
+  def createIndex(osIndexName: String, mapping: String, settings: String): Unit = {
     logInfo(s"create $osIndexName")
 
     using(flintClient.createClient()) { client =>
       val request = new CreateIndexRequest(osIndexName)
-      request.mapping(mapping, XContentType.JSON)
+      request.mapping(mapping, XContentType.JSON).settings(settings, XContentType.JSON)
```
noCharger, line 75: what would happen if settings passed in is None?

noCharger (commented):
Thanks for the change. Let's add
1. UT on malformed settings
2. IT to verify if the object is searchable within 1 second

dai-chen (commented):
Could you add PR description? Any side effect and is it worth making this configurable via Spark conf?

spark-sql-application/src/main/scala/org/apache/spark/sql/OSClient.scala
toepkerd: The short answer: a similar thing that would happen if the mapping passed in is None.

This approach closely models after how OS index mappings are passed in during index creation. There is only one caller to `OSClient.createIndex` : `FlintJobExecutor.scala`. Similar to index mapping, a fixed index setting string is declared at the top of this file as an effective const. This value is then passed into the OS.createIndex call. As such, the index settings inherits this guarantee from the index mapping approach that the passed in value will not be None.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
```diff
@@ -363,6 +374,10 @@ trait FlintJobExecutor {
     compareJson(inputJson, mappingJson) || compareJson(mappingJson, inputJson)
   }
 
+  def isValidJson(input: String): Boolean = {
+    Try(Json.parse(input)).isSuccess
```
noCharger, line 378: Can we depend on OpenSearch client's built-in JSON parsing capabilities instead of manually using JSON.parse() to avoid potential performance bottlenecks?

noCharger (approved):

PR #1077, opened 2025-03-04
Sanitize sparkUpgrade error message

Siritejagaddameedi:
### Description
Sanitize sparkUpgrade error message

We noticed a when user enter incorrect query related to date time formatter, spark throws sparkUpgradeException. This error message contains sparkUpgrade related text and we are sanitizing the error message

### Related Issues
* We need to trim the error messages to make it customer friendly

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
@@ -17,6 +17,7 @@ import org.opensearch.flint.common.scheduler.model.LangType
 import org.opensearch.flint.core.metrics.{MetricConstants, MetricsSparkListener, MetricsUtil}
 import org.opensearch.flint.spark.FlintSpark
 
+import org.apache.spark.SparkUpgradeException
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.flint.config.FlintSparkConf
 import org.apache.spark.sql.util.ShuffleCleaner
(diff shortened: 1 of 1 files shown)


dai-chen (commented):
Could you elaborate when this will happen and add IT for the changes?

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -153,6 +154,11 @@ case class JobOperator(
           }
         })
       } catch {
+        case e: SparkUpgradeException =>
+          incrementCounter(MetricConstants.RESULT_WRITER_FAILED_METRIC)
+          throwableHandler.recordThrowable(
+            s"Failed to write to result. Cause='${sanitizeSparkUpgradeErrorMessage(e.getMessage)}'",
+            e)
```
noCharger, line 161: https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala#L445 should be the start point handling query exception.

noCharger (changes requested):
UT can be added here https://github.com/opensearch-project/opensearch-spark/blob/main/spark-sql-application/src/test/scala/org/apache/spark/sql/FlintJobTest.scala

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
Siritejagaddameedi: Here the usecase being having to sanitize/trim the error message, where as `handleQueryException` handles various exceptions and adds a prefix to the error message

seankao-az:
@Siritejagaddameedi 
Are you still working on this? I believe it's pending your change.

Siritejagaddameedi:
Closing this CR. Moving this error sanitation from opensearch spark

PR #1081, opened 2025-03-14
Feature: Track index refresh progress

Swiddis:
### Description
WIP -- opening as draft to run in CI

### Related Issues
N/A

### Check List
- [ ] Implemented unit tests


flint-commons/src/main/scala/org/opensearch/flint/common/metadata/FlintMetadata.scala
```diff
@@ -41,7 +41,9 @@ case class FlintMetadata(
      */
     latestLogEntry: Option[FlintMetadataLogEntry] = None,
     /** Optional Flint index settings. TODO: move elsewhere? */
-    indexSettings: Option[String]) {
+    indexSettings: Option[String],
+    /** Information on the current progress of the index job */
+    currentProgress: Option[Map[String, Double]]) {
```
dai-chen, line 46: I think progress information should be included in each log entry, similar to other table formats like Iceberg, where each log entry tracks the number of rows or files added.

flint-core/src/main/scala/org/opensearch/flint/core/metrics/ProgressListener.scala
```diff
@@ -0,0 +1,57 @@
...
+    this.elapsedSeconds = 0.001 * (System.currentTimeMillis() - this.startTime)
+  }
+
+  def currentProgress(): Progress = Progress(
+    this.bytesProcessed,
+    // Lower-clamp divisors to avoid division by 0. In these cases it's fine if the results aren't accurate,
+    // it should stabilize quickly as more tasks are completed.
+    this.numTasks * (this.bytesProcessed
+      .asInstanceOf[Double] / Math.max(1.0, this.tasksComplete)),
+    this.bytesProcessed / Math.max(1.0, this.elapsedSeconds))
```
dai-chen, line 52: Just to confirm: are we estimating progress based on **processing rate** (i.e., bytes processed per task so far × total tasks) instead of the more typical (**progressed / total**) formula?

Swiddis: I think bytes makes more sense because users probably have a better estimate on-hand of their data scale than their file scale. "200,000 files / 1,000,000 files" doesn't say as much as "80 GB / 400 GB" or "8 TB / 40 TB." It also feels more natural to say "Throughput: 15 GB/minute" instead of "Throughput: 37500 files/minute".

flint-commons/src/main/scala/org/opensearch/flint/common/metadata/FlintMetadata.scala
Swiddis: That's the intention, I'm still working on understanding enough of the logging system to implement that...

PR #1088, opened 2025-03-26
Multiple tables as join right without subsearch is not allowed

### LINKED ISSUE #1073: [BUG] Join queries with multiple tables as join right fails with stack overflow error
**What is the bug?**
```
search source=table_a
  | INNER JOIN on column1 = column5 table_b, table_b_2
  | fields column1, column2, column3, column5
```
Failed with stack overflow error.

We can disable it to support two-table join only. But should we support query such as following query?
```
search source=table_a
  | INNER JOIN on column1 = column5 [search source = table_b, table_b_2 ]
  | fields column1, column2, column3, column5
```
This query will stackoverflow in Spark internal as well.

LantaoJin:
### Description
Multiple tables as join right without subsearch is not allowed:
- `JOIN on id = uid table1,table2` is not allowed
- `JOIN on id = uid table1,table2 as t2` is not allowed
- `JOIN on id = uid [ source = table1,table2 ]` is allowed
- `JOIN on id = uid [ source = table1,table2 ] as t2` is allowed

### Related Issues
Resolves #1073 

### Check List
- [x] Implemented unit tests

### DIFF
--- a/ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
+++ b/ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
@@ -354,8 +354,9 @@ public class CatalystQueryPlanVisitor extends AbstractNodeVisitor<LogicalPlan, C
             Optional<Expression> joinCondition = node.getJoinCondition()
                 .map(c -> expressionAnalyzer.analyzeJoinCondition(c, context));
             context.resetNamedParseExpressions();
+            LogicalPlan join = join(left, right, node.getJoinType(), joinCondition, node.getJoinHint()).clone();
             context.retainAllPlans(p -> p);
-            return join(left, right, node.getJoinType(), joinCondition, node.getJoinHint());
+            return join;
         });
     }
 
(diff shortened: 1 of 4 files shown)


ppl-spark-integration/src/main/java/org/opensearch/sql/ppl/CatalystQueryPlanVisitor.java
```diff
@@ -354,8 +354,9 @@ public LogicalPlan visitJoin(Join node, CatalystPlanContext context) {
             Optional<Expression> joinCondition = node.getJoinCondition()
                 .map(c -> expressionAnalyzer.analyzeJoinCondition(c, context));
             context.resetNamedParseExpressions();
+            LogicalPlan join = join(left, right, node.getJoinType(), joinCondition, node.getJoinHint()).clone();
             context.retainAllPlans(p -> p);
-            return join(left, right, node.getJoinType(), joinCondition, node.getJoinHint());
+            return join;
```
LantaoJin, line 359: The stackoverflow is due to plans were retained before building a join. This bug happened when the Union was added in ppl `source = table, table2`. So we should clone it then call `retainAllPlans`.

LantaoJin:
Does anyone know how to check which query failed in end2end test?

LantaoJin:
ping @penghuo @ykmr1224 @YANG-DB

YANG-DB (approved):

PR #1106, opened 2025-04-14
Call processDataframe before writeDataFrame for non Streaming Or Batch queries

manaslohani:
**…results**

### Description
This change ensures correct DataFrame formatting for non-streaming/batch queries by calling the processing step before writing results. Following the same standard here as pre #979 FlintREPL which was responsible for handling these queries. 
### Related Issues
NA

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
@@ -149,7 +149,8 @@ case class JobOperator(
             writeDataFrameToOpensearch(df, resultIndex, osClient)
           } else {
             val queryResultWriter = instantiateQueryResultWriter(sparkSession, commandContext)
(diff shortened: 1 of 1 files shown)


noCharger (approved):
Non-blocking: Please add IT to verify the df schema change.

dai-chen (approved):
Could you elaborate a bit why query result of streaming/batch doesn't need this processing?

manaslohani:
@dai-chen This is to ensure backward compatibility with #979 before which FlintREPL was acting as the entry point for these queries  https://github.com/opensearch-project/opensearch-spark/blob/bf8aa05b0908983d4f24465976a1b7f515006381/spark-sql-application/src/main/scala/org/apache/spark/sql/FlintREPL.scala#L857

manaslohani:
@noCharger surely, will follow up the IT for this.

PR #1110, opened 2025-04-14
Implement predicate pushdown for ip_equal function

ykmr1224:
### Description
- Implement predicate pushdown for ip_equal function
- Ideally, IpEqual expression should be able to be directly considered to be predicate and pushed down, but it was not possible due to following reasons
  - To enable pushdown, we need to convert UDF to ApplyFunctionExpression within Catalyst optimization rule so V2ExpressionBuilder can convert it to V2Expression
  - ApplyFunctionExpression doesn't implement Predicate interface, and it cannot be considered as Predicate.
  - When we specify UDF returning boolean like `WHERE ip_equal(ip, '192.168.0.10')`, it won't be considered as predicate due to above reason.
  - Even if we try to workaround like `WHERE ip_equal(ip, '192.168.0.10') = true`, spark automatically convert it to `WHERE ip_equal(ip, '192.168.0.10')` and fail with the same reason.
- To workaround above blocker, I made Catalyst optimizer rule to convert `ip_equal(ip, '192.168.0.10')` to `ip_compare(ip, '192.168.0.10') = 0` so the condition would generate `EqualTo` predicate instead of `ApplyFunctionExpression`
- Previous PR to implement IPAddress type: https://github.com/opensearch-project/opensearch-spark/pull/1100

### Related Issues
https://github.com/opensearch-project/opensearch-spark/issues/1034

### Check List
- [ ] Implemented unit tests


flint-spark-integration/src/main/scala/org/opensearch/flint/spark/OpenSearchCidrMatchConvertRule.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+/**
+ * Catalyst Optimizer rule for converting cidrmatch function to ip_compare for predicate pushdown.
+ * This conversion is required since Spark cannot handle UDF returning boolean in predicate
+ * pushdown logic. This is a workaround to convert to a EqualTo predicate. The converted predicate
+ * will be pushed down to OpenSearch query by {@link FlintQueryCompiler}.
+ */
+object OpenSearchCidrMatchConvertRule extends Rule[LogicalPlan] {
```
dai-chen, line 47: I feel this is really complicated. With more UDF registered in PPL in the future, we have to convert all of them to Predicate, right? If this is the right approach, any chance to make this generic enough to work with any UDF?

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/storage/FlintQueryCompiler.scala
```diff
@@ -108,20 +108,27 @@ case class FlintQueryCompiler(schema: StructType) {
...
-      for {
-        field <- compileOpt(p.children()(0))
-        value <- compileOpt(p.children()(1))
-        result <-
-          if (isTextField(field)) {
-            getKeywordSubfield(field) match {
-              case Some(keywordField) =>
-                Some(s"""{"term":{"$keywordField":{"value":$value}}}""")
-              case None => None // Return None for unsupported text fields
+      if (isIpCompare(p)) {
```
dai-chen, line 111: Is it possible to make each UDF an individual Predicate instead of deep nesting conditional check in this single case?

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/FlintPartitionReaderFactory.scala
```diff
@@ -16,9 +17,11 @@ case class FlintPartitionReaderFactory(
     schema: StructType,
     options: FlintSparkConf,
     pushedPredicates: Array[Predicate])
-    extends PartitionReaderFactory {
+    extends PartitionReaderFactory
+    with Logging {
   override def createReader(partition: InputPartition): PartitionReader[InternalRow] = {
     val query = FlintQueryCompiler(schema).compile(pushedPredicates)
+    logInfo(s"Creating FlintPartitionReader with query: $query");
```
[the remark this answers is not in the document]
ykmr1224: nice catch! fixed

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/FlintScanBuilder.scala
```diff
@@ -30,7 +30,10 @@ case class FlintScanBuilder(
 
   override def pushPredicates(predicates: Array[Predicate]): Array[Predicate] = {
     val (pushed, unSupported) =
-      predicates.partition(FlintQueryCompiler(schema).compile(_).nonEmpty)
+      predicates.partition {
+        case p => FlintQueryCompiler(schema).compile(p).nonEmpty
+        case _ => false
```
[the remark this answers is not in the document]
ykmr1224: Seems I left codes for debugging. fixed.

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/storage/FlintQueryCompiler.scala
ykmr1224: That was something I tried, but it was not possible due to current implementation of Spark. It is only for the case when we want to filter by UDF returning boolean, but Spark convert UDF to UserDefinedScalarFunc, but it cannot be considered as Predicate, and it won't be pushed down.

RyanL1997:
Closing this due to the overlap with calcite, may open if we need in the future.

PR #1123, opened 2025-04-18
[Vended Dashboards] Enable config of meta data cache by default

RyanL1997:
### Description
Enable config of meta data cache

### Related Issues

### Check List
- [x] Implemented unit tests

### DIFF
--- a/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
+++ b/flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
@@ -314,7 +314,7 @@ object FlintSparkConf {
       .createOptional()
   val METADATA_CACHE_WRITE = FlintConfig("spark.flint.metadataCacheWrite.enabled")
     .doc("Enable Flint metadata cache write to Flint index mappings")
-    .createWithDefault("false")
+    .createWithDefault("true")
 
   val CUSTOM_SESSION_MANAGER =
     FlintConfig("spark.flint.job.customSessionManager")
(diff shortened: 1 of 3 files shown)


flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
```diff
@@ -314,7 +314,7 @@ object FlintSparkConf {
       .createOptional()
   val METADATA_CACHE_WRITE = FlintConfig("spark.flint.metadataCacheWrite.enabled")
     .doc("Enable Flint metadata cache write to Flint index mappings")
-    .createWithDefault("false")
+    .createWithDefault("true")
```
Swiddis, line 317: thought: Do we still need the config at all?

We could simplify on all sides by removing the option and deleting all the `false` branches.

Swiddis (approved):

flint-spark-integration/src/main/scala/org/apache/spark/sql/flint/config/FlintSparkConf.scala
RyanL1997: I'd say we can just flip the default value for now, since there are still some tests relaying on the disable setup.

dai-chen (approved):
Thanks for the changes!

One concern is we were previously considering deprecating this cache, but now it's becoming a common interaction point between the frontend and backend. @seankao-az any potential side effects in your mind?

seankao-az:
Previous consideration was mainly due to writing to this cache means some information are duplicated between this and the datasource request index, thus need to handle consistency if any write failure occurs. So far, we're not seeing such thing happen in production, but it's still worth keeping an eye on

PR #1140, opened 2025-05-05
Modifed Integration Test for VPC, CloudTrail and WAF to test without fetching data from _source

ahkcs:
### Description

Changes are only in ```FlintSparkMaterializedViewIntegrationsITSuite.scala```

Modified integration test for Materialized View.

Originally the integration test uses data fetched from ```_source```, with the implementation of the feature of disabling _source it's no longer sufficient for testing.

This PR rewrote the integration test to make it fetch data from ```openSearchClient``` instead of ```_source```

Implemented testing for 
```VPC flow``` with unfiltered time series chart and pie chart.
```CloudTrail``` with unfiltered time series chart and pie chart.
```WAF``` with unfiltered time series chart and pie chart.


dai-chen (commented):
Does this PR include previous code changes? I think you need to rebase your branch since previous one is already merged.

Issue #1146, opened 2025-05-09
[BUG] fields identifier in PPL should be case sensitive


LantaoJin:
**What is the bug?**
Assuming the index `t` contains fields `name`, `age` and  `address`
PPL query
`source=t | fields Name, AGE, aDDress` should not work. Case sensitive in PPL should be consistency between Spark PPL and OpenSearch PPL.

PR #1148, opened 2025-05-09
Fields identifier in PPL should be case sensitive

LantaoJin:
### Description
Fields identifier in PPL should be case sensitive which is aligning with OpenSearch PPL. Ref https://github.com/opensearch-project/opensearch-spark/issues/1146

### Related Issues
Resolves https://github.com/opensearch-project/opensearch-spark/issues/1146

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/ppl-spark-integration/src/main/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLParser.scala
+++ b/ppl-spark-integration/src/main/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLParser.scala
@@ -36,6 +36,7 @@ import org.apache.spark.sql.catalyst.{FunctionIdentifier, TableIdentifier}
 import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.catalyst.parser._
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.internal.SQLConf
 import org.apache.spark.sql.types.{DataType, StructType}
 
 /**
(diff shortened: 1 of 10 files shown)


LantaoJin:
Gentle ping @dai-chen

ppl-spark-integration/src/main/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLParser.scala
```diff
@@ -57,6 +58,7 @@ class FlintSparkPPLParser(sparkParser: ParserInterface, val spark: SparkSession)
     try {
       // if successful build ppl logical plan and translate to catalyst logical plan
       val context = new CatalystPlanContext
+      spark.conf.set(SQLConf.CASE_SENSITIVE.key, value = true)
```
dai-chen, line 61: Any better place to do this? I recall we set all Spark conf on the higher level?

dai-chen (approved):

ppl-spark-integration/src/main/scala/org/opensearch/flint/spark/ppl/FlintSparkPPLParser.scala
LantaoJin: @dai-chen  Any other suggestions? I don't know is there a place to united the Spark configs. I added this conf here since 1. This is in the PPL module which has the smallest impact scope and the most familiar module for me. 2. This configuration must not be overridden by higher level, especially by users as it is a language feature.

PR #1160, opened 2025-05-14
change the version on parent metadata file

ahkcs:
change the version on parent metadata file


noCharger (commented):
will this remove the extra line of "0.8.0-commit-metadata-poc-SNAPSHOT" from metadata?

noCharger (approved):

PR #1194, opened 2025-05-29
update dependency flow using build.sbt, map commit_id and actual version

ahkcs:
update dependency flow using build.sbt, map commit_id and actual version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,7 @@ val sparkMinorVersion = sparkVersion.split("\\.").take(2).mkString(".")
 
 ThisBuild / organization := "org.opensearch"
 
-ThisBuild / version := "1.0.0-SNAPSHOT"
+ThisBuild / version := "0.8.0-commit-metadata-poc-SNAPSHOT"
 
 ThisBuild / scalaVersion := scala212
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -191,14 +284,55 @@ lazy val pplSparkIntegration = (project in file("ppl-spark-integration"))
       "com.stephenn" %% "scalatest-json-jsonassert" % "0.2.5" % "test",
       "com.github.sbt" % "junit-interface" % "0.13.3" % "test",
       "org.projectlombok" % "lombok" % "1.18.30",
-      "com.github.seancfoley" % "ipaddress" % "5.5.1",
+      "com.github.seancfoley" % "ipaddress" % "5.5.1"
```
noCharger, line 287: is it possible to leverage the maven repo as a dependency?

noCharger: > Because we need to fetch dynamic version info

can we set `0.1.0-SNAPSHOT` as a var and depend on it? why do we need to handle the logic of getLatestSnapshotVersion?

ahkcs: We will need specific version information, otherwise dependencies won't be accurately fetched like this:
```
[error]   not found: https://aws.oss.sonatype.org/content/repositories/snapshots/org/opensearch/language-grammar/0.1.0-SNAPSHOT/language-grammar-0.1.0-SNAPSHOT.pom
```

noCharger (approved):

PR #1196, opened 2025-05-30
Introducing propery shouldExecuteNextQuery to dynamically stop executing the next query

arpitjadiya:
### Description
Adding support to terminate the loop dynamically by setting canProceed to false in WarmpoolJob

### Related Issues

### Check List
- [N/A ] Updated documentation (docs/ppl-lang/README.md)
- [Y ] Implemented unit tests

### DIFF
--- a/spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
+++ b/spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
@@ -89,6 +89,10 @@ case class WarmpoolJob(
               spark.conf.get(FlintSparkConf.TERMINATE_JVM.key, "true").toBoolean
             jobOperator.start()
 
+            if (!spark.conf.get(FlintSparkConf.SHOULD_EXECUTE_NEXT_QUERY.key, "true").toBoolean) {
+              canProceed = false
+            }
+
           case _ =>
             canProceed = false
         }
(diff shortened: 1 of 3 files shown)


spark-sql-application/src/main/scala/org/apache/spark/sql/WarmpoolJob.scala
```diff
@@ -89,6 +89,10 @@ case class WarmpoolJob(
               spark.conf.get(FlintSparkConf.TERMINATE_JVM.key, "true").toBoolean
             jobOperator.start()
 
+            if (!spark.conf.get(FlintSparkConf.SHOULD_EXECUTE_NEXT_QUERY.key, "true").toBoolean) {
+              canProceed = false
+            }
```
penghuo, line 94: IMO, the expectation is first call to statementExecutionManager.getNextStatement(), return statement, then following call return None, is it correct understanding?

@noCharger  does SingleStatementExecutionManager design for this purpose?

noCharger: > @noCharger does SingleStatementExecutionManager design for this purpose?

Yes it is.

arpitjadiya: @noCharger @penghuo 
Here we are executing two statements, one is pre init and the other one is actual query.

noCharger (commented):
What issue does this PR address? The `canProceed` flag does not stop a query that is already in process. If we only want to halt the warmpool job, can we use emr cancel?

RyanL1997:
Hi @arpitjadiya, do we still need this one and also I saw @noCharger left some comments, could you also address that? Thanks.

arpitjadiya:
this is not needed now

PR #1216, opened 2025-06-12
set correct Flint Statement State after query execution

manaslohani:
### Description
Previously, the JobOperator was not updating the FlintStatement status before passing it to the QueryResultWriter, causing it to receive statements with incorrect/stale status information. This pr Updates the statement based on query execution result.

### Related Issues

### Check List
- [ ] Implemented unit tests

### DIFF
--- a/flint-commons/src/main/scala/org/opensearch/flint/common/model/FlintStatement.scala
+++ b/flint-commons/src/main/scala/org/opensearch/flint/common/model/FlintStatement.scala
@@ -66,7 +66,6 @@ class FlintStatement(
 
   def isWaiting: Boolean = state.equalsIgnoreCase(StatementStates.WAITING)
 
-
   // Does not include context, which could contain sensitive information.
   override def toString: String =
     s"FlintStatement(state=$state, statementId=$statementId, queryId=$queryId, langType=$langType, submitTime=$submitTime, error=$error)"
--- a/flint-commons/src/main/scala/org/opensearch/flint/common/model/FlintStatement.scala
+++ b/flint-commons/src/main/scala/org/opensearch/flint/common/model/FlintStatement.scala
@@ -66,6 +66,8 @@ class FlintStatement(
 
   def isWaiting: Boolean = state.equalsIgnoreCase(StatementStates.WAITING)
 
+  def isTimeout: Boolean = state.equalsIgnoreCase(StatementStates.TIMEOUT)
+
   // Does not include context, which could contain sensitive information.
   override def toString: String =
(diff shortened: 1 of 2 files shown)


Siritejagaddameedi (approved):

flint-commons/src/main/scala/org/opensearch/flint/common/model/FlintStatement.scala
```diff
@@ -66,6 +66,8 @@ class FlintStatement(
 
   def isWaiting: Boolean = state.equalsIgnoreCase(StatementStates.WAITING)
 
+  def isTimeout: Boolean = state.equalsIgnoreCase(StatementStates.TIMEOUT)
```
noCharger, line 69: seems `isTimeout` is not used

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
```diff
@@ -95,6 +95,7 @@ case class JobOperator(
           case Right(_) => data
           case Left(err) =>
             throwableHandler.setError(err)
+            statement.fail()
```
noCharger, line 98: all of these logics are covered by line 176. why do we still need this?

noCharger (commented):
Please add UT and IT to verify this change. Also not sure if this change has necessity

noCharger (changes requested):
Please add UT and IT to verify this change. Also not sure if this change has necessity

flint-commons/src/main/scala/org/opensearch/flint/common/model/FlintStatement.scala
manaslohani: Since Timeout constructs an errorDataFrame, this is useful in inferring statement state in QueryResultWriter

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
manaslohani: a statement failed at execution could still be marked as failed earlier and passed to the result writer. This can help any result writer implementation infer if the query execution failed. 
For a success case the statement stays in running state till it completes the writer flow. The actual update statement call is for covering failure/success in both execution + writer .

manaslohani:
Have retriggered Link Check and E2E tests but I believe they are failing regardless. shouldn't be an issue with the pr.

spark-sql-application/src/main/scala/org/apache/spark/sql/JobOperator.scala
noCharger: Let's use https://github.com/opensearch-project/opensearch-spark/issues/1226 to track the issue.

noCharger (approved):

PR #1247, opened 2025-08-07
fix snapshot uploading

ahkcs:
Update snapshot uploading

I changed the path to fetch the publish-snapshot.sh(https://github.com/opensearch-project/opensearch-build/blob/main/publish/publish-snapshot.sh) from opensearch-build-libraries to opensearch-build since this file no longer exists in opensearch-build-libraries

### DIFF
--- a/.github/publish-utils.sh
+++ b/.github/publish-utils.sh
@@ -233,7 +233,11 @@ publish_snapshots_and_update_metadata() {
   echo "::add-mask::$SONATYPE_PASSWORD"
   export SNAPSHOT_REPO_URL="https://aws.oss.sonatype.org/content/repositories/snapshots/"
 
-  # Publish snapshots to maven
+  mkdir -p build/resources/publish/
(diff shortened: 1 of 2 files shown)


.github/publish-utils.sh
```diff
@@ -233,7 +233,11 @@ publish_snapshots_and_update_metadata() {
   echo "::add-mask::$SONATYPE_PASSWORD"
   export SNAPSHOT_REPO_URL="https://aws.oss.sonatype.org/content/repositories/snapshots/"
 
-  # Publish snapshots to maven
+  mkdir -p build/resources/publish/
```
dai-chen, line 236: Add a line comment for this step? Previous it worked without this copy?

dai-chen (approved):

.github/publish-utils.sh
ahkcs: Added comment

noCharger (commented):
@ahkcs  what are the branches to backport?

PR #1320, opened 2026-05-27
Redact query content from Flint SQL parser fallback log

AnshMohta-cloud:
### Description
 Drops the SQL text from the fallback logInfo so customer query content
  is not written to Spark driver logs when the Flint parser cannot parse
  and we fall back to the Spark parser.
  
### Related Issues
Query String will not be logged in successfull EMR job execution

### Check List
- [ ] Implemented unit tests


YANG-DB (approved):

dai-chen (approved):
Thanks for the fix!

PR #1324, opened 2026-06-18
Redact logical plan and SQL text from analysis/parse exceptions

mengweieric:
## Problem

On a query analysis/syntax failure, the exception message embedded the Spark **logical plan** (filter literals, ARNs, account ids, column names) or the raw **SQL text**. A single `CustomLogging.logError` call in `handleQueryException` then leaked it through three channels: the persisted error JSON (→ query result store + downstream consumers), the `exception.message` log attribute, and the rendered log4j stack trace.

## Root cause

`ExtendedAnalysisException` and `ParseException` are the only types that add this content in `getMessage`. Both extend `AnalysisException`, whose `getSimpleMessage` Spark documents as *"Outputs an exception without the logical plan."*

## Fix

- Redact **structurally** (no regex, no plan-node matching, no string-splitting of the plan): use `getSimpleMessage` for any `AnalysisException`; first-line only for other throwables (defense-in-depth).
- Applied **once** at the catch point via a `RedactedException` wrapper that exposes only the sanitized message but preserves the original type name + stack frames (and carries no cause, so there's no `Caused by:` re-leak). Same sanitized throwable backs both the logs and the persisted/forwarded error.

## Scope

This PR covers redaction of the logical plan / SQL text in the **logged and persisted error message** only — the path through `handleQueryException` → `redactThrowable` → `RedactedException`.

It does **not** touch the Spark UI / SQL tab. Earlier revisions of this PR also set two `createSparkConf` defaults (`spark.sql.redaction.string.regex` and the `spark.sql.ui.redactPlanDescription.enabled` "ZOA" flag) to scrub the plan Spark publishes to the UI / event log. Those have been **removed**: the ZOA flag is only honored by a patched runtime Spark build, and we are not patching Apache Spark or the internal Spark package, so it was a no-op; the regex default was out of scope for this log-focused fix. Suppressing the plan in the Spark UI is tracked separately (it has to be addressed in the Spark build itself, not here).

## Tests

`FlintREPLTest` covers: `ExtendedAnalysisException` with a plan, `ParseException` (`== SQL ==` block), nested-in-`SparkException`, `plan = None`, non-analysis multi-line/single-line/null, and direct `sanitizedMessage`/`redactThrowable`/`RedactedException` units. Each redaction test asserts on the persisted error string (the real leak channel) and includes a precondition check that the unredacted message actually contained the secret. Verified locally: `FlintREPLTest` — 61 tests, 0 failed (sbt 1.8.2 / JDK 11).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/flint-commons/src/main/scala/org/apache/spark/sql/exception/RedactedException.scala
+++ b/flint-commons/src/main/scala/org/apache/spark/sql/exception/RedactedException.scala
@@ -1,0 +1,33 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.apache.spark.sql.exception
+
+/**
+ * A throwable that exposes only an already-sanitized message while preserving the original
+ * exception's type name and stack trace.
+ *
+ * Spark's `ExtendedAnalysisException` and `ParseException` embed customer query content in their
+ * `getMessage` (the logical plan tree and the raw SQL text respectively). The caller computes a
+ * sanitized message (e.g. via `AnalysisException.getSimpleMessage`, which Spark documents as
+ * emitting the diagnostic without the plan) and wraps the original throwable in this class before
+ * it reaches any logger or is forwarded downstream. That ensures the query content cannot leak
+ * through `getMessage`, `getLocalizedMessage`, `toString`, or the rendered stack trace, while
+ * still keeping the human-readable diagnostic and frames needed to debug where the error arose.
+ *
+ * @param originalClassName fully qualified name of the original exception type, surfaced in the
+ *                          rendered stack trace header so the error remains recognizable
+ * @param redactedMessage   the already-sanitized message (must not contain the plan or SQL text)
+ */
+class RedactedException(originalClassName: String, redactedMessage: String)
+    extends RuntimeException(redactedMessage) {
+
+  // log4j renders a throwable's stack trace header from toString(); override it so the header
+  // reads "<original type>: <redacted message>" instead of exposing this wrapper's class name.
+  override def toString: String = {
+    val msg = getMessage
+    if (msg == null) originalClassName else s"$originalClassName: $msg"
+  }
+}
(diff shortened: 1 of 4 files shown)


flint-commons/src/main/scala/org/apache/spark/sql/exception/RedactedException.scala
```diff
@@ -0,0 +1,35 @@
...
+ * through `getMessage`, `getLocalizedMessage`, `toString`, or the rendered stack trace, while
+ * still keeping the human-readable diagnostic and frames needed to debug where the error arose.
+ *
+ * @param originalClassName
+ *   fully qualified name of the original exception type, surfaced in the rendered stack trace
+ *   header so the error remains recognizable
+ * @param redactedMessage
+ *   the already-sanitized message (must not contain the plan or SQL text)
+ */
+class RedactedException(originalClassName: String, redactedMessage: String)
```
dai-chen, line 26: Is this only for hiding class name or we throw it finally? Could you confirm if any impact because I remember there is code to emit different metrics based on exception type.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
```diff
@@ -452,16 +452,59 @@ trait FlintJobExecutor {
       CleanerFactory.cleaner(streaming))
   }
 
+  /**
```
dai-chen, line 455: np: please simplify the Javadoc and move it to some public API if possible.

flint-commons/src/main/scala/org/apache/spark/sql/exception/RedactedException.scala
mengweieric: No regression here — the wrapper is only used for the log line + the persisted message string, never thrown/stored/type-matched. The type dispatch and the S3/Glue counters run on the original root cause in processQueryException, before redactThrowable is ever called, and exception.type is set from the original t.getClass.getName. Added a test to lock that in so it cant silently break later.

spark-sql-application/src/main/scala/org/apache/spark/sql/FlintJobExecutor.scala
mengweieric: Done, simplified it. Kept it private[sql] since theres no external caller — can promote to a shared util later if you want.

dai-chen (approved):
Thanks for the changes!

PR #1325, opened 2026-06-21
Retry transient connection-level IOExceptions by default

mengweieric:
### Description

FlintRetryOptions has no default for retry.exception_class_names, so by default no exception is retried (ExceptionClassNameFailurePredicate returns ex -> false on an empty list). A single transient connection fault on an idempotent request -- e.g. a ConnectionClosedException from the REST client during an index-exists check (FlintSpark.describeIndex) -- then fails permanently with all retries unused, which can tear down a long-running streaming/auto-refresh job on one blip.

Default retry.exception_class_names to java.io.IOException. The predicate matches by isInstance over the full cause-chain, and ConnectionClosedException, NoHttpResponseException, SocketTimeoutException and ConnectException all extend java.io.IOException, so one default covers them; the JDK supertype also stays correct under HTTP-client shading. Bulk requests are unaffected (getBulkRetryPolicy uses handleIf(ex -> false)). Override still available via retry.exception_class_names.

### Check List
- [ ] Implemented unit tests


dai-chen:
Please fix scalafmtCheck failure in CI.

docs/index.md
```diff
@@ -548,7 +548,7 @@ In the index mapping, the `_meta` and `properties`field stores meta and schema i
 - `spark.datasource.flint.retry.bulk.max_retries`: max retries on failed bulk request. default value is 10. Use 0 to disable retry.
 - `spark.datasource.flint.retry.bulk.initial_backoff`: initial backoff in seconds for bulk request retry, default is 4.
 - `spark.datasource.flint.retry.http_status_codes`: retryable HTTP response status code list. default value is "429,502" (429 Too Many Request and 502 Bad Gateway).
-- `spark.datasource.flint.retry.exception_class_names`: retryable exception class name list. by default no retry on any exception thrown.
+- `spark.datasource.flint.retry.exception_class_names`: additional retryable exception class name list. by default no extra exception is added here. note: transient connection-level faults (e.g. `ConnectException`, `ConnectionClosedException`, `NoHttpResponseException`, `ConnectionPoolTimeoutException`, `SocketTimeoutException`) are always retried regardless of this setting.
```
dai-chen: Can we make these exception class default value of this setting instead of "regardless of this setting"? I'm thinking in this way if anything wrong, we can simply configure this setting to override them.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/ConnectionExceptionRetryPredicate.java
```diff
@@ -0,0 +1,50 @@
...
+   * Simple class names of transient connection-level exceptions that are safe to retry.
+   * Kept as simple names so matching is independent of shading/relocation.
+   */
+  static final Set<String> RETRYABLE_CONNECTION_EXCEPTION_SIMPLE_NAMES =
+      Arrays.stream(new String[] {
+              "ConnectException",            // java.net - connection could not be established
+              "ConnectionClosedException",   // org.apache.http - connection closed mid-flight
+              "NoHttpResponseException",     // org.apache.http - server closed before responding
+              "ConnectionPoolTimeoutException", // org.apache.http - timed out leasing a connection
+              "SocketTimeoutException"       // java.net - read/connect timed out
```
dai-chen: Can we use Class.getName to handle shade jar case automatically?

docs/index.md
mengweieric: Good call, done. `retry.exception_class_names` now defaults to these connection faults instead of being a separate always-on check, so you can override the list to drop or add any. Updated the doc to match.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/ConnectionExceptionRetryPredicate.java
mengweieric: Switched to name matching. It now checks both `getClass().getName()` and `getSimpleName()`, so FQN configs keep working and the shaded classes match via simple name. We can't `Class.forName`/`isInstance` the shaded ones since the relocation prefix isn't known at compile time. Dropped the separate predicate as part of this.

flint-core/src/main/scala/org/opensearch/flint/core/http/handler/ExceptionClassNameFailurePredicate.java
```diff
@@ -5,27 +5,39 @@
...
  */
 public class ExceptionClassNameFailurePredicate extends ErrorStacktraceFailurePredicate {
 
   private static final Logger LOG = LoggerFactory.getLogger(ExceptionClassNameFailurePredicate.class);
 
   /**
-   * Retryable exception class types.
+   * Retryable exception class names (each entry is either a fully-qualified or a simple name).
    */
-  private final Set<Class<? extends Throwable>> retryableExceptions;
```
dai-chen, line 28: One concern here is subtype matching by `isInstance` is lost. Connection-refused surfaces as `HttpHostConnectException` (extends ConnectException), so the default "ConnectException" won't match it.

mengweieric: Good catch, fixed. It now walks the exception's class hierarchy (the class and its superclasses) and matches each level by FQN or simple name, so "ConnectException" again catches HttpHostConnectException while staying shading-agnostic. Added a test for that subclass case.

dai-chen (approved):
Thanks for the changes!

PR #1332, opened 2026-06-26
Get status code from runtime exception

ritvibhatt:
### Description
Fix metric misclassification where HTTP errors (e.g., 403, 429) were reported as generic 5xx failures. Unwraps AmazonServiceException from the cause chain to extract the correct status code.

### Related Issues

### Check List
- [x] Implemented unit tests

### DIFF
--- a/flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
+++ b/flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
@@ -111,7 +111,7 @@ public interface IRestHighLevelClient extends Closeable {
             CustomLogging.logError(new OperationMessage("OpenSearch Operation failed.", statusCode), openSearchException);
         } else {
             AmazonServiceException amazonServiceException = extractAmazonServiceException(t);
-            if (amazonServiceException != null) {
+            if (amazonServiceException != null && amazonServiceException.getStatusCode() != 0) {
                 statusCode = amazonServiceException.getStatusCode();
             } else {
                 // Fall back to parsing the status code from the exception message when it is
@@ -141,7 +141,7 @@ public interface IRestHighLevelClient extends Closeable {
     static OpenSearchException extractOpenSearchException(Throwable e) {
         if (e instanceof OpenSearchException) {
             return (OpenSearchException) e;
-        } else if (e.getCause() == null) {
+        } else if (e.getCause() == null || e.getCause() == e) {
             return null;
         } else {
             return extractOpenSearchException(e.getCause());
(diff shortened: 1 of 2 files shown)


flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
```diff
@@ -102,11 +105,20 @@ static void recordLatency(String metricNamePrefix, long latencyMilliseconds) {
...
-        int statusCode = openSearchException != null ? openSearchException.status().getStatus() : 500;
+        int statusCode;
         if (openSearchException != null) {
+            statusCode = openSearchException.status().getStatus();
             CustomLogging.logError(new OperationMessage("OpenSearch Operation failed.", statusCode), openSearchException);
         } else {
-            CustomLogging.logError("OpenSearch Operation failed with an exception.", t);
+            AmazonServiceException amazonServiceException = extractAmazonServiceException(t);
+            if (amazonServiceException != null) {
+                statusCode = amazonServiceException.getStatusCode();
```
dai-chen, line 115: Could you confirm if AmazonServiceException.getStatusCode() returns 0 when the SDK could not determine an HTTP status?

flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
```diff
@@ -135,4 +147,47 @@ static OpenSearchException extractOpenSearchException(Throwable e) {
...
     }
+
+    /**
+     * Extracts an AmazonServiceException from the given Throwable.
+     * Checks if the Throwable is an instance of AmazonServiceException or caused by one.
+     *
+     * @param e the exception to be checked
+     * @return the extracted AmazonServiceException, or null if not found
+     */
+    static AmazonServiceException extractAmazonServiceException(Throwable e) {
```
dai-chen, line 158: I recall some exception may have itself as cause. Please check what's the standard way to extract root cause. The extractOpenSearchException may have similar issue if so.

flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
```diff
@@ -135,4 +147,47 @@ static OpenSearchException extractOpenSearchException(Throwable e) {
...
+            return (AmazonServiceException) e;
+        } else if (e.getCause() == null) {
+            return null;
+        } else {
+            return extractAmazonServiceException(e.getCause());
+        }
+    }
+
+    /** Pattern matching an HTTP status code embedded in an exception message, e.g. "Status Code: 403". */
+    Pattern STATUS_CODE_PATTERN = Pattern.compile("Status Code:\\s*(\\d{3})");
```
dai-chen: Just wondering can we narrow this down. Since this is only for error message, probably we only extract 4xx or 5xx?

flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
ritvibhatt: It does, updated to check if it is 0 so it will fall back to default if it can't determine status code thanks!

flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
ritvibhatt: Updated extractAmazonServiceException and extractOpenSearchException to check if cause is equal to itself
```
else if (e.getCause() == null || e.getCause() == e) {
            return null;
```
following ExceptionMessages.java from this same package.

flint-core/src/main/java/org/opensearch/flint/core/IRestHighLevelClient.java
ritvibhatt: Makes sense updated to only extract 4xx/5xx!

dai-chen (approved):
Thanks for the fix!

